PR #10, opened 2023-06-19
LC-273:  Indexer send to index based on doc field.

kyle-kmw:
Implemented support in Indexer for sending to index based on a field stored on the documents to be indexed. Added MultiBatch to manage batches targeting different indexes. Added multi collection support to SolrIndexer implementation.


lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -48,10 +49,11 @@ public void terminate() {
   public Indexer(Config config, IndexerMessageManager manager, String metricsPrefix) {
     this.manager = manager;
     this.idOverrideField = config.hasPath("indexer.idOverrideField") ? config.getString("indexer.idOverrideField") : null;
+    this.indexOverrideField = config.hasPath("index.indexOverrideField") ? config.getString("indexer.indexOverrideField") : null;
```
rseitz, line 52: first occurrence should be should be "indexer.indexOverrideField", not "Index.indexOverrideField"

lucille-core/src/main/java/com/kmwllc/lucille/core/SingleBatch.java
```diff
@@ -0,0 +1,93 @@
+package com.kmwllc.lucille.core;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.LinkedBlockingQueue;
+
+/**
```
rseitz, line 9: This class-level javadoc and all the method-level javadocs should be copied or moved to the Batch interface.

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -48,10 +49,11 @@ public void terminate() {
   public Indexer(Config config, IndexerMessageManager manager, String metricsPrefix) {
     this.manager = manager;
     this.idOverrideField = config.hasPath("indexer.idOverrideField") ? config.getString("indexer.idOverrideField") : null;
+    this.indexOverrideField = config.hasPath("indexer.indexOverrideField") ? config.getString("indexer.indexOverrideField") : null;
```
rseitz, line 52: What happens if someone specifies indexer.indexOverrideField with an elasticsearch or opensearch indexer? Seems like everything would work, but we'd be creating a MultiBatch without actually sending docs to separate indices. Do we plan to support this property for elastic and opensearch or is it intended as solr-only? Should probably throw an exception if someone tries to use it in a context where it's not supported.

lucille-core/src/main/java/com/kmwllc/lucille/core/SingleBatch.java
kyle-kmw: Thanks @rseitz I moved them.

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
kyle-kmw: @rseitz I just added exceptions to the ElasticIndexer, OpenSearchIndexer, and CSVIndexer. Please take a look.

I do think it makes sense to eventually support this in the other indexers, but this LC-273 was for supporting it in Solr.

rseitz: The current approach forces every Indexer subclass that doesn't support indexer.OverrideField to check for it and throw an exception.

It might be better to have a design where the exception is thrown by default unless the subclass does something special to support this property. However, if we're planning to support this property in all the other Indexer subclasses, the point might be moot.

So basically, I'm OK with what's there now but wanted to mention this for you to consider.

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -102,17 +100,32 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
+        solrDocsByCollection.get(docIndex).add(solrDoc);
+      } else {
+        List<SolrInputDocument> solrDocs = new LinkedList<>();
+        solrDocsByCollection.put(docIndex, solrDocs);
+        solrDocs.add(solrDoc);
+      }
+    }
+    for (String collection: solrDocsByCollection.keySet()) {
+      List<SolrInputDocument> solrDocs = solrDocsByCollection.get(collection);
+      if(collection == null) {
```
rseitz, line 123: space between "if" and "("

lucille-core/src/test/java/com/kmwllc/lucille/indexer/SolrIndexerTest.java
```diff
@@ -387,6 +446,37 @@ public void testIndexerWithChildDocWithNestedJsonWithObjects() throws Exception
       Event.Type.FAIL, equalTo(events.get(0).getType()));
   }
 
+  /**
```
rseitz, line 449: I added this test last week. Looks like it's now redundant with testSolrIndexerWithMultipleCollections() that you added. If that's the case, feel free to remove it.

In the future if I'm making changes to a PR I'll let you know beforehand or else do it in a separate PR. Sorry about that. I had sketched this out and thought it would be easier to commit than describe it in a comment.

lucille-core/src/test/java/com/kmwllc/lucille/indexer/SolrIndexerTest.java
```diff
@@ -210,6 +209,66 @@ public void testIndexerWithNestedChildDocs() throws Exception {
...
+      assertEquals(events.get(i).getDocumentId(), sDoc.getFieldValue(Document.ID_FIELD));
+      assertEquals(Event.Type.FINISH, events.get(0).getType());
+    }
+    assertEquals(2, collectionDocs.size());
+    assertEquals(2, collectionDocs.get("col1").size());
+    assertEquals("doc1", collectionDocs.get("col1").get(0).getFieldValue(Document.ID_FIELD));
+    assertEquals("doc3", collectionDocs.get("col1").get(1).getFieldValue(Document.ID_FIELD));
+    assertEquals(2, collectionDocs.get("col2").size());
+    assertEquals("doc2", collectionDocs.get("col2").get(0).getFieldValue(Document.ID_FIELD));
+    assertEquals("doc4", collectionDocs.get("col2").get(1).getFieldValue(Document.ID_FIELD));
```
rseitz, line 269: Should confirm that the indexOverrideField is not present on the docs that get sent.

rseitz (approved):
Approved.

Suggest creating a separate ticket for updating Batch to have flushByCollection() that returns a Map<String, List> and updating Indexer.sendToIndex to take Map<String, List>.

Suggest creating a separate ticket for fixing remaining instances of "if(" and "for(" in the codebase.

PR #14, opened 2023-07-06
database connector updates to support some joining database connector use cases

kwatters:


lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -341,19 +379,38 @@ public void setIdField(String idField) {
...
-      throw new ConnectorException("Error closing the connection", e);
+    for (Connection connection : connections) {
+      if (connection == null) {
+        // no-op
+        continue;
+      }
+      try {
+        connection.close();
+      } catch (SQLException e) {
+        throw new ConnectorException("Error closing the connection", e);
```
kyle-kmw, line 390: If there are multiple connections and closing the first one causes an exception, we won't close the other connections because of the throw here.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -256,15 +294,15 @@ private String[] getColumnNames(ResultSet rs) throws SQLException {
     return names;
   }
 
-  private void runSql(String sql) {
+  private void runSql(Connection connection, String sql) {
     if (!StringUtils.isEmpty(sql)) {
       try {
         Statement state = connection.createStatement();
         state.executeUpdate(sql);
         state.close();
```
kyle-kmw, line 302: If executeUpdate throws an exception the statement won't be closed. The statement can be put inside a try-with-resources.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -176,7 +180,7 @@ public void execute(Publisher publisher) throws ConnectorException {
       ors.close();
```
kyle-kmw, line 180: I realize this isn't new code but it feels like there should be a TODO/task to clean up the way that the result sets are closed so that they are performed in a finally block or opened in a try-with-resources.

lucille-core/src/main/java/com/kmwllc/lucille/core/HashMapDocument.java
```diff
@@ -613,6 +613,10 @@ public boolean hasChildren() {
     return data.contains(CHILDREN_FIELD) && data.getMany(CHILDREN_FIELD).size() > 0;
   }
 
+  public void removeChildren() {
```
rseitz, line 616: new method on Document here -- should add coverage for this in DocumentTest

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
```diff
@@ -218,18 +218,11 @@ public void testClose() throws ConnectorException, SQLException {
...
-    Connection connection = connector.getConnection();
-
-    // verify that the connection has opened
-    assertFalse(connection.isClosed());
-
+    assertFalse(connector.isClosed());
     connector.close();
-
     // verify that the connection is actually closed
-    assertTrue(connection.isClosed());
```
rseitz, line 233: we should still check that the connection itself is closed

lucille-core/src/main/java/com/kmwllc/lucille/stage/CollapseChildrenDocuments.java
```diff
@@ -0,0 +1,42 @@
...
+
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+
+    if (doc.hasChildren()) {
+      // We need to iterate the children and copy their field values to the parent.
+      for (Document child : doc.getChildren()) {
+        for (String fieldToCopy : fieldsToCopy) {
+          if (child.has(fieldToCopy)) {
+            doc.addToField(fieldToCopy, child.getString(fieldToCopy));
```
rseitz, line 30: this is treating all child field values as strings, what if some are ints etc?

kwatters: we would need a plain untyped method on document that returns Object to do this field value copy in a way that worked for all types of objects.  For now, we've chosen string because it generally works with numbers and text...  other field values like dates or bytes probably won't get collapsed down properly here.

rseitz: you can use the following method on Document:
  public void setOrAdd(String name, Document other)

It takes a field name and another document. It'll add the field from the other document to the current one in a way that preserves the type.

We faced the same problem when implementing the collapsing feature on Connector -- this method, along with   public void setOrAddAll(Document other), was created to handle this situation properly.

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
kwatters: connector.getConnection() no longer exists...  I'd have to add some getters to the class to get to the underlying list connections now.   I feel like isClosed() on the connector is probably ok.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -62,28 +68,28 @@ public DatabaseConnector(Config config) {
...
   // create a jdbc connection
-  private void createConnection() throws ClassNotFoundException, SQLException {
+  private Connection createConnection() throws ClassNotFoundException, SQLException {
+    Connection connection = null;
     try {
       Class.forName(driver);
     } catch (ClassNotFoundException e) {
-      // TODO better error handling/logging
-      e.printStackTrace();
+      log.error("Driver not found {} check classpath to make sure the jdbc driver jar file is there.", driver, e);
```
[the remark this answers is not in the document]
kwatters: no double logging anymore.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
kwatters: now we catch it and continue to try to close the rest.

lucille-core/src/main/java/com/kmwllc/lucille/stage/CollapseChildrenDocuments.java
kwatters: updated to use setOrAdd(fieldname, Document)

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
kwatters: TODO added.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
kwatters: it's in a try with resources now.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -62,28 +68,28 @@ public DatabaseConnector(Config config) {
...
       setState(ConnectorState.ERROR);
       throw(e);
     }
     try {
       connection = DriverManager.getConnection(connectionString, jdbcUser, jdbcPassword);
     } catch (SQLException e) {
-      // TODO better logging/error handling.
-      e.printStackTrace();
+      log.error("Unable to connect to database {} user:{}", connectionString, jdbcUser, e);
```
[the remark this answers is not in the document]
kwatters: no longer double logging the exception.

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
kyle-kmw: I think it's good not to have a getter that exposes the underlying connections. I think it would be better to have a test that checks that the connections are really closed though.

I think we can use the connection that dbHelper has open to inspect the number of sessions against the h2 database.

```
  public int checkNumConnections() throws SQLException {
    try(ResultSet rs = RunScript.execute(conn, new StringReader("select count(*) from information_schema.sessions;"))) {
      rs.next();
      return rs.getInt(1);
    }
  }
```
Should be 1 after connector.close() is called (1 being the connection owned by dbHelper)

kwatters: cool...  turns out DBTestHelper holds onto a connection from the beginning to the end of the unit tests.  So, I updated the tests to assert the number of connection is 1 before and after each of the test cases.

kyle-kmw (approved):

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
```diff
@@ -86,10 +89,16 @@ public void testDatabaseConnector() throws Exception {
     assertEquals("Blaze", docsSentForProcessing.get(2).getStringList("name").get(0));
     assertEquals("Cat", docsSentForProcessing.get(2).getStringList("type").get(0));
     
+    connector.close();
```
rseitz, line 92: Should call assertEquals(2, dbHelper.checkNumConnections()) or something similar, right before connector close. Otherwise wise we're asserting 1 connection at the beginning and 1 at the end but never seeing it change anywhere.

rseitz (approved):

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -62,123 +68,117 @@ public DatabaseConnector(Config config) {
...
-    runSql(preSql);
 
-    ResultSet rs = null;
     try {
+      setState(ConnectorState.RUNNING);
+      // connect to the database.
+      Connection connection = createConnection();
+      // run the pre-sql (if specified)
+      runSql(connection, preSql);
+      // TODO: make sure we cleanup result set/statement/connections properly.
```
rseitz, line 104: Is there somewhere we know or suspect we're not cleaning up connections?

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -62,123 +68,117 @@ public DatabaseConnector(Config config) {
...
+          // this is the primary key that the result set is ordered by.
+          Integer joinId = rs.getInt(idField);
+          int childId = -1;
+          int j = 0;
+          for (ResultSet otherResult : otherResults) {
+            iterateOtherSQL(otherResult, otherColumns.get(j), doc, joinId, childId, otherJoinFields.get(j));
+            j++;
+          }
         }
+        // Fix the duplicate id field value that occurs
```
rseitz, line 167: is this comment still current?

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -193,19 +193,35 @@ private void flush() {
...
         // we've gone too far.. lets back up and break out , move forward the primary result set.
-        rs2.previous();
-        break;
+        // we should leave the cursor here so we can test again when the primary result set is advanced.
+        return;
       }
+      
+      // here we have a match for the join keys.. let's create the child doc for this joined row.
       childId++;
       Document child = Document.create(Integer.toString(childId));
```
rseitz, line 226: so the child IDs are just integers without any prefix based on the parent ID?

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
kwatters: nope , i've removed the comment.

lucille-core/src/test/java/com/kmwllc/lucille/stage/CollapseChildrenDocumentsTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    Document child2 = Document.create("child2");
+    child2.setField("childfield1", "foo");
+    child2.setField("childfield2", "bar");
+    child2.setField("childfield3", "baz");
+    doc.addChild(child2);
+
+    stage.processDocument(doc);
+    
+    assertFalse(doc.hasChildren());
```
rseitz, line 39: should ideally also have a test where the config has dropChildren: false 
and we make sure the children aren't dropped

lucille-core/src/test/java/com/kmwllc/lucille/stage/CollapseChildrenDocumentsTest.java
```diff
@@ -0,0 +1,49 @@
...
+    assertTrue(doc.has("childfield1"));
+    assertTrue(doc.has("childfield2"));
+    assertFalse(doc.has("childfield3"));
+    
+    List<String> values = doc.getStringList("childfield1");
```
rseitz, line 44: should ideally look at values in childfield1 and childfield2, make sure values are correct and occur in proper order. suggest adding foo1 and bar1 to first child doc and foo2 and bar2 to second child doc so we can test that the ordering is correct after collapse

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
```diff
@@ -156,13 +170,17 @@ public void testJoiningDatabaseConnector() throws Exception {
     // TODO: better verification / edge cases.. also formalize the "children" docs.
     String expected ="{\"id\":\"1\",\"name\":\"Matt\",\".children\":[{\"id\":\"0\",\"meal_id\":\"1\",\"animal_id\":\"1\",\"name\":\"breakfast\"},{\"id\":\"1\",\"meal_id\":\"2\",\"animal_id\":\"1\",\"name\":\"lunch\"},{\"id\":\"2\",\"meal_id\":\"3\",\"animal_id\":\"1\",\"name\":\"dinner\"}],\"run_id\":\"testRunId\"}";
     assertEquals(expected, docs.get(0).toString());
+    
+    connector.close();
+    assertEquals(1, dbHelper.checkNumConnections());
 
   }
   
   // TODO: not implemented yet.
```
rseitz, line 179: outdated TODO?

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
```diff
@@ -156,13 +170,17 @@ public void testJoiningDatabaseConnector() throws Exception {
...
+    
+    connector.close();
+    assertEquals(1, dbHelper.checkNumConnections());
 
   }
   
   // TODO: not implemented yet.
   // @Test
   public void testCollapsingDatabaseConnector() throws Exception {
     // TODO: implement me
```
rseitz, line 182: outdated TODO?

lucille-examples/lucille-joining-db-connector-example/scripts/run_ingest.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+java -Dconfig.file=conf/joining-db-connector-example.conf -cp 'lib/*' com.kmwllc.lucille.core.Runner -local
```
rseitz, line 2: where is conf/joining-db-connector-example.conf in this PR?

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
rseitz, line 1: looking at this again, I'm seeing a lot of new coverage added in the PR. So how are we specifically flexing the new functionality that's been added to the connector?

lucille-core/src/test/java/com/kmwllc/lucille/stage/CollapseChildrenDocumentsTest.java
kwatters: i added another test to assert that children documents were not dropped.

lucille-core/src/test/java/com/kmwllc/lucille/stage/CollapseChildrenDocumentsTest.java
kwatters: updated unit test to validate the order of the values.

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
kwatters: This changes some details of how the connections are made to the database, and how the result sets are iterated, the existing unit tests covered the functionality and verified that we didn't break anything.

lucille-examples/lucille-joining-db-connector-example/scripts/run_ingest.sh
kwatters: missing...  I'll add an example.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
rseitz: We should go through all TODOs in the codebase, but since that's not part of a regular process yet, how about making a ticket for this -- or fixing it with this PR?

rseitz (approved):
Approving this now to not block merging while I'm away.
Requesting the missing conf file & either a fix to the connection cleanup TODO or a ticket created for it before merging.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
meghanboyd: Created a ticket LC-316 to address connection cleanup so that the merging of this PR is not otherwise blocked.

PR #15, opened 2023-07-07
LC-274: Implemented Support for in order deletes for the SolrIndexer implementation.

kyle-kmw:


lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrDocRequests.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.kmwllc.lucille.core.Document;
+import org.apache.solr.common.SolrInputDocument;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class SolrDocRequests {
```
rseitz, line 11: class-level javadoc would be helpful here

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrDocRequests.java
```diff
@@ -0,0 +1,62 @@
...
+      List<SolrInputDocument> solrAddUpdateDocs,
+      Set<String> solrAddUpdateIds,
+      Set<String> solrDeleteIds) {
+    this.solrAddUpdateDocs = solrAddUpdateDocs;
+    this.solrAddUpdateIds = solrAddUpdateIds;
+    this.solrDeleteIds = solrDeleteIds;
+  }
+
+  public void addDocForAddUpdate(SolrInputDocument doc) {
+    this.solrAddUpdateDocs.add(doc);
```
rseitz, line 30: "this." isn't needed here -- this is outside a constructor and there's no ambiguity about solrAddUpdateDocs (it can only be the instance variable with that name)... same comment for the other methods in this class

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrDocRequests.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class SolrDocRequests {
+  private final List<SolrInputDocument> solrAddUpdateDocs;
+  private final Set<String> solrAddUpdateIds;
+  private final Set<String> solrDeleteIds;
```
rseitz, line 14: For the names of solrAddUpdateDocs, solrAddUpdateIds, and solrDeleteIds, I'd suggest that "solr" isn't needed because it's already part of the class name "SolrDocRequests". However, addUpdateDocs, addUpdateIds, and deleteIds sound like they could be method names, so I'm suggesting "docsToAddOrUpdate", "idsToAddOrUpdate", and "idsToDelete"

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -48,12 +51,38 @@ public void terminate() {
...
+            ? config.getStringList("indexer.ignoreFields")
+            : null;
+    this.deletionMarkerField =
+        config.hasPath("indexer.deletionMarkerField")
+            ? config.getString("indexer.deletionMarkerField")
+            : null;
+    this.deletionMarkerFieldValue =
+        config.hasPath("indexer.deletionMarkerFieldValue")
+            ? config.getString("indexer.deletionMarkerFieldValue")
+            : null;
```
rseitz, line 73: if people are going to use "isDelete" or "delete" as the marker field most of the time, it might be nice to set a default of "true" for the deletionMarkerFieldValue, but I guess we can't be sure of that. Thinking out loud -- no action needed.

lucille-core/src/test/java/com/kmwllc/lucille/indexer/SolrIndexerTest.java
```diff
@@ -275,7 +296,176 @@ public void testSolrIndexerWithMultipleCollections() throws Exception {
...
+            .withValue("indexer.batchSize", ConfigValueFactory.fromAnyRef(1))
+            .withValue(
+                "indexer.deletionMarkerField", ConfigValueFactory.fromAnyRef(deletionMarkerField))
+            .withValue(
+                "indexer.deletionMarkerFieldValue",
+                ConfigValueFactory.fromAnyRef(deletionMarkerFieldValue));
+    PersistingLocalMessageManager manager = new PersistingLocalMessageManager();
+
+    Document doc1 = Document.create("doc1", "test_run");
+    doc1.addToField(deletionMarkerField, deletionMarkerFieldValue);
```
rseitz, line 317: Should also create a doc that has a the deletionMarkerField populated with a value that's different from deletionMarkerFieldValue and confirm that this doc _isn't_ deleted.

lucille-core/src/test/java/com/kmwllc/lucille/indexer/SolrIndexerTest.java
```diff
@@ -275,7 +296,176 @@ public void testSolrIndexerWithMultipleCollections() throws Exception {
...
+    doc1.addToField("version", 1);
+    Document doc2 = Document.create("doc1", "test_run");
+    doc2.addToField(deletionMarkerField, deletionMarkerFieldValue);
+    Document doc3 = Document.create("doc1", "test_run");
+    doc3.addToField("version", 2);
+
+    SolrClient solrClient = mock(SolrClient.class);
+    Indexer indexer = new SolrIndexer(config, manager, solrClient, "");
+    manager.sendCompleted(doc1);
+    manager.sendCompleted(doc2);
```
rseitz, line 359: It might be worth adding an unrelated add before doc1 and confirming that the both adds are sent before the deleteById... In other words, demonstrate that delete causes all pending adds to be processed first, whether the pending add is for the doc with ID "doc1" or something different.

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -152,6 +152,7 @@ protected void sendToIndex(List<Document> documents) throws Exception {
 
         if (solrDocRequests.containsIdForDeletion(solrId)) {
           sendDeletionBatch(collection, solrDocRequests.getDeleteIds());
+          solrDocRequests.resetDeletes();
```
rseitz, line 155: Good catch here. Were any tests failing when this was omitted? If not, I suggest flexing this in a test. Thinking out loud -- we'd need a delete, then an add (to cause the deletions to be sent), then another delete which should be for a _different_ doc. Then when deletes are sent the second time, need to make sure the delete for the first doc isn't issued again...

kyle-kmw: Both of the two most recently added tests "testSolrIndexerUnrelatedAddAddDeleteAdd" and "testSolrIndexerDeleteAddUnrelatedDelete" will fail if the line to resetDeletes is removed.

rseitz (approved):

PR #24, opened 2023-07-31
LC-276 New fetch uri stage and tests

abijitran:


lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,112 @@
...
+  private final String source;
+  private final String dest;
+  private final String statusSuffix;
+  private final String sizeSuffix;
+
+  private final String errorSuffix;
+  private final int maxDownloadSize;
+  private CloseableHttpClient client;
+
+  public FetchUri(Config config) {
```
rseitz, line 37: Should have some class-level javadoc saying what this stage does. Should mention any behaviors that might not be obvious e.g. the stage does nothing if the source field is absent on a document, or empty

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+    if (!doc.has(source) || doc.getString(source).isEmpty()) {
+      return null;
+    }
+    String uri = doc.getString(source);
+
+    // Following checks for first valid URL and if not, then valid URI, and then if not places error message in error field
```
rseitz, line 64: A URI can be a URL or URN. I don't think we need to try to support URNs here, so it should be sufficient to simply require a valid URL.

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
abijitran: Added, also added javadoc for all config params and what they are for

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,112 @@
...
+        .withOptionalProperties("size_suffix", "status_suffix", "max_size", "error_suffix"));
+    this.source = config.getString("source");
+    this.dest = config.getString("dest");
+    this.statusSuffix = ConfigUtils.getOrDefault(config, "status_suffix", "status_code");
+    this.sizeSuffix = ConfigUtils.getOrDefault(config, "size_suffix", "size");
+    this.errorSuffix = ConfigUtils.getOrDefault(config, "error_suffix", "error");
+    this.maxDownloadSize = ConfigUtils.getOrDefault(config, "max_size", Integer.MAX_VALUE);
+  }
+
+  void setClient(CloseableHttpClient client) {
```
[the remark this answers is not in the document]
abijitran: Added comment for this

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
abijitran: Changed the validation to now only check for valid URL, updated comment as well

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,122 @@
...
+import com.kmwllc.lucille.core.StageException;
+import com.typesafe.config.Config;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.util.Iterator;
+import nl.altindag.ssl.util.UriUtils;
```
kyle-kmw, line 15: Do we need this import?

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
rseitz: Great! But where does the doc mention what the stage does if the source field is not present on the doc, or is present but empty?

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
abijitran: Not required, updated this in new commit

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,120 @@
...
+      InputStream content = ent.getContent();
+      BoundedInputStream boundedContentStream = new BoundedInputStream(content, maxDownloadSize);
+      byte[] bytes = IOUtils.toByteArray(boundedContentStream);
+      long contentSize = bytes.length;
+
+      doc.setField(dest, bytes);
+      doc.setField(source + "_" + statusSuffix, statusCode);
+      doc.setField(source + "_" + sizeSuffix, contentSize);
+
+      EntityUtils.consume(ent);
```
kyle-kmw, line 100: This should be in a finally I think?

abijitran: I think it should probably be in a finally; I updated it in the new commit

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,130 @@
...
+      doc.setField(dest, bytes);
+      doc.setField(source + "_" + statusSuffix, statusCode);
+      doc.setField(source + "_" + sizeSuffix, contentSize);
+    } catch (ClientProtocolException e) {
+      setErrorField(doc, e);
+    } catch (IOException e) {
+      setErrorField(doc, e);
+    } finally {
+      try {
+        EntityUtils.consume(ent);
```
kyle-kmw, line 106: If ent is null here will EntityUtils handle it okay? Just checking if we need our own null check here.

abijitran: Looking at the code for .consume, it checks whether it's null. It doesn't do anything if it's null

abijitran: `public static void consume(HttpEntity entity) throws IOException {
    if (entity != null) {
      if (entity.isStreaming()) {
        InputStream inStream = entity.getContent();
        if (inStream != null) {
          inStream.close();
        }
      }

    }
  }`

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,120 @@
...
+      doc.setField(source + "_" + errorSuffix, e.getClass().getCanonicalName() + " " + e.getMessage());
+      return null;
+    }
+
+    HttpGet httpGet = new HttpGet(url);
+    try (CloseableHttpResponse httpResponse = client.execute(httpGet)) {
+      int statusCode = httpResponse.getStatusLine().getStatusCode();
+      HttpEntity ent = httpResponse.getEntity();
+      InputStream content = ent.getContent();
+      BoundedInputStream boundedContentStream = new BoundedInputStream(content, maxDownloadSize);
```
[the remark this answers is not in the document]
abijitran: Opened in try-with-resources now

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
rseitz: I'm wondering if we need EntityUtils.consume(ent) at all now... all it does is close the content InputStream.
We're creating a BoundedInputStream from ent.getContent() in a try-with-resources.
It looks like BoundedInputStream.close() propogates the close() to the underling InputStream when isPropagateClose() is true. Looking at code for BoundedInputStream, true is the default. Thoughts?
https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/BoundedInputStream.html#close()

abijitran: I looked at the source code for both .close() and .consume(which uses .close()) and I think that we can remove the .consume. I also was reading this https://stackoverflow.com/questions/65298317/how-does-consuming-till-the-end-of-stream-using-entityutils-consumehttpentity and came to the same conclusion. With .consume, if the .getContent() from the method creates an IOException, the stream wouldn't close. Would this cause us to favor .close() instead?

kyle-kmw: Yeah I agree. The EntityUtils.consume can be removed and also the finally block.

Ultimately, we want all of the closeable stuff that we've opened to be closed cleanly and consistently which I think now that we have all of the Closable things in try-with-resources that is achieved without needing EntityUtils.

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -0,0 +1,124 @@
...
+    } catch (URISyntaxException e) {
+      setErrorField(doc, e);
+      return null;
+    } catch (MalformedURLException e) {
+      setErrorField(doc, e);
+      return null;
+    }
+
+    HttpGet httpGet = new HttpGet(url);
+    HttpEntity ent = null;
```
rseitz, line 88: looks like this declaration can be moved inside the try, l91?

rseitz (approved):

kyle-kmw (approved):

abijitran:
Thanks for reviewing the pr, appreciate the suggestions!

PR #25, opened 2023-08-03
LC-314 add multivalued JsonNode Document field support

h0rban:
Action items:
- [x] add tests to new document functionality. specifically the difference between ArrayNode and JsonNode
- [ ] remove Postgres dependency from examples
- [ ] check what to do about the driver in DatabaseConnector


lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
```diff
@@ -511,6 +516,35 @@ public List<byte[]> getBytesList(String name) {
     return result;
   }
 
+  @Override
+  public JsonNode getJson(String name) {
```
rseitz, line 520: Need to test how this behaves when we add a JsonNode that happens to be an ArrayNode as a single value on a document and then try to get it back via getJson(String name). Ideally we'd get back the ArrayNode itself -- the same thing we added.

However getSingleNode() is going to check isMultiValued() which is going to check whether we're looking at an ArrayNode. The document will think the ArrayNode represents multiple values and getJson() will return the first contained value in the ArrayNode?  I believe this is why we didn't implement these json getters and setters before. It's not clear how to distinguish an ArrayNode that should be treated as a single value versus an ArrayNode that was created internally by JsonDocument to represent a multi-valued field.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -149,6 +151,13 @@ public void execute(Publisher publisher) throws ConnectorException {
         // log.info("Add Field {} Value {} -- Doc {}", columns[i-1].toLowerCase(), rs.getString(i), doc);
         String fieldValue = rs.getString(i);
         if (fieldValue != null) {
+
+          // TODO THROWS ERROR IF IS SET AND THE RESULT HAS A COLUMN WITH THE SAME NAME
+          // todo also consider removing the field with the original name to avoid duplication
+          if (fieldName.equals("id")) {
+            fieldName = "database_id";
```
kwatters, line 158: We already have the "id" on the document because of the "idField" ..  we should skip additionally setting this "id" on the document...

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -168,7 +165,7 @@ public void execute(Publisher publisher) throws ConnectorException {
       log.info("Processing rows...");
       while (rs.next()) {
         // Need the ID column from the RS.
-        String id = createDocId(rs.getString(idColumn));
+        String id = idField == null ?  UUID.randomUUID().toString() : createDocId(rs.getString(idColumn));
```
rseitz, line 168: Random UUID will make it impossible to reindex into an existing index without creating duplicates. 
So this seems dangerous as a default behavior when idField isn't provided.
Could you tell me more about the situation where you needed to do this?

lucille-core/src/main/java/com/kmwllc/lucille/stage/DropIfFieldEquals.java
```diff
@@ -0,0 +1,55 @@
...
+import com.typesafe.config.Config;
+
+import java.util.Iterator;
+
+/**
+ * This stage will drop a document if it contains a given field and the value of that field matches a given value.
+ * Stage gives an option to do a case-sensitive comparison but is not case-sensitive by default.
+ * Stage also gives an option to invert the match, so that the document is dropped if the value does not match.
+ */
+public class DropIfFieldEquals extends Stage {
```
rseitz, line 16: Stages already support conditional execution. See the examples in the StageTest resources directory. I have included one example below.

I believe you can just make this stage drop every document it sees... but rely on conditions to prevent it from acting on any document that shouldn't be dropped.

For case-insensitivity, might need to use this together with NormalizeText to lowercase the field value... or investigate whether there's a way to add-case insensitivity as an option in the conditions framework.

conditions = [
  {
    fields = ["user_id", "state", "country"]
    values = ["MA", "1234", "US", "Russia", "4567"]
    operator = "must_not"
  }
]

PR #26, opened 2023-08-08
LC-320 add DatabaseConnector field handling

h0rban:
**Issue**
Currently, the database connector will silently fail when a Lucille document reserved field (such as id) encounters the same field name (column) in the database it is connecting to.

**Desired Solution**
Skip setting field value if the field name is called id

**Implemented Solution**
- explicit exception for reserved table column names
- introduction of the `ignoreColumns` configuration parameter

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
@@ -1,6 +1,7 @@
 package com.kmwllc.lucille.connector.jdbc;
 
 import com.kmwllc.lucille.connector.AbstractConnector;
+import com.kmwllc.lucille.core.ConfigUtils;
 import com.kmwllc.lucille.core.ConnectorException;
 import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.Publisher;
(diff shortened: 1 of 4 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
```diff
@@ -249,4 +249,74 @@ public void testClose() throws ConnectorException, SQLException {
...
+    configValues.put("idField", "other_id");
+
+    Config config = ConfigFactory.parseMap(configValues);
+    DatabaseConnector connector = new DatabaseConnector(config);
+
+    try {
+      connector.execute(publisher);
+    } catch (ConnectorException e) {
+      // expected
+      assertEquals("Field name \"id\" is reserved, please rename it or add it to the ignore list",
```
rseitz, line 272: This assertion won't be tested if no exception is thrown.
Better to use assertThrows here. See https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertThrows(java.lang.Class,%20org.junit.function.ThrowingRunnable)

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -72,6 +76,14 @@ public DatabaseConnector(Config config) {
       otherSQLs = new ArrayList<>();
       otherJoinFields = null;
     }
+    ignoreFields = new HashSet<>();
+    if (config.hasPath("ignoreFields")) {
```
rseitz, line 80: This is about ignoring specific _columns_ of the table before we convert them into document fields. So the name "ignoreColumns" would be more accurate.

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
h0rban: I like your suggestion. I think initially I was going to add a fail statement like this
```java
    try {
      connector.execute(publisher);
      fail("Should have thrown an exception");
    } catch (ConnectorException e) {
      // expected
      assertEquals("Field name \"id\" is reserved, please rename it or add it to the ignore list",
          e.getCause().getMessage());
    }
```
is this method also acceptable?

rseitz: The fail() approach works, but assertThrows is the recommended way and is a little more concise. Go with assertThrows unless there's a specific reason why fail() is better in your scenario.

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
```diff
@@ -277,4 +277,72 @@ public void testIdColumnException() {
...
+
+    // document id is coming from the "other_id" column
+    assertEquals("id1", doc1.getId());
+    assertEquals("id2", doc2.getId());
+
+    // "id" column is renamed to "table_id"
+    assertEquals("1", doc1.getString("table_id"));
+    assertEquals("2", doc2.getString("table_id"));
+
+    // "other_id" is not in the document
```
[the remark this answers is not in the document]
h0rban: updated the comment

Originally I wanted to ignore adding a field used for the id value since it creates duplication but leaving it as is was consistent with the previous implementation

rseitz (approved):

PR #27, opened 2023-08-09
LC-321 Created tests for StageUtils class

abijitran:
Found that the StageUtils.java class under utils in lucille-core was missing tests. This is not the only utils class missing tests, but created a draft of how tests might look like for these classes.

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/util/StageUtilsTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/util/StageUtilsTest.java
@@ -1,6 +1,5 @@
 package com.kmwllc.lucille.util;
 
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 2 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/util/StageUtilsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    String expectedMessage = "Unequal length field lists supplied to TestStage";
+    String actualMessage = exception.getMessage();
+
+    assertTrue(actualMessage.contains(expectedMessage));
+  }
+
+  @Test
+  public void testValidateFieldNumsSeveralToOne() {
```
rseitz, line 40: Is this testing validateFieldNumsSeveralToOne() ? Where is that method being invoked here?

abijitran: Mistake on my part, fixed this in new commit

lucille-core/src/test/java/com/kmwllc/lucille/util/StageUtilsTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    String expectedMessage = "TestStage was supplied with an invalid number of fields in the inputted field lists";
+
+    assertTrue(actualMessageFirst.contains(expectedMessage));
+    assertTrue(actualMessageSecond.contains(expectedMessage));
+  }
+
+  @Test
+  public void testValidateFieldNumNotZeroNoException() {
+    try {
```
rseitz, line 60: We don't need a try catch here or in the other new tests.
Just declare "public void testValidateFieldNumNotZeroNoException() throws StageException"

PR #29, opened 2023-08-14
added validation to both CreateStaticTeaser and DictionaryLookup stages

abijitran:


rseitz:
This approach is consistent with the way StageUtils validation methods are called elsewhere in the codebase, so I will approve this.
However, it brings up a larger issue which is that validation of config parameters happens in stage constructors and should not also be done in the start() method as that means it's happening in two places. The start() methods are for creating any expensive resources or connections that only be delayed until the time they are needed. 
The current StageUtils methods would be cumbersome to call from inside stage constructors, because they throw StageExceptions which would need to be caught and rethrown. Therefore, the StageUtils methods should be updated to throw IllegalArgumentExceptions (which is consistent with the exception type we use for other validation errors) and usage of these methods should be moved to constructors. That can be done in a separate ticket/PR.

rseitz (approved):

PR #31, opened 2023-08-15
LC-317 SolrIndexer should support connecting to Zookeeper Client

h0rban:
**Description**
Currently Lucille’s SolrIndexer supports connecting via direct URL to the Solr client. 
When Solr is running in cloud mode, connecting via zookeeper is a better option to support high availability. 
We would like to retain the ability to connect via direct URL but add the option to connect via zookeeper.

**Implementation Notes**
The Solr Indexer should support two additional config parameters “zkHosts” (includes zkChroot) and defaultCollection
If the useCloudClient solr indexer config property is set to “true” then the zkHosts string property should be required. 
The zkHosts and optional zkChroot properties should be used to create the CloudSolrClient used by the indexer.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
@@ -19,6 +19,7 @@ import org.apache.solr.client.solrj.io.Tuple;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * Utility methods for communicating with Solr.
@@ -39,9 +40,7 @@ public class SolrUtils {
   public static SolrClient getSolrClient(Config config) {
     SSLUtils.setSSLSystemProperties(config);
     if (config.hasPath("useCloudClient") && config.getBoolean("useCloudClient")) {
-      CloudSolrClient cloudSolrClient = requiresAuth(config) ?
-        new CloudSolrClient.Builder(getSolrUrls(config)).withHttpClient(getHttpClient(config)).build() :
-        new CloudSolrClient.Builder(getSolrUrls(config)).build();
+      CloudSolrClient cloudSolrClient = getCloudClient(config);
       if (config.hasPath("defaultCollection")) {
         cloudSolrClient.setDefaultCollection(config.getString("defaultCollection"));
       }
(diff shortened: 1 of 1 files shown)


h0rban:
The bottom of this section of the reference guide prompts to add a `solr-solrj-zookeeper` artifact, but I was able to verify indexing documents without it

lucille-core/src/test/resources/SolrIndexerTest/test.conf
rseitz, line 1: Where are test.conf, test2.conf, and test3.conf being used? 
If they are not being used and are only meant as examples, they shouldn't go in test resources. Could leave them out of this PR and we can find another place for them.
I'm also seeing absolute paths on you local machine starting with "/Users/h0rban" which shouldn't be checked in.

h0rban: definitely left them on purpose :) 
We could use a good place for example configuration files

rseitz (approved):

lucille-core/src/test/resources/SolrIndexerTest/test.conf
rseitz: I would put the new supported options in application-example.conf (see https://github.com/kmwtechnology/lucille/blob/main/application-example.conf)
They can be commented out and you can add comments that explain what they're for.

That's the easiest step. Going further we could add a module under lucille-examples that demonstrates how to configure a run that's indexing to Solr... but that's a larger task.

lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
```diff
@@ -53,6 +52,33 @@ public static SolrClient getSolrClient(Config config) {
     }
   }
 
+  private static CloudSolrClient getCloudClient(Config config) {
+    // todo review if zkHosts is required
```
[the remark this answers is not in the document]
rseitz: I think we should reflect what SolrJ supports.
Since it allows for creating a CloudSolrClient without zkHosts, it's OK for us to allow the same
It would be useful to have some INFO level logging about what kind of connection is being created and with what parameters -- but that could be left as a separate task.
Tagging @kyle-kmw to get his opinion too

lucille-core/src/test/resources/SolrIndexerTest/test.conf
h0rban: added examples

lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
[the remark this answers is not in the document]
kyle-kmw: I almost added a comment asking to force it to be required but I ultimately didn't with the same reasoning as @rseitz . Adding logging as a separate task (separate PR) makes sense to me. I am thinking that for the case where the indexer is configured with solr urls in cloud mode we should add a WARN level log entry that using zk based connections will be more reliable.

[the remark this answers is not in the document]
h0rban: removed the todo comment and added [LC-336](https://kmwllc.atlassian.net/browse/LC-336) to backlog

application-example.conf
```diff
@@ -46,10 +46,28 @@ elastic {
   sendEnabled: false
 }
 
+# Solr default mode
 solr {
   url: "http://localhost:8983/solr/callbacktest"
 }
 
+# Solr cloud mode with url
```
rseitz, line 54: Should comment out the second two solr examples so there's only one that's active
This conf file tries to be an example of a valid configuration that someone could actually use
The file has elastic, solr, and opensearch configs alongside each other, but that's OK because the one that's used is determined by the indexer type (which we're not specifying explicitly here -- this file needs a deeper review -- backlog ticket needed for that)
Wouldn't want to have multiple solr blocks in the same config though as they could collide -- so keep the first one uncommented and comment the other two.

h0rban: commented out and added a ticket [LC-338](https://kmwllc.atlassian.net/browse/LC-338)

PR #32, opened 2023-08-16
LC-17 modularize tika

abijitran:
Created module `lucille-tika' as part of lucille-plugins for text extraction from various document types. Supports both byte input and file path input. 

LC-17


lucille-plugins/lucille-tika/src/test/java/com/kmwllc/lucille/tika/stage/StageFactory.java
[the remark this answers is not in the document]
meghanboyd: Decison: Move stageFactory into source > main > java > util

This class is a utility that is only for use in testing. Normally this utility would live in test treee, but we are moving to source tree so that tests would be available to other modules.

abijitran:
Example of how a doc might look like in Solr / Opensearch (this is opensearch with a .txt file parsed by tika)
image

lucille-plugins/lucille-tika/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
+      </dependency>
+
+      <!-- tika -->
+      <dependency>
+        <groupId>org.apache.tika</groupId>
+              <artifactId>tika-core</artifactId>
+              <version>${tika.version}</version>
```
kyle-kmw, line 43: We should think about including the tika project's Bill of Material (BOM) in a dependency management for this module. That would (hopefully) make it less painful for someone downstream to pull this module into an existing codebase.

https://github.com/apache/tika/blob/main/README.md#maven-dependencies

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -0,0 +1,131 @@
...
+  private Parser parser;
+  private ParseContext parseCtx;
+
+  public TextExtractor(Config config) {
+    super(config, new StageSpec().withOptionalProperties("text_field", "file_path_field", "byte_array_field", "tika_config_path",
+        "metadata_prefix"));
+    textField = config.hasPath("text_field") ? config.getString("text_field") : "text";
+    filePathField = config.hasPath("file_path_field") ? config.getString("file_path_field") : "filepath";
+    byteArrayField = config.hasPath("byte_array_field") ? config.getString("byte_array_field") : "byte_array";
+    metadataPrefix = config.hasPath("metadata_prefix") ? config.getString("metadata_prefix") : "tika";
```
kyle-kmw, line 53: Thanks for sharing the example with the metadata @abijitran. I think we will very quickly find that we want to be able to 
  1. suppress certain metadata from being extracted with a whitelist or blacklist and also 
  2. to have an ability to rename certain metadata fields to allow users to make sure that tika doesn't produce fields that will result in schema mismatch errors. (I'm not sure if the mapping is something we would defer to other stage(s) or if we want the tika stage to be self-sufficient. @rseitz do you have an opinion on this?)

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -0,0 +1,131 @@
...
+    cleanName = cleanName.replaceAll(":", "_");
+    return cleanName;
+  }
+
+  /**
+   * Parses given input stream and adds the text data and metadata to given document
+   */
+  public void parseInputStream(Document doc, InputStream inputStream) {
+    Metadata metadata = new Metadata();
+    ContentHandler bch = new BodyContentHandler();
```
kyle-kmw, line 114: IIRC the BodyContentHandler has a constructor that takes a numeric value that allows you to limit the size of the content that it will produce. We should allow this value to be configurable.

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
abijitran: 1. Sure, will add this functionality soon
2. There exists a tika_prefix config property that should prevent schema mismatch errors (am going to support that to be empty so that users can have no prefix as well). Talking with Rudi, since RenameFields already handles renaming, we're thinking that it's best to leave that role to that stage.

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
abijitran: Changed this in new commit, should now take in config property to limit size (there is also a test for this as well now)

lucille-plugins/lucille-tika/pom.xml
abijitran: Done in new commit

lucille-plugins/lucille-tika/src/test/java/com/kmwllc/lucille/tika/stage/StageFactory.java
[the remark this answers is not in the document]
abijitran: Done, moved not under util but stage seeing that it requires to be in the same directory as stage

kyle-kmw (approved):

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -0,0 +1,158 @@
...
+  private List<String> metadataWhitelist;
+  private List<String> metadataBlacklist;
+  private Parser parser;
+  private ParseContext parseCtx;
+
+  public TextExtractor(Config config) throws StageException {
+    super(config, new StageSpec().withOptionalProperties("text_field", "file_path_field", "byte_array_field", "tika_config_path",
+        "metadata_prefix", "metadata_whitelist", "metadata_blacklist", "text_content_limit"));
+    textField = config.hasPath("text_field") ? config.getString("text_field") : "text";
+    filePathField = config.hasPath("file_path_field") ? config.getString("file_path_field") : "filepath";
```
rseitz, line 67: should the default value be "file_path" with the underscore considering the default byte array field is "byte_array"?

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -0,0 +1,158 @@
...
+    Metadata metadata = new Metadata();
+    ContentHandler bch = new BodyContentHandler(textContentLimit);
+    try {
+      parser.parse(inputStream, bch, metadata, parseCtx);
+    } catch (IOException | SAXException | TikaException e) {
+      log.warn("Tika Exception: {}", e.getMessage());
+    }
+
+    doc.addToField(textField, bch.toString());
+    parseCtx.toString();
```
rseitz, line 144: what is this doing?

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
abijitran: I agree, makes more sense with the underscore, change added in new commit

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
abijitran: Seems to be left behind from debugging, removed in new commit

rseitz (approved):

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -39,8 +39,9 @@
  * metadata_blacklist (StringList, Optional) : list of metadata names that are not to be included in document
  *
  * TODO
- * - add test case for empty metadata_prefix, if metadata_prefix is empty,
- *  do not include underscore i.e. "" -> content_type not _content_type
+ * - force the user to provide both a file_path_field and byte_array_field (no default value), throw error if they don't provide
```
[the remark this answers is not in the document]
abijitran: Yes, removed this in new commit

PR #34, opened 2023-08-24
LC-234 add timezone to ParseDate stage

h0rban:


rseitz:
These were the failures from the last build in September. Looks like it might just be a question of updating some test assertions to match the new behavior?

```
2023-09-13T19:16:32.9157760Z [ERROR] Failures: 
2023-09-13T19:16:32.9158880Z [ERROR]   ParseDateTest.test2:126 expected:<Tue Feb 02 00:00:00 [EST] 2021> but was:<Tue Feb 02 00:00:00 [UTC] 2021>
2023-09-13T19:16:32.9159520Z [ERROR]   ParseDateTest.test3:148 expected:<Tue Feb 02 0[3:00:00 EST] 2021> but was:<Tue Feb 02 0[8:00:00 UTC] 2021>
2023-09-13T19:16:32.9160400Z [ERROR]   ParseDateTest.test4:170 expected:<Mon Feb 01 [18:00:00 EST] 2021> but was:<Mon Feb 01 [23:00:00 UTC] 2021>
2023-09-13T19:16:32.9160770Z [INFO] 
2023-09-13T19:16:32.9161080Z [ERROR] Tests run: 501, Failures: 3, Errors: 0, Skipped: 2
```

lucille-core/src/test/java/com/kmwllc/lucille/stage/ParseDateTest.java
```diff
@@ -52,7 +59,119 @@ public void testGetLegalProperties() throws StageException {
...
+
+      // parse date and confirm it matches the expected formatted date
+      stage.processDocument(doc);
+      assertEquals("Failed for time_zone_id=" + timeZoneId,
+          entry.getValue(), doc.getString("output"));
+    }
+  }
+
+  @Test
+  public void test2() {
```
h0rban, line 116: Tests 2, 3, and 4 demonstrate how the data is modified in the stage with three different time zones.

Thoughts:
1. we should not only allow the start of day dates but correct time too. maybe add a separate ParseInstant stage
3. should we be able to specify the output timezone too?

lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseDate.java
```diff
@@ -40,20 +37,48 @@ public class ParseDate extends Stage {
...
+      throw new IllegalArgumentException("Invalid time zone ID: \"" + timeZoneId
+          + "\", must be one of TimeZone.getAvailableIDs()");
+    }
+    return timeZoneId;
+  }
+
+  private List<DateFormat> getFormats(Config config) {
+
+    // convert format strings to a set to remove duplicates
+    Set<String> formatStrings = new HashSet<>(ConfigUtils.getOrDefault(config, "format_strs", new ArrayList<>()));
```
rseitz, line 71: We should preserve the order of the formatStrings. You could do this by using a LinkedHashSet instead of a HashSet here, but I would recommend `strList.stream().distinct().collect(Collectors.toList())` instead

lucille-core/src/test/java/com/kmwllc/lucille/stage/ParseDateTest.java
```diff
@@ -52,7 +59,119 @@ public void testGetLegalProperties() throws StageException {
...
+
+  @Test
+  public void testParseDateWithTimeZone() throws StageException {
+
+    // create a test config without a time zone
+    HashMap<String,Object> configValues = new HashMap<>();
+    configValues.put("class", "com.kmwllc.lucille.stage.ParseDate");
+    configValues.put("source", List.of("date"));
+    configValues.put("dest", List.of("output"));
+    configValues.put("format_strs", List.of("yyyy-MM-dd"));
```
rseitz, line 83: We should test two cases:

- if we the date we're parsing doesn't include zone information, then time_zone_id is used
- if the date we're parsing does include zone information, then that info is used instead of time_zone_id

This test could be updated to use two format_strs like this:

`    configValues.put("format_strs", List.of("yyyy-MM-dd z", "yyyy-MM-dd"));`

When running the Stage, we should try it on two different dates:

- "2021-02-02"
- "2021-02-02 EST"

and make sure we get back the right date (in Zulu format) in each case

this could also be split into a separate test

lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseDate.java
```diff
@@ -40,20 +37,48 @@ public class ParseDate extends Stage {
   private final List<String> sourceFields;
   private final List<String> destFields;
   private final UpdateMode updateMode;
+  private final String timeZoneId;
+
+  private static final String DEFAULT_TIME_ZONE_ID = Calendar.getInstance().getTimeZone().getID();
```
rseitz, line 42: I'd call this "localTimeZoneId" instead of "DEFAULT_TIME_ZONE_ID"
reasons:

1. yes, it's the default for the system where lucille is being run at the moment, but it's not a global, unchanging default -- same lucille distribution could have different values here depending on what server it's executed on
2. yes, it's a private static final String, but Calendar.getInstance().getTimeZone().getID() isn't a "constant expression" as described above, so I think this is a scenario where the uppercase convention shouldn't apply

lucille-core/src/test/java/com/kmwllc/lucille/stage/ParseDateTest.java
```diff
@@ -52,7 +59,119 @@ public void testGetLegalProperties() throws StageException {
...
+      doc.setField("date", "2021-02-02");
+
+      // parse date and confirm it matches the expected formatted date
+      stage.processDocument(doc);
+      assertEquals("Failed for time_zone_id=" + timeZoneId,
+          entry.getValue(), doc.getString("output"));
+    }
+  }
+
+//  @Test
```
rseitz, line 115: we'd like to keep commented-out code to a minimum

if this is here to explain something important, leave it, but please:

- remove the enclosing method (i.e. public void test2()) because that makes it look like an actual test that we disabled for some reason
- add a few comments saying why this is here and what it's showing

lucille-core/src/test/java/com/kmwllc/lucille/stage/ParseDateTest.java
rseitz: created a ticket here: https://kmwllc.atlassian.net/browse/LC-381

rseitz (approved):

PR #35, opened 2023-08-24
LC-342 forbid null field names on document

h0rban:
I noticed that JsonDocument implementation does not rely on the `validateNotReservedField` function in the Document interface, so this in this PR I:
- added an exception for an empty string as a field name
- renamed `validateNotReservedField` to `validateFieldNames` to represent a wider range of the function's logic
- removed a validation method from `JsonDocument` and instead made it rely on the interface function
- added a test for all methods that rely on the validation of the field name


lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -257,13 +257,13 @@ static Document createFromJson(String json, UnaryOperator<String> idUpdater)
     return JsonDocument.fromJsonString(json, idUpdater);
   }
 
-  static void validateNotReservedField(String... names) throws IllegalArgumentException {
+  static void validateFieldNames(String... names) throws IllegalArgumentException {
```
rseitz, line 260: I think we should make this method "default" rather than "static". That way 1) we don't have to statically import it from each document implementation, 2) document implementations would be able to override it if necessary.

rseitz (approved):

PR #37, opened 2023-08-30
LC-347 fix dependency conflict that prevents lucille run

rseitz:
https://kmwllc.atlassian.net/browse/LC-347


abijitran (approved):

kyle-kmw (approved):

kyle-kmw:
Oops! That is definitely supposed to be test scoped. Thanks @rseitz .

PR #38, opened 2023-09-01
LC-316 Cleaned up connection for result sets in database connector

abijitran:
LC-316
Cleans up the connection by moving the code into a finally block

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
@@ -120,7 +120,8 @@ public class DatabaseConnector extends AbstractConnector {
   public void execute(Publisher publisher) throws ConnectorException {
     ResultSet rs = null;
     ArrayList<ResultSet> otherResults = null;
-    try (Connection connection = createConnection()) {
+    try {
+      Connection connection = createConnection();
       // run the pre-sql (if specified)
       runSql(connection, preSql);
       // TODO: make sure we cleanup result set/statement/connections properly.
--- a/lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
@@ -121,7 +121,7 @@ public class DatabaseConnector extends AbstractConnector {
     ResultSet rs = null;
     ArrayList<ResultSet> otherResults = null;
     Connection connection = null;
-    Statement state = null;
+    Statement statement = null;
     try {
       connection = createConnection();
       // run the pre-sql (if specified)
@@ -132,11 +132,11 @@ public class DatabaseConnector extends AbstractConnector {
 
       try {
         // state is closed at the end as it is still being used while gathering result sets
-        state = connection.createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
+        statement = connection.createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
         if (fetchSize != null) {
-          state.setFetchSize(fetchSize);
+          statement.setFetchSize(fetchSize);
         }
-        rs = state.executeQuery(sql);
+        rs = statement.executeQuery(sql);
       } catch (SQLException e) {
         throw e;
       }
@@ -225,9 +225,9 @@ public class DatabaseConnector extends AbstractConnector {
           }
         }
       }
-      if (state != null) {
+      if (statement != null) {
         try {
-          state.close();
+          statement.close();
         } catch (SQLException e) {
           log.warn("Unable to close Statement", e);
         }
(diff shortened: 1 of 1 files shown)


abijitran:
testClose test in DataBaseConnectorTest seems to fail when I move the connection creation line into a try-with-resources, need to look into this

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -199,16 +200,29 @@ public void execute(Publisher publisher) throws ConnectorException {
...
       }
 
-      // close all results
-      rs.close();
-      for (ResultSet ors : otherResults) {
-        ors.close();
-      }
       // the post sql.
       runSql(connection, postSql);
       flush();
```
rseitz, line 205: The flush() method definition and this reference can be removed. flush() is a private method with an empty body -- it is not serving a purpose

abijitran: Done

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -199,16 +200,29 @@ public void execute(Publisher publisher) throws ConnectorException {
         publisher.publish(doc);
       }
 
-      // close all results
-      rs.close();
-      for (ResultSet ors : otherResults) {
-        ors.close();
-      }
       // the post sql.
       runSql(connection, postSql);
```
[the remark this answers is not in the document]
abijitran: I moved the runSql call to after the closing of the ResultSets, but I didn't surround it with a try catch because the SQLException is already caught in runSql. Let me know if I'm misunderstanding it.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -199,24 +201,30 @@ public void execute(Publisher publisher) throws ConnectorException {
...
+        try {
+          rs.close();
+        } catch (SQLException e) {
+          log.warn("Result set unable to be closed");
+        }
+      }
+      if (otherResults != null) {
+        for (ResultSet ors : otherResults) {
+          try {
+            ors.close();
```
kyle-kmw, line 219: need to null check each ors before calling close

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -118,15 +118,17 @@ private int getIdColumnIndex(String[] columns) throws ConnectorException {
...
   @Override
   public void execute(Publisher publisher) throws ConnectorException {
-
+    ResultSet rs = null;
+    ArrayList<ResultSet> otherResults = null;
+    Connection connection = null;
     try {
-      // connect to the database.
-      Connection connection = createConnection();
+      connection = createConnection();
```
kyle-kmw, line 125: where is the connection closed? Am I missing it? Can we declare the connection within the above try? i.e. 
```
try(Connection connection = createConnection) {
...
```

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -118,15 +118,17 @@ private int getIdColumnIndex(String[] columns) throws ConnectorException {
...
       // run the pre-sql (if specified)
       runSql(connection, preSql);
       // TODO: make sure we cleanup result set/statement/connections properly.
-      ResultSet rs;
+
       log.info("Running primary sql");
+
       try {
         Statement state = connection.createStatement(ResultSet.TYPE_FORWARD_ONLY, ResultSet.CONCUR_READ_ONLY);
```
kyle-kmw, line 133: statement is never closed

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
kyle-kmw: Ahh okay I thought the connection was declared within the method but now I see that is removed code. Seems good then.

abijitran: I think Rudi commented already, but it's closed in the .close() method the number of connections created is not known until we collect them all into the globally defined array.

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
abijitran: Closed in new commit, didn't do it in a try-with-resources because the statement is being used outside of the scope of the try-catch

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
abijitran: Added null checking in new commit

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -118,17 +118,21 @@ private int getIdColumnIndex(String[] columns) throws ConnectorException {
 
   @Override
   public void execute(Publisher publisher) throws ConnectorException {
-
+    ResultSet rs = null;
+    ArrayList<ResultSet> otherResults = null;
+    Connection connection = null;
+    Statement state = null;
```
rseitz, line 124: While we're in this code, can we rename "state" to "statement" for clarity?

rseitz (approved):
approving, but left a minor request to rename "state" variable to "statement"

PR #41, opened 2023-09-08
Changed existing Rest High Level Client in OpenSearchIndexer to Java Client

abijitran:
LC-335
Changed from RestHighLevelClient to JavaClient for OpenSearch in light of the fact that OpenSearch 3.0.0 will no longer support RestHighLevelClient.

### DIFF
--- a/lucille-core/pom.xml
+++ b/lucille-core/pom.xml
@@ -19,15 +19,9 @@
   <dependencies>
     <dependency>
       <groupId>org.opensearch.client</groupId>
-      <artifactId>opensearch-rest-client</artifactId>
-      <version>2.9.0</version>
-    </dependency>
-    <dependency>
-      <groupId>org.opensearch.client</groupId>
       <artifactId>opensearch-java</artifactId>
       <version>2.6.0</version>
     </dependency>
-
     <dependency>
       <groupId>co.elastic.clients</groupId>
       <artifactId>elasticsearch-java</artifactId>
--- a/lucille-core/pom.xml
+++ b/lucille-core/pom.xml
@@ -17,13 +17,6 @@
   <name>Lucille Core</name>
 
   <dependencies>
-
-<!--    <dependency>-->
-<!--      <groupId>org.opensearch.client</groupId>-->
-<!--      <artifactId>opensearch-rest-high-level-client</artifactId>-->
-<!--      <version>${opensearch-rest-high-level-client.version}</version>-->
-<!--    </dependency>-->
-
     <dependency>
       <groupId>org.opensearch.client</groupId>
       <artifactId>opensearch-rest-client</artifactId>
--- a/lucille-core/pom.xml
+++ b/lucille-core/pom.xml
@@ -18,10 +18,21 @@
 
   <dependencies>
 
+<!--    <dependency>-->
+<!--      <groupId>org.opensearch.client</groupId>-->
+<!--      <artifactId>opensearch-rest-high-level-client</artifactId>-->
+<!--      <version>${opensearch-rest-high-level-client.version}</version>-->
+<!--    </dependency>-->
+
     <dependency>
       <groupId>org.opensearch.client</groupId>
-      <artifactId>opensearch-rest-high-level-client</artifactId>
-      <version>${opensearch-rest-high-level-client.version}</version>
+      <artifactId>opensearch-rest-client</artifactId>
+      <version>2.9.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.opensearch.client</groupId>
+      <artifactId>opensearch-java</artifactId>
+      <version>2.6.0</version>
     </dependency>
 
     <dependency>
--- a/lucille-core/pom.xml
+++ b/lucille-core/pom.xml
@@ -23,6 +23,16 @@
       <version>2.6.0</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.httpcomponents.core5</groupId>
+      <artifactId>httpcore5</artifactId>
+      <version>5.1.5</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5</artifactId>
+      <version>5.1.4</version>
+    </dependency>
+    <dependency>
       <groupId>co.elastic.clients</groupId>
       <artifactId>elasticsearch-java</artifactId>
       <version>8.6.0</version>
--- a/lucille-core/pom.xml
+++ b/lucille-core/pom.xml
@@ -20,17 +20,17 @@
     <dependency>
       <groupId>org.opensearch.client</groupId>
       <artifactId>opensearch-java</artifactId>
-      <version>2.6.0</version>
+      <version>${opensearch-java.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents.core5</groupId>
       <artifactId>httpcore5</artifactId>
-      <version>5.1.5</version>
+      <version>${httpcore5.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents.client5</groupId>
       <artifactId>httpclient5</artifactId>
-      <version>5.1.4</version>
+      <version>${httpclient5.version}</version>
     </dependency>
     <dependency>
       <groupId>co.elastic.clients</groupId>
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
@@ -138,7 +138,7 @@ public class OpenSearchIndexer extends Indexer {
       }
     }
     BulkResponse response = client.bulk(br.build());
-    if (response.errors()) {
+    if (response != null && response.errors()) {
       for (BulkResponseItem item : response.items()) {
         log.error("OpenSearchIndexer response has received error(s)", item.error().reason());
       }
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
@@ -3,6 +3,7 @@ package com.kmwllc.lucille.indexer;
 import com.kmwllc.lucille.core.ConfigUtils;
 import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.Indexer;
+import com.kmwllc.lucille.core.IndexerException;
 import com.kmwllc.lucille.core.KafkaDocument;
 import com.kmwllc.lucille.message.IndexerMessageManager;
 import com.kmwllc.lucille.message.KafkaIndexerMessageManager;
@@ -141,6 +142,7 @@ public class OpenSearchIndexer extends Indexer {
     if (response != null && response.errors()) {
       for (BulkResponseItem item : response.items()) {
         log.error("OpenSearchIndexer response has received error(s)", item.error().reason());
+        throw new IndexerException(item.error().reason());
       }
     }
   }
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
@@ -145,9 +145,7 @@ public class OpenSearchIndexer extends Indexer {
     BulkResponse response = client.bulk(br.build());
     if (response != null && response.errors()) {
       for (BulkResponseItem item : response.items()) {
-        String reason = item.error().reason();
-        log.error("OpenSearchIndexer response has received error(s)", reason);
-        throw new IndexerException(reason);
+        throw new IndexerException(item.error().reason());
       }
     }
   }
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
@@ -3,6 +3,7 @@ package com.kmwllc.lucille.indexer;
 import com.kmwllc.lucille.core.ConfigUtils;
 import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.Indexer;
+import com.kmwllc.lucille.core.KafkaDocument;
 import com.kmwllc.lucille.message.IndexerMessageManager;
 import com.kmwllc.lucille.message.KafkaIndexerMessageManager;
 import com.kmwllc.lucille.util.OpenSearchUtils;
@@ -12,6 +13,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.client.opensearch._types.VersionType;
 import org.opensearch.client.opensearch.core.BulkRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -26,6 +28,8 @@ public class OpenSearchIndexer extends Indexer {
 
   private final String routingField;
 
+  private final VersionType versionType;
+
   //flag for using partial update API when sending documents to opensearch
   private final boolean update;
 
@@ -39,6 +43,8 @@ public class OpenSearchIndexer extends Indexer {
     this.index = OpenSearchUtils.getOpenSearchIndex(config);
     this.routingField = config.hasPath("indexer.routingField") ? config.getString("indexer.routingField") : null;
     this.update = config.hasPath("opensearch.update") ? config.getBoolean("opensearch.update") : false;
+    this.versionType =
+        config.hasPath("indexer.versionType") ? VersionType.valueOf(config.getString("indexer.versionType")) : null;
   }
 
   public OpenSearchIndexer(Config config, IndexerMessageManager manager, boolean bypass, String metricsPrefix) {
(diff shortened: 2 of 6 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -92,47 +92,50 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
-
-      if (versionType != null) {
-        indexRequest.versionType(versionType);
-        if (versionType == VersionType.EXTERNAL || versionType == VersionType.EXTERNAL_GTE) {
-          // the partition doesn’t need to be included in the version. We assume the doc id is used as the
-          // kafka message key, which should guarantee that all “versions” of a given document have the
-          // same kafka message key and therefore get mapped to the same topic partition.
-          indexRequest.version(((KafkaDocument) doc).getOffset());
-        }
+      // seems as though passing null to .routing(), .version(), or .versionType() will not affect the indexing process
```
rseitz, line 110: re: "seems as though" -- how sure of this are we and what has been done to confirm it?

abijitran: Changed the implementation to be explicit on whether it is adding versioning or routing

lucille-core/pom.xml
```diff
@@ -17,11 +17,15 @@
   <name>Lucille Core</name>
 
   <dependencies>
-
     <dependency>
       <groupId>org.opensearch.client</groupId>
-      <artifactId>opensearch-rest-high-level-client</artifactId>
-      <version>${opensearch-rest-high-level-client.version}</version>
+      <artifactId>opensearch-rest-client</artifactId>
+      <version>2.9.0</version>
```
rseitz, line 23: lucille-core/pom.xml has a mix of hardcoded versions with variables defined in lucille-parent/pom.xml
If we're hardcoding 2.9.0 here and not using ${opensearch-rest-high-level-client.version} we should remove ${opensearch-rest-high-level-client.version} from lucille-parent/pom.xml
But before we do that, perhaps @kyle-kmw could comment on when versions should be specified in lucille-core/pom.xml versus in lucille-parent/pom.xml

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -92,47 +99,50 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
-      indexRequest.source(indexerDoc);
-
-      if (versionType != null) {
-        indexRequest.versionType(versionType);
-        if (versionType == VersionType.EXTERNAL || versionType == VersionType.EXTERNAL_GTE) {
-          // the partition doesn’t need to be included in the version. We assume the doc id is used as the
-          // kafka message key, which should guarantee that all “versions” of a given document have the
-          // same kafka message key and therefore get mapped to the same topic partition.
-          indexRequest.version(((KafkaDocument) doc).getOffset());
+      Function<UpdateOperation.Builder, ObjectBuilder<UpdateOperation>> fn = new Function<Builder, ObjectBuilder<UpdateOperation>>() {
```
rseitz, line 117: where is this used?

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -92,47 +99,50 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
-
-      // add indexRequest to bulkRequest
-      bulkRequest.add(indexRequest);
-    }
-
-    BulkResponse response = client.bulk(bulkRequest, RequestOptions.DEFAULT);
-    if (response.hasFailures()) {
-      log.error(response.buildFailureMessage());
     }
+    client.bulk(br.build());
```
rseitz, line 145: why are we not looking at the response and checking for errors?

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
```diff
@@ -56,16 +58,17 @@ public static RestHighLevelClient getOpenSearchRestClient(Config config) {
...
           @Override
           public HttpAsyncClientBuilder customizeHttpClient(HttpAsyncClientBuilder httpClientBuilder) {
             return httpClientBuilder.setDefaultCredentialsProvider(provider).setSSLContext(sslFactory.getSslContext())
                 .setSSLHostnameVerifier(sslFactory.getHostnameVerifier());
           }
-        });
+        }).build();
 
-    RestHighLevelClient client = new RestHighLevelClient(builder);
+    final OpenSearchTransport transport = new RestClientTransport(restClient, new JacksonJsonpMapper());
```
rseitz, line 70: Is there a reason why we need to use RestClientTransport here instead of the default ApacheHttpClient5Transport?
https://opensearch.org/docs/latest/clients/java/

if so, should add a code comment with brief explanation

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
abijitran: Not used, removed this in new commit

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -92,46 +101,47 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
-      bulkRequest.add(indexRequest);
     }
-
-    BulkResponse response = client.bulk(bulkRequest, RequestOptions.DEFAULT);
-    if (response.hasFailures()) {
-      log.error(response.buildFailureMessage());
+    BulkResponse response = client.bulk(br.build());
+    if (response.errors()) {
+      for (BulkResponseItem item : response.items()) {
+        log.error("OpenSearchIndexer response has received error(s)", item.error().reason());
```
kyle-kmw, line 143: We should discuss what is the desired behavior here @rseitz, what do you think? 

It seems like if we have an error here we should at least throw an IndexerException so that the batch is failed. Otherwise, we are recording success events for documents that weren't indexed. We aren't doing this in the current implementation of the ElasticIndexer or OpensearchIndexer but I believe that is a gap. 

Beyond that, can we take advantage of the granularity that the bulk API provides so as to SUCCEED and FAIL the correct documents within the batch? Support for that would probably have to be implemented at the base Indexer level.

rseitz: Thx for bringing this up, I was thinking about it too. An indexer subclass is supposed to fail the whole batch if any doc in that batch has an error. But OpenSearchIndexer and ElasticsearchIndexer are not doing that. I think they should both be updated to inspect the BulkResponse and throw an IndexerException if any error is encountered. I'm trying to remember if there was a reason we didn't do that during the initial implementations of these indexers, but I can't see a justification. SolrIndexer should be considered too -- solrClient.add() returns an UpdateResponse that we're currently ignoring, but as far as I can see there's nothing in there and, in the solr case, we can that any errors would be indicated as exceptions thrown by add().

It would be nice to take advantage of the granularity of the bulk API, it would be a good design ticket for the backlog, but I think we should address the immediate gap first.

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
abijitran: Changed this in new commit

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
abijitran: The following forum post suggests that the RestClientTransport will be deprecated with OpenSearch 3.0. It also has HTTP/2.0 (with OpenSearch 3.0) which is preferable. Will push implementation with ApacheClient soon. 

https://forum.opensearch.org/t/difference-between-apache-httpclient-5-transport-and-restclient-transport/14650

lucille-core/pom.xml
abijitran: Done

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
abijitran: Changes relevant to this have been pushed

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
kyle-kmw: No. Logging + Throwing makes the application log more difficult to read, usually it results in the same error logged twice. It is better to just throw in this case and let the upstream caller handle logging. Just make sure that the IndexerException has the item errors passed into it such that when the exception is handled upstream it will include the information.

abijitran: Included in new commit

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -92,46 +96,57 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
-
-      // add indexRequest to bulkRequest
-      bulkRequest.add(indexRequest);
     }
-
-    BulkResponse response = client.bulk(bulkRequest, RequestOptions.DEFAULT);
-    if (response.hasFailures()) {
-      log.error(response.buildFailureMessage());
+    BulkResponse response = client.bulk(br.build());
+    if (response != null && response.errors()) {
```
rseitz, line 146: we should check that response.errors() is reliable
it might be sufficient to not look at response.errors() but simply iterate through the items and look for any item with an error

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -92,46 +96,57 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
+      for (BulkResponseItem item : response.items()) {
+        throw new IndexerException(item.error().reason());
```
rseitz, line 148: we only want to throw an exception for the first item that has an error
so the throw should be inside a conditional that looks for the presence of an error

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
```diff
@@ -1,73 +1,110 @@
...
+              }
+            }
 
-    return client;
+            // Potentially disable SSL/TLS verification for when testing locally
+            boolean allowInvalidCert = getAllowInvalidCert(config);
+            TlsStrategy tlsStrategy = null;
+            SSLContext sslContext = null;
+            try {
+            if (allowInvalidCert) {
```
rseitz, line 72: indentation

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
```diff
@@ -1,73 +1,110 @@
...
+              tlsStrategy = ClientTlsStrategyBuilder.create()
+                  .setSslContext(sslContext)
+                  .setHostnameVerifier(NoopHostnameVerifier.INSTANCE)
+                  .build();
+            } else {
+              sslContext = SSLContextBuilder.create()
+                    .build();
+              tlsStrategy = ClientTlsStrategyBuilder.create()
+                  .setSslContext(sslContext)
+                  //.setHostnameVerifier(new DefaultHostnameVerifier())
```
rseitz, line 86: remove comment

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
```diff
@@ -1,73 +1,110 @@
...
-    // get user info from URI if present and setup BasicAuth credentials if needed
-    String userInfo = hostUri.getUserInfo();
-    if (userInfo != null) {
-      int pos = userInfo.indexOf(":");
-      String username = userInfo.substring(0, pos);
-      String password = userInfo.substring(pos + 1);
-      provider.setCredentials(AuthScope.ANY,
-          new UsernamePasswordCredentials(username, password));
-    }
+  public static OpenSearchClient getOpenSearchRestClientInternal(Config config) {
```
rseitz, line 38: please add a comment with a link to the place where we adapted the SSL setup from (the example repo you found)
please mention anything that's different about our code from the example provided for opensearch

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
abijitran: Here's javadoc for BulkResponseItem.errors(): https://javadoc.io/doc/org.opensearch.client/opensearch-java/latest/org/opensearch/client/opensearch/core/BulkResponse.html#items()

Looking at the code for errors in BulkResponse, it seems to be built in a builder the same way as .items() is. Is there something specific I should be looking for to validate whether it is accurate to use .errors()?

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
abijitran: Done

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
abijitran: Done

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
abijitran: Done

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
abijitran: Done

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
abijitran: Addressed

lucille-core/src/test/java/com/kmwllc/lucille/indexer/OpenSearchIndexerTest.java
```diff
@@ -9,46 +13,44 @@
...
-import static org.mockito.Mockito.verify;
+import org.opensearch.client.opensearch.OpenSearchClient;
+import org.opensearch.client.opensearch._types.VersionType;
+import org.opensearch.client.opensearch.core.BulkRequest;
+import org.opensearch.client.opensearch.core.BulkResponse;
+import org.opensearch.client.opensearch.core.bulk.BulkOperation;
+import org.opensearch.client.opensearch.core.bulk.IndexOperation;
+import org.opensearch.client.transport.endpoints.BooleanResponse;
 
 public class OpenSearchIndexerTest {
```
abijitran, line 33: Add tests for the changes to error handling in OpenSearchIndexer.java

abijitran: Done

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -92,46 +96,60 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
-    BulkRequest bulkRequest = new BulkRequest(index);
+    BulkRequest.Builder br = new BulkRequest.Builder();
 
-    // determine what field to use as id field and iterate over the documents
     for (Document doc : documents) {
-      Map<String, Object> indexerDoc = getIndexerDoc(doc);
+      Map<String, Object> indexerDoc = doc.asMap();
 
       // remove children documents field from indexer doc (processed from doc by addChildren method call below)
       indexerDoc.remove(Document.CHILDREN_FIELD);
```
rseitz, line 105: i know this PR didn't change this line, but I'm wondering: why is this being done separately from the addChildren call below? i.e. why aren't those two lines grouped?

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
```diff
@@ -1,73 +1,109 @@
...
+
+            return httpClientBuilder
+                .setDefaultCredentialsProvider(credentialsProvider)
+                .setConnectionManager(connectionManager);
+          })
+          .build();
+
+      return new OpenSearchClient(transport);
+    } catch (Exception e) {
+      System.out.println("failed to make transport client");
```
rseitz, line 104: remove println; add log statement

rseitz (approved):
approving, have given the most attention to the bulk logic and tests; have reviewed the ssl stuff but not as extensively -- would be great if @kyle-kmw could take a look as well

lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
abijitran: Done

kyle-kmw (approved):

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
abijitran: Ticket was created

PR #43, opened 2023-09-13
LC-341 support join fields in ElasticsearchIndexer

h0rban:


lucille-core/src/main/java/com/kmwllc/lucille/util/ElasticsearchUtils.java
```diff
@@ -125,4 +129,79 @@ public static boolean getAllowInvalidCert(Config config) {
     }
     return false;
   }
+
+  public static class ElasticJoinData {
```
rseitz, line 133: should move to a top-level class
since this is only used by ElasticSearchIndexer it can go in the indexer package itself

ElasticsearchUtils is meant to be a holder of static utility methods relating to elastic, but not a place to nest inner classes

lucille-core/src/main/java/com/kmwllc/lucille/util/ElasticsearchUtils.java
```diff
@@ -125,4 +129,79 @@ public static boolean getAllowInvalidCert(Config config) {
...
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    private String joinFieldName;
+    private boolean isChild;
+    private String parentName;
+    private String childName;
+    private String parentDocumentIdSource;
+
+    public void addJoinData(Document doc) {
```
rseitz, line 143: a method starting with "add" usually adds the argument to the object that the method is called on
in this case, we're copying from the object onto the argument
suggest renaming this populateJoinData() or copyJoinDataToDocument()

rseitz (approved):

PR #44, opened 2023-09-13
LC-340 support routing in ElasticsearchIndexer

h0rban:


lucille-core/src/test/resources/ElasticsearchIndexerTest/routing.conf
```diff
@@ -0,0 +1,21 @@
+indexer {
+  type: "Elasticsearch"
+  routingField: "routing"
+
+  # todo whats the point of having this list?
```
rseitz, line 5: this tells the indexer to not send the id and routing fields to elasticsearch as part of the document
these fields aren't needed as document fields in ES because ES captures them elsewhere
the test should work fine without this setting
however, it is arguably a good idea to leave this setting in place in the test to make sure that routing info is being set even when "routing" is in the list of ignoreFields

where possible we should avoid leaving TODOs/questions in the code so instead please instead leave a comment explaining that this is for

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
```diff
@@ -91,24 +95,27 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
 
+      // we don't have to worry about routing being null or not in the document since the indexer will ignore it
+      String routing = doc.getString(routingField);
+
       if (update) {
         br.operations(op -> op
             .update(up -> up
                 .id(docId)
                 .index(index)
+                .routing(routing)
```
rseitz, line 106: abi has been working on the equivalent code in OpenSearchIndexer
we couldn't find documentation saying that .routing(null) would be ignored, so we decided to be on the safe side and only set the routing when non-null
along with routing, there is also version info that we should set if versioning is configured
ideally the ElasticsearchIndexer and OpenSearchIndexers should be kept as close as possible, so let's compare and reconcile
here is the current code in abi's PR: https://github.com/kmwtechnology/lucille/blob/9bff23eeafa78c00431b747fd3d83b97a864ca01/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java#L124

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
```diff
@@ -144,6 +174,22 @@ private void addChildren(Document doc, Map<String, Object> indexerDoc) {
     }
   }
 
+  private static VersionType versionTypeFromString(String versionType) {
```
rseitz, line 177: VersionType is an enum so we shouldn't have to write this method
can do VersionType.valueOf(str) but that would require the str matches the enum value exactly, so it could be "Internal", "External", "ExternalGte", or "Force"
Do we need to allow the user to enter "internal" "external" "external_get" or "force" ?
If so, we could support it like this:
VersionType.valueOf(CaseUtils.toCamelCase("external_gte", true, '_'))

see https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/CaseUtils.html

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
```diff
@@ -40,6 +44,10 @@ public ElasticsearchIndexer(Config config, IndexerMessageManager manager, Elasti
     this.client = client;
     this.index = ElasticsearchUtils.getElasticsearchIndex(config);
     this.update = config.hasPath("elasticsearch.update") ? config.getBoolean("elasticsearch.update") : false;
+
+    // todo consider making routing a Document class member
```
rseitz, line 48: pls remove the TODO comment from code -- we have a ticket to track this now

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
rseitz: abi's OpenSearch PR is using VersionType.valueOf without the CaseUtils part
unless there's a known reason why we need to support the version type names in snake case, let's just do VersionType.valueOf() and update any configs to use "Internal", "External", "ExternalGte", or "Force"
See https://github.com/kmwtechnology/lucille/blob/15e83f3375a48b3e9a2eb2bd4c2e41bc6ba95b95/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java#L51

rseitz (approved):

PR #45, opened 2023-09-18
Lc 350 Simple CSV to Solr Example

abijitran:


lucille-examples/lucille-simple-csv-solr-example/conf/simple-csv-solr-example.conf
```diff
@@ -0,0 +1,30 @@
+# this example illustrates how we can use Lucille to take something as simple as a CSV and index it into a search engine like Solr
```
rseitz, line 1: this example illustrates how Lucille can handle a simple use case like indexing the contents of a CSV file into Solr

rseitz (approved):
approving this now so merging is not blocked if I'm OOO
have left some comments which should be addressed before merging though

lucille-examples/lucille-simple-csv-solr-example/conf/simple-csv-solr-example.conf
abijitran: Done

PR #48, opened 2023-09-21
LC-362 Changed how validate connection works for HTTPSolrClient versus CloudSolrClient

abijitran:
Saw issues with HTTPSolrClient not correctly being validated, changed it so that it is now using a .ping() for httpclient and the old method for the Cloud client.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
@@ -10,7 +10,10 @@ import com.kmwllc.lucille.util.SolrUtils;
 import com.typesafe.config.Config;
 import org.apache.solr.client.solrj.SolrClient;
 import org.apache.solr.client.solrj.SolrServerException;
+import org.apache.solr.client.solrj.impl.CloudSolrClient;
+import org.apache.solr.client.solrj.impl.HttpSolrClient;
 import org.apache.solr.client.solrj.request.CollectionAdminRequest;
+import org.apache.solr.client.solrj.response.SolrPingResponse;
 import org.apache.solr.common.SolrInputDocument;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.SimpleOrderedMap;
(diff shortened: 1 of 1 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -53,28 +56,46 @@ public boolean validateConnection() {
...
-    Integer status = (Integer) ((SimpleOrderedMap) response.get("responseHeader")).get("status");
-    if (status != 0) {
-      log.error("Non zero response when checking solr cluster status: " + status);
-      return false;
+    if (solrClient instanceof HttpSolrClient) {
+      try {
+        SolrPingResponse resp = solrClient.ping();
+        int status = resp.getStatus();
+        if (status != 0) {
+          log.error("Non zero response when checking solr cluster status: " + status);
```
rseitz, line 64: update error message to talk about pinging rather than checking cluster status

rseitz (changes requested):
realized we need test coverage for both scenarios...
ideally the test would create a mock solr client of each type
create a SolrIndexer using the constructor that lets us pass in a solr client
then call validateConnection on the solr client
finally check that right method was called on the mock client

kyle-kmw:
I noticed some existing tests are failing on this branch.

2023-09-21T20:01:29.9832230Z [ERROR] Errors: 
2023-09-21T20:01:29.9833750Z [ERROR]   RunnerTest.testIndexerConnectFailure:738 » IllegalArgument port out of range:1...
2023-09-21T20:01:29.9834500Z [ERROR]   SolrIndexerIntegrationTest.testDeleteByField:168 » RemoteSolr Error from serve...
2023-09-21T20:01:29.9835110Z [ERROR]   SolrIndexerIntegrationTest.testDeleteById:98 » RemoteSolr Error from server at...
2023-09-21T20:01:29.9835720Z [ERROR]   SolrIndexerIntegrationTest.testDeleteByIdsAndFields:246 » RemoteSolr Error fro...

> realized we need test coverage for both scenarios... ideally the test would create a mock solr client of each type create a SolrIndexer using the constructor that lets us pass in a solr client then call validateConnection on the solr client finally check that right method was called on the mock client

I agree with extending test coverage but I think the test will be more resilient if we write solr teste framework tests (embedded solr) instead of client mocks.

abijitran:
```suggestion
ttest
```

PR #49, opened 2023-09-25
LC-361 addrandomfield stage

abijitran:
Created a new stage that adds random data to a document

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
@@ -23,18 +23,30 @@ import java.util.stream.IntStream;
 
 
 /**
- * Fetches byte data of a given URL field and places data into a specified field
+ * Adds random data to a document field given parameters
  * <br>
  * Config Parameters -
  * <br>
- * inputDataPath (String, Optional) : file path to a text file that stores datapoints to be randomly placed into field,
+ * <p>
+ * <b>inputDataPath</b> (String, Optional) : file path to a text file that stores datapoints to be randomly placed into field,
  *  defaults to numeric data based on cardinality
- * fieldName (String, Optional) : Field name of field where data is placed, defaults to "data"
- * cardinality (int, Optional) : size of the subset of datapoints to be grabbed either from
+ * </p>
+ * <p>
+ * <b>fieldName</b> (String, Optional) : Field name of field where data is placed, defaults to "data"
+ *  </p>
+ * <p>
+ * <b>cardinality</b> (int, Optional) : size of the subset of datapoints to be grabbed either from
  *  given datapath or from random numbers
- * minNumOfTerms (Integer, Optional) : minimum number of terms to be in the field, defaults to null
- * maxNumOfTerms (Integer, Optional) : maximum number of terms to be in the field, defaults to null
- * fieldStructure (FieldType, Optional) : setting for structure of field, default or nested, allows for further settings to be easily added
+ * </p>
+ *  <p>
+ * <b>minNumOfTerms</b> (Integer, Optional) : minimum number of terms to be in the field, defaults to null
+ * </p>
+ * <p>
+ * <b>maxNumOfTerms</b> (Integer, Optional) : maximum number of terms to be in the field, defaults to null
+ * </p>
+ * <p>
+ * <b>fieldStructure</b> (FieldType, Optional) : setting for structure of field, default or nested, allows for further settings to be easily added
+ * </p>
  */
 public class AddRandomField extends Stage {
 
@@ -64,7 +76,7 @@ public class AddRandomField extends Stage {
     this.uniqueValues = null;
 
     if (this.minNumOfTerms == null ^ this.maxNumOfTerms == null) {
-      throw new StageException("Both minimum and maximum number of terms must be specified");
+      throw new StageException("Both minimum and maximum number of terms must be specified if either is specified");
     }
     if (this.minNumOfTerms == null && this.maxNumOfTerms == null) {
       this.minNumOfTerms = 1;
@@ -114,13 +126,12 @@ public class AddRandomField extends Stage {
   }
 
   private List<String> getFileData(String inputDataPath) throws StageException {
-    List<String> data = null;
     try {
-      data = Files.readAllLines(Path.of(inputDataPath));
+      List<String> data = Files.readAllLines(Path.of(inputDataPath));
+      return data;
     } catch (IOException e) {
       throw new StageException("Could not read provided file path", e);
     }
-    return data;
   }
 
   private List<String> getUniqueValues(boolean dataExists, List<String> inputData) {
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
@@ -1,25 +1,20 @@
 package com.kmwllc.lucille.stage;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.kmwllc.lucille.core.ConfigUtils;
 import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.Stage;
 import com.kmwllc.lucille.core.StageException;
 import com.typesafe.config.Config;
 import java.io.IOException;
-import java.io.InputStream;
-import java.net.MalformedURLException;
-import java.net.URISyntaxException;
-import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.Iterator;
-import org.apache.commons.io.IOUtils;
-import org.apache.commons.io.input.BoundedInputStream;
-import org.apache.http.HttpEntity;
-import org.apache.http.client.ClientProtocolException;
-import org.apache.http.client.methods.CloseableHttpResponse;
-import org.apache.http.client.methods.HttpGet;
-import org.apache.http.impl.client.CloseableHttpClient;
-import org.apache.http.impl.client.HttpClients;
-import org.apache.http.util.EntityUtils;
+import java.util.List;
+import java.util.Set;
 
 
 /**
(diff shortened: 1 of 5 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,148 @@
...
+  private final List<String> uniqueValues;
+
+  public AddRandomField(Config config) throws StageException {
+    super(config, new StageSpec().withOptionalProperties("input_data_path", "field_name", "cardinality", "min_num_of_terms",
+        "max_num_of_terms", "field_type"));
+    this.inputDataPath = ConfigUtils.getOrDefault(config, "input_data_path", null);
+    this.fieldName = ConfigUtils.getOrDefault(config, "field_name", "data");
+    this.minNumOfTerms = ConfigUtils.getOrDefault(config, "min_num_of_terms", null);
+    this.maxNumOfTerms = ConfigUtils.getOrDefault(config, "max_num_of_terms", null);
+    this.fieldtype = FieldType.valueOf(ConfigUtils.getOrDefault(config, "field_type", FieldType.DEFAULT.toString()));
```
rseitz, line 58: may eventually want to support different data types like strings, ints, longs, dates, etc.
should probably rename "fieldType" to something else like "isNested"
so that "fieldType" can be used later for data type

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,148 @@
...
+    this.dataArr = this.inputDataPath != null ? getFileData(this.inputDataPath) : null;
```
rseitz, line 59: loading dictionaries should be done in start()

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: Changed to "FieldStructure"

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: Moved to start()

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+
+/**
+ * Fetches byte data of a given URL field and places data into a specified field
```
rseitz, line 26: this line seems to have been copied from a different stage and doesn't apply here

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,155 @@
...
+    this.fieldName = ConfigUtils.getOrDefault(config, "field_name", "data");
+    this.minNumOfTerms = ConfigUtils.getOrDefault(config, "min_num_of_terms", null);
+    this.maxNumOfTerms = ConfigUtils.getOrDefault(config, "max_num_of_terms", null);
+    this.isNested = ConfigUtils.getOrDefault(config, "is_nested", false);
+    this.dataArr = null;
+    this.cardinality = null;
+    this.uniqueValues = null;
+
+    if (this.minNumOfTerms == null ^ this.maxNumOfTerms == null) {
+      throw new StageException("Both minimum and maximum number of terms must be specified");
```
rseitz, line 67: "Both minimum and maximum number of terms must be specified if either is specified."

if you have a better wording that expresses the same complete point, feel free to use yours

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,155 @@
...
+  private void populateFieldsDefault(String fieldName, Document doc, ArrayList<String> fieldDataArr) {
+    for (String value : fieldDataArr) {
+      doc.setOrAdd(fieldName, value);
+    }
+  }
+
+  private List<String> getFileData(String inputDataPath) throws StageException {
+    List<String> data = null;
+    try {
+      data = Files.readAllLines(Path.of(inputDataPath));
```
rseitz, line 119: you can put the return here, inside the try
that elimates the declaration above and the return below

lucille-core/src/test/java/com/kmwllc/lucille/stage/AddRandomFieldTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+import com.kmwllc.lucille.stage.StageFactory;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Test;
+
+public class AddRandomFieldTest {
+
+  private StageFactory factory = StageFactory.of(AddRandomField.class);
```
rseitz, line 22: can be final

abijitran: done

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: done

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: done

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+
+/**
+ * Fetches byte data of a given URL field and places data into a specified field
+ * <br>
+ * Config Parameters -
+ * <br>
```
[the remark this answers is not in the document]
abijitran: done

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: done

lucille-core/src/test/java/com/kmwllc/lucille/stage/AddRandomFieldTest.java
```diff
@@ -0,0 +1,110 @@
...
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Test;
+
+public class AddRandomFieldTest {
+
+  private final StageFactory factory = StageFactory.of(AddRandomField.class);
+
+  /**
```
rseitz, line 24: would be ideal to describe what this test is doing
however, if not planning to do that, this empty comment should be removed

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,166 @@
...
+ * </p>
+ * <p>
+ * <b>fieldName</b> (String, Optional) : Field name of field where data is placed, defaults to "data"
+ *  </p>
+ * <p>
+ * <b>cardinality</b> (int, Optional) : size of the subset of datapoints to be grabbed either from
+ *  given datapath or from random numbers
+ * </p>
+ *  <p>
+ * <b>minNumOfTerms</b> (Integer, Optional) : minimum number of terms to be in the field, defaults to null
```
rseitz, line 42: comment should explain the behavior that a caller of this method would observe
in that sense, even though there's an initial default of null, the actual default is 1, right?

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,166 @@
...
+      return data;
+    } catch (IOException e) {
+      throw new StageException("Could not read provided file path", e);
+    }
+  }
+
+  private List<String> getUniqueValues(boolean dataExists, List<String> inputData) {
+    List<String> uniqueValues = null;
+    if (dataExists) {
+      List<String> initialData = new ArrayList<>(inputData);
```
rseitz, line 140: initialData might contain duplicates
if two identical values are chosen randomly and added to uniqueValuesSet, the size of that set only increases by 1
so at the end of the for loop, uniqueValuesSet could be smaller than the desired cardinality
therefore it might be better to start by putting initialData into a Set, then what is currently named uniqueValuesSet can just be a List

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+    if (this.minNumOfTerms > this.maxNumOfTerms) {
+      throw new StageException("Minimum number of terms must be less than or equal to maximum");
+    }
+  }
+
+  @Override
+  public void start() throws StageException {
+    this.dataArr = this.inputDataPath != null ? getFileData(this.inputDataPath) : null;
+    this.cardinality = ConfigUtils.getOrDefault(config, "cardinality",
```
rseitz, line 93: wondering if rangeSize would be clearer than "cardinality"

lucille-core/src/test/java/com/kmwllc/lucille/stage/AddRandomFieldTest.java
abijitran: added comments for all test methods

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: changed to be 1 instead of null

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: Changed so that the initialData is passed into a set and then into an array so that all duplicates are removed

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
abijitran: Changed name to rangeSize (range_size for config)

PR #54, opened 2023-10-05
LC-359 DropDocument stage

h0rban:
### Summary:
Added a stage that allows to drop documents based on certain conditions. A use case that motivated such a stage was dropping a document if its ID is in a set of already indexed IDs.

### Completed Actions:
- added a stage that drops documents
- verified that the conditional framework works if the first stage sets a field, for example, `"found": true`, and then a conditional stage drops the documents based on the value in the found field
- Added configurations to the DictionaryLookup stage that allow checking if the values of the document's fields are in the dictionary

### Future Considerations:
- consider stop applying additional stages if one stage sets a document as dropped. However, the current implementation allows for one stage to drop a document and the next one to undrop it 
- a possibility of exploring the creation of a combined stage that would simplify chaining stages. This specific use case would be a combination of the following:
<ul>
  <ol>
    <li>  looking up if a document field is a set and setting a field "found" </li>
    <li> dropping the document based on the "found" field  </li>
    <li> removing "found" field for the documents that were not dropped</li>
  <ol>
</ul>

### Pending Tasks:
- [ ] research how to share memory across threads and the potential to introduce unexpected behavior


lucille-core/src/main/java/com/kmwllc/lucille/stage/StringHashSet.java
```diff
@@ -0,0 +1,71 @@
...
+import com.kmwllc.lucille.util.FileUtils;
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.util.HashSet;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StringHashSet implements LookupCollection<String> {
```
rseitz, line 13: it's confusing to have a class named StringHashSet that is not a kind of HashSet but something else like LookupCollection<String>

It looks to me like we could use a regular java.util.HashSet for our purposes here
we simply need some code to properly populate the HashSet
that code could be put in a utility method inside the stage
I'm not sure I'm seeing a justification for why we need the LookupCollection interface

lucille-core/src/main/java/com/kmwllc/lucille/stage/DropDocument.java
```diff
@@ -0,0 +1,22 @@
+package com.kmwllc.lucille.stage;
+
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.Stage;
+import com.kmwllc.lucille.core.StageException;
+import com.typesafe.config.Config;
+import java.util.Iterator;
+
+public class DropDocument extends Stage {
```
rseitz, line 9: please add javadoc that explains what this stage is for
mention that this stage is intended to be used with conditions

lucille-core/src/main/java/com/kmwllc/lucille/stage/LookupCollection.java
```diff
@@ -0,0 +1,10 @@
+package com.kmwllc.lucille.stage;
+
+public interface LookupCollection<T> {
```
rseitz, line 3: why do we need this interface?
why wouldn't it suffice to use a HashSet for our purposes?

lucille-core/src/test/java/com/kmwllc/lucille/stage/DropDocumentTest.java
```diff
@@ -0,0 +1,69 @@
...
+import com.kmwllc.lucille.core.StageException;
+import org.junit.Test;
+
+public class DropDocumentTest {
+
+  private final StageFactory factory = StageFactory.of(DropDocument.class);
+
+
+  @Test
+  public void testPublisher() {
```
rseitz, line 16: publishing happens at the beginning of the pipeline
we can't "drop" a document unless it has first been published
there are already tests in RunnerTest to make sure that dropped documents are handled properly by the lucille framework
DropDocumentsTest should focus only on making sure that the DropDocument stage does what it is supposed to do

lucille-core/src/main/java/com/kmwllc/lucille/stage/SetLookup.java
```diff
@@ -0,0 +1,67 @@
...
+    // optional
+    destination = ConfigUtils.getOrDefault(config, "destination", "setContains");
+    ignoreMissingSource = ConfigUtils.getOrDefault(config, "ignore_missing_source", false);
+    ignoreCase = ConfigUtils.getOrDefault(config, "ignore_case", false);
+    dropFound = ConfigUtils.getOrDefault(config, "drop_found", false);
+  }
+
+  @Override
+  public void start() throws StageException {
+    values = new StringHashSet(file, ignoreCase);
```
rseitz, line 42: this is where we can use a java.util.HashSet and call a utility method to populate it with the contents of the file

lucille-core/src/main/java/com/kmwllc/lucille/stage/SetLookup.java
```diff
@@ -0,0 +1,67 @@
+package com.kmwllc.lucille.stage;
+
+import com.kmwllc.lucille.core.ConfigUtils;
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.Stage;
+import com.kmwllc.lucille.core.StageException;
+import com.typesafe.config.Config;
+import java.util.Iterator;
+
+public class SetLookup extends Stage {
```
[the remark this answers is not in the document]
rseitz: suggest renaming the stage to LookupValue or DictionaryLookup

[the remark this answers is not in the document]
rseitz: per lucille conversation, we already have a DictionaryLookup stage -- let's aim to use that instead of creating a new SetLookup stage

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -94,44 +116,61 @@ private HashMap<String, String[]> buildHashMap(String dictPath) throws StageExce
...
   @Override
   public Iterator<Document> processDocument(Document doc) throws StageException {
+
+    // todo consider what should happen if using set and destination field is already populated
+    if (setOnly) {
+      for (String destField : destFields) {
+        doc.removeField(destField);
+      }
+    }
+
```
h0rban, line 154: Is it okay to remove fields that are already present in the document for this use case? I am considering adding a warning or documentation that a field is being removed.

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -156,4 +193,40 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+    try (BufferedReader reader = new BufferedReader(getFileReader(filename))) {
+      int lines = 0;
+      while (reader.readLine() != null) {
+        lines++;
+      }
+      return lines;
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
h0rban, line 231: these two methods are good ones to reuse
what is a good place to put them?

lucille-core/src/test/java/com/kmwllc/lucille/stage/DictionaryLookupTest.java
```diff
@@ -80,7 +83,139 @@ public void testGetLegalProperties() throws StageException {
...
         stage.getLegalProperties());
   }
+
+  @Test
+  public void testFileDoesNotExist() {
+    Throwable e = assertThrows(StageException.class, () -> factory.get("DictionaryLookupTest/missing_file.conf"));
+    // todo should consider propagating the underlying exception message to root (right now I see it in e.cause.target)
+    // assertEquals("File does not exist: classpath:DictionaryLookupTest/missing.txt", e.getMessage());
+    assertEquals("Failed to instantiate stage: com.kmwllc.lucille.stage.DictionaryLookup", e.getMessage());
+  }
```
h0rban, line 98: I would be very interested in investigating propagating error messages to the highest level for Lucille debugging. Can you help me draft a ticket for that?

lucille-core/src/test/java/com/kmwllc/lucille/stage/DictionaryLookupTest.java
```diff
@@ -80,7 +83,139 @@ public void testGetLegalProperties() throws StageException {
             "dest",
             "conditions",
             "class",
-            "dict_path"),
+            "dict_path",
+            "set_only",
+            "ignore_missing_source"),
         stage.getLegalProperties());
   }
```
h0rban, line 90: I am pretty sure that we do not currently validate the stage configurations by default without running Lucille with a verification flag. I would like to revisit this decision and potentially verify stage configurations every time the stages are initialized. 

I am bringing this up because it took me a few minutes to find a mistake I made in the configuration of specifying the `setOnly` parameter instead of `set_only`.

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -42,25 +51,34 @@ public class DictionaryLookup extends Stage {
   private final boolean usePayloads;
   private final UpdateMode updateMode;
   private final boolean ignoreCase;
+  private final boolean setOnly;
+  private final boolean ignoreMissingSource;
 
-  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+  // Dummy value to indicate that a key is present in the HashMap
+  private static final String[] PRESENT = new String[0];
+  private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
rseitz, line 59: The convention we use is to put "log" in lowercase. That's because it's not a "constant" even though it is static final. In other words, its state can change. You can read a discussion of this question here: https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -70,9 +88,13 @@ public void start() throws StageException {
    * @return the populated HashMap
    */
   private HashMap<String, String[]> buildHashMap(String dictPath) throws StageException {
-    HashMap<String, String[]> dict = new HashMap<>();
-    try (CSVReader reader = new CSVReader(FileUtils.getReader(dictPath))) {
-      // For each line of the dictionary file, add a keyword/payload pair to the Trie
+
+    // count lines and create a set with correct capacity
+    int lineCount = countLines(dictPath);
```
rseitz, line 93: Please add a comment about why & when it's important to set the initial capacity. Does it have a significant performance impact when creating a very large hashset? How large was the hashset that motivated this?

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
h0rban: makes sense, thank you

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
rseitz: Let's try to keep the behavior of this stage as consistent as possible whether or not setOnly is true or false. When setOnly==false we don't delete the destFields. We use doc.update along with the provided updateMode, which allows the user to control whether they want replace or append behavior -- see: `doc.update(destField, updateMode, outputValues.toArray(new String[0]))`.  Let's use the same approach when setOnly==true -- don't delete destFields at the beginning. Do the update via doc.update()

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
rseitz: For counting lines in a file, something like this is more concise:

`    try (Stream<String> fileStream = Files.lines(Paths.get(filename))) {
        int noOfLines = (int) fileStream.count();}`

You can put it in `com.kmwllc.lucille.util.FileUtils`

lucille-core/src/test/java/com/kmwllc/lucille/stage/DictionaryLookupTest.java
rseitz: The `Stage` constructor line 88 calls validateConfigWithConditions() so we should be doing validation when any stage is constructed.  Jake, I think you were the one who implemented that :) Can you give steps for reproducing the problem you encountered? If validation is not happening somewhere it's supposed to, yes, we should definitely make a ticket.

lucille-core/src/test/java/com/kmwllc/lucille/stage/DictionaryLookupTest.java
rseitz: How are you looking for the exception stacktrace to be structured?

If you do `e.printStackTrace()` inside this test you currently get this:

`com.kmwllc.lucille.core.StageException: Failed to instantiate stage: com.kmwllc.lucille.stage.DictionaryLookup
	at com.kmwllc.lucille.stage.StageFactory.get(StageFactory.java:68)
	at com.kmwllc.lucille.stage.StageFactory.get(StageFactory.java:45)
	at com.kmwllc.lucille.stage.DictionaryLookupTest.lambda$testFileDoesNotExist$0(DictionaryLookupTest.java:94)
	at org.junit.Assert.assertThrows(Assert.java:1001)
	at org.junit.Assert.assertThrows(Assert.java:981)
	at com.kmwllc.lucille.stage.DictionaryLookupTest.testFileDoesNotExist(DictionaryLookupTest.java:94)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:69)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:64)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:500)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:481)
	at com.kmwllc.lucille.stage.StageFactory.get(StageFactory.java:63)
	... 30 more
Caused by: com.kmwllc.lucille.core.StageException: File does not exist: classpath:DictionaryLookupTest/missing.txt
	at com.kmwllc.lucille.stage.DictionaryLookup.getFileReader(DictionaryLookup.java:210)
	at com.kmwllc.lucille.stage.DictionaryLookup.countLines(DictionaryLookup.java:223)
	at com.kmwllc.lucille.stage.DictionaryLookup.buildHashMap(DictionaryLookup.java:94)
	at com.kmwllc.lucille.stage.DictionaryLookup.<init>(DictionaryLookup.java:72)
	... 36 more
`

rseitz: formatting of comment got lost -- but the idea is that there's a top-level StageException that gets thrown when StageFactory fails to construct the new stage instance. Looking deeper into the causes we see that the stage constructor encountered a missing file. That's how it usually works -- top level exceptions have causes that you find further down

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
rseitz: thanks for adding the comment -- but I'm curious -- did you observe a specific performance improvement when setting the initial capacity, or did you add the line-counting code just assuming it would be a good idea? Fine either way, I'm just looking for a bit more info.

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
h0rban: moved to FileUtils. your version is more concise. however, we want a way to reuse the getReader method so that tests that rely on the following path are interpreted correctly `classpath:path/to/file`

lucille-core/src/test/java/com/kmwllc/lucille/stage/DictionaryLookupTest.java
h0rban: i see the desired message like this `assertEquals("File does not exist: classpath:DictionaryLookupTest/missing.txt", e.getCause().getCause().getMessage())` 
It would be great if we say the cause message in the top-level exception.

lucille-core/src/test/java/com/kmwllc/lucille/stage/DictionaryLookupTest.java
h0rban: you are right
I might not have noticed that the message is in the cause

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
h0rban: Code:
```java
import java.io.BufferedWriter;
import java.io.File;
import java.io.FileWriter;
import java.io.IOException;
import java.util.*;

public class Main {
  public static void main(String[] args) {

    List<Long> timesNoCapacity = new ArrayList<>();
    List<Long> timesCapacity = new ArrayList<>();

    for (int i = 1; i < 9; i ++) {

      int size = (int) Math.pow(10, i);

      String filename = "test" + size + ".txt";
      writeRandomInts(size, filename);

      // no capacity
      long start1 = System.currentTimeMillis();
      HashSet<String> set1 = new HashSet<>();
      addValuesFromFile(filename, set1);
      timesNoCapacity.add(System.currentTimeMillis() - start1);

      // clear
      set1.clear();

      // capacity
      long start2 = System.currentTimeMillis();
      HashSet<String> set2 = new HashSet<>((int) (size / .75f) + 1);
      addValuesFromFile(filename, set2);
      timesCapacity.add(System.currentTimeMillis() - start2);

      // clear
      set2.clear();
    }


    System.out.println("size,timesNoCapacity,timesCapacity");
    for (int i = 0; i < timesCapacity.size(); i ++) {
      System.out.println("size=" + ((int) Math.pow(10, i + 1)) + " -> " +  timesNoCapacity.get(i) + ", " + timesCapacity.get(i));
    }

    System.out.println("done");
  }


  // writes n random integers to file
  public static void writeRandomInts(int n, String filename) {
    try {
      FileWriter fw = new FileWriter(filename);
      BufferedWriter bw = new BufferedWriter(fw);
      for (int i = 0; i < n; i ++) {
        bw.write(UUID.randomUUID().toString());
        bw.newLine();
      }
      bw.close();
    } catch (IOException e) {
      throw new RuntimeException(e);
    }
  }

  // add values from file to a given set
  public static void addValuesFromFile(String filename, Set<String> set) {
    try {
      Scanner scanner = new Scanner(new File(filename));
      while (scanner.hasNextLine()) {
        set.add(scanner.nextLine());
      }
    } catch (IOException e) {
      throw new RuntimeException(e);
    }
  }
}
```


Output: 
```
size,timesNoCapacity,timesCapacity
size=10 -> 11, 1
size=100 -> 1, 1
size=1000 -> 6, 5
size=10000 -> 42, 19
size=100000 -> 77, 41
size=1000000 -> 500, 419
size=10000000 -> 5266, 4311
size=100000000 -> 109685, 83468
```

rseitz: This is an awesome comparison -- thank you! -- but, what does "timesCapacity" look like if you include the time to read all the file lines in an initial pass to find the number of lines, rather than knowing the number of lines in advance?

h0rban: updated the code in the comment above.
results with reading lines:
```
size,timesNoCapacity,timesCapacity
size=10 -> 15, 1
size=100 -> 1, 2
size=1000 -> 4, 5
size=10000 -> 13, 8
size=100000 -> 62, 52
size=1000000 -> 541, 487
size=10000000 -> 5268, 4823
size=100000000 -> 106658, 94633
```
Looks like a consistent 10% processing time decrease even with reading the number of lines first

lucille-core/src/test/java/com/kmwllc/lucille/stage/DictionaryLookupTest.java
rseitz: Thx for the comment on this. Let's continue this discussion elsewhere if we need to as it won't affect this particular PR.

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
rseitz: That's a cool finding -- thanks for testing this. Could you update the code comment to mention this conclusion -- say that we've observed a 10% time reduction for various sizes above 1000 when setting the initial capacity based on the number of file lines.

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -94,64 +115,83 @@ private HashMap<String, String[]> buildHashMap(String dictPath) throws StageExce
...
 
-      List<String> outputValues = new ArrayList<>();
-      for (String value : doc.getStringList(sourceField)) {
-        if (ignoreCase) {
-          value = value.toLowerCase();
+        // default value is true if this is the first field or if there are multiple fields
+        // in case where we have one destination field and multiple source fields we want to retrieve the current value
+        boolean defaultValue = true;
+        if (i != 0 && destFields.size() == 1) {
+          defaultValue = doc.getBoolean(destField);
```
rseitz, line 158: If we see a false here, it looks like the value is always going to stay false, so we could break out of the loop, right? Except the updateMode could be APPEND or SKIP in which case doc.getBoolean() is not going to give us the result of the previous set lookup.
I think it we are using multiple source fields with one destination, we have to prohibit UpdateMode.APPEND from being specified, otherwise the logic doesn't work -- what do you think?

rseitz (approved):

PR #55, opened 2023-10-10
LC-372 Fix bug with Intermitten Failures for AddRandomFieldTest

abijitran:
Saw errors due to the randomness of this stage, fixed by changing the range of the values being generated from closed to open. Confirmed no errors by using IntelliJ repeat feature to run all tests for this Stage 10,000+ times

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
@@ -29,7 +29,7 @@ import java.util.stream.IntStream;
  * <br>
  * <p>
  * <b>inputDataPath</b> (String, Optional) : file path to a text file that stores datapoints to be randomly placed into field,
- *  defaults to numeric data based on range size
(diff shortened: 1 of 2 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -146,7 +146,7 @@ private List<String> getUniqueValues(boolean dataExists, List<String> inputData)
       }
     } else {
       // create sequential list of numbers ending at range size
-      List<Integer> seqList = IntStream.rangeClosed(1, rangeSize)
+      List<Integer> seqList = IntStream.range(1, rangeSize)
```
rseitz, line 149: we still want to make sure the list's size is equal to rangeSize
so should the lower value be 0?

abijitran: Yea, I thought this was for IDs, but this is just the data that goes inside the random field so makes sense for it to be 0

rseitz (approved):

PR #56, opened 2023-10-11
LC-338 application-example.conf documentation

abijitran:
Reorganized application-example.conf and went through all mentions of config properties for Lucille items other than stages and plug-ins. There is definitely more to be added from reviewers and some unknown properties that need to be questioned / resolved.

### DIFF
--- a/application-example.conf
+++ b/application-example.conf
@@ -115,12 +115,24 @@ worker {
   # number of worker threads to start for each pipeline when running lucille in local / single-JVM mode
   threads: 2
 
-  # maximum time to spend processing a message, after which the worker will assume a problem and shut down
-  # NOTE: not supported while com.kmwllc.lucille.core.Worker and PipelineWorker are being reconciled
+  # maximum time to spend processing a message, after which the worker will assume a problem and shut down, assuming that
+  # worker.exitOnTimeout=true
   maxProcessingSecs: 600 # 10 minutes
 
   # name of pipeline
   pipeline: "pipeline_name"
+
+  # tells the worker to System.exit(1) assuming that the worker has not polled before the maximum time given
+  exitOnTimeout: "true"
+
+  # maximum time for the worker to poll back / process
+  maxProcessingSecs: 60000
+
+  # if set to true, will enable heartbeat checks between worker threads to check that they are alive
+  enableHeartbeat: true
+
+  # time in ms, between successive heartbeat checks
+  period: 6000
 }
 
 ################
--- a/application-example.conf
+++ b/application-example.conf
@@ -8,10 +8,17 @@
 # Each connector reads data from a source and emits that data as a sequence of Documents to be processed by a specific Lucille pipeline
 connectors: [
   {
-    name: "connector1", # name to assign to this connector, for use in logging / console output
-    class: "com.kmwllc.lucille.connector.CSVConnector", # connector implementation class
-    pipeline: "pipeline1", # name of pipeline that will process this connector's output; see Pipeline config block below
-    path: "/Volumes/Work/lucille/src/test/resources/test.csv" # property specific to CSVConnector indicating path to source file
+    # name to assign to this connector, for use in logging / console output
+    name: "connector1",
+
+    # connector implementation class
+    class: "com.kmwllc.lucille.connector.CSVConnector",
+
+    # name of pipeline that will process this connector's output; see Pipeline config block below
+    pipeline: "pipeline1",
+
+    # property specific to CSVConnector indicating path to source file
+    path: "/Volumes/Work/lucille/src/test/resources/test.csv"
   },
   {
     name: "connector2",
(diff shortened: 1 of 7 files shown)


application-example.conf
```diff
@@ -1,133 +1,192 @@
...
+
+    maxRequestSize: 250000000
+
+    securityProtocol: "SSL"
+
+    consumerPropertyFile: ".../consumer-conf/consumer.properties"
+
+    producerPropertyFile: ".../consumer-conf/producer.properties"
+
+    documentDeserializer: # TODO
```
abijitran, line 151: Am unsure what these items are doing, I see that KafkaUtils.java is placing them in a properties map that is then fed into the AdminUser creation, but not sure what a good example for these properties might be.

abijitran (commented):
Self-reviewing to pose a question

application-example.conf
```diff
@@ -1,133 +1,192 @@
...
   }
+## ----------------------------------------------------------------------------------------------------------------------------------
 
-  elastic {
-    host: "localhost",
-    port: "9200"
-    sendEnabled: ${?ELASTIC_SEND_ENABLED}
+
+## Misc: -----------------------------------------------------------------------------------------------------------------
+  # TODO: Do we still need this dev block and relevant features?
```
meghanboyd, line 165: Voted in Lucille planning to remove

application-example.conf
abijitran: Follow up: Might need to add tests to cover a lot of these properties, as they currently don't exist

lucille-core/src/main/java/com/kmwllc/lucille/core/ConfigUtils.java
```diff
@@ -25,9 +25,9 @@ public static Config loadConfig() throws ConfigException {
 
     log.info("pipeline.env system property = " + envName);
     // Override default settings in application.conf with settings for the named environment.
-    // For example, if application.conf contains x=1 and dev.x=2, and if the the
+    // For example, if application.conf contains x=1 and dev.x=2, and if the
     // pipeline.env system property is set to "dev", then accessing "x" will return 2
-    // even though the "dev" prefix was not included
+    // even though the "dev" prefix was not provided when accessing "x"
```
rseitz, line 30: this entire method is obsolete now so we should remove it

application-example.conf
```diff
@@ -89,45 +137,35 @@ kafka {
...
 
-  # maximum number of times across all workers that an attempt should be made to process any given document;
-  # when this property is not provided, retries will not be tracked and no limit will be imposed;
-  # this property requires that zookeeper is available at the specified zookeeper.connectString
-  #maxRetries: 2
+  producerPropertyFile: ".../consumer-conf/producer.properties"
 
-  # number of worker threads to start for each pipeline when running lucille in local / single-JVM mode
-  threads: 2
+  documentDeserializer: # TODO
```
rseitz, line 146: can use "com.kmwllc.lucille.message.KafkaDocumentDeserializer"
try to mention that this is the default and doesn't need to be set explicitly unless someone wants to use a custom implementation of Deserializer<Document>

Deserializer<Document> is an implementation of org.apache.kafka.common.serialization.Deserializer that can convert a message on a Kafka topic into a Lucille Document

application-example.conf
```diff
@@ -89,45 +137,35 @@ kafka {
...
 
-  # maximum time to spend processing a message, after which the worker will assume a problem and shut down
-  # NOTE: not supported while com.kmwllc.lucille.core.Worker and PipelineWorker are being reconciled
-  #maxProcessingSecs: 600 # 10 minutes
-}
+  documentSerializer: # TODO
```
rseitz, line 148: can use "com.kmwllc.lucille.message.KafkaDocumentSerializer"
try to mention that this is the default and doesn't need to be set explicitly unless someone wants to use a custom implementation of Serializer<Document>

Serializer<Document> is an implementation of org.apache.kafka.common.serialization.Serializer that can convert a Lucille document into a byte[] that can be written as a message on a Kafka topic

application-example.conf
```diff
@@ -89,45 +137,35 @@ kafka {
...
 
-publisher {
+  events: # TODO
```
rseitz, line 150: true is the default
can set to false to indicate that "events" (indicating whether a document succeeded or failed indexing) should not be sent as Kafka message
a setting of false would be used in a context where Lucille is running without a Runner/Publisher that would receive events and determine when the run is complete; in this context, there might be a WorkerIndexer running that reads documents directly from a given Kafka topic and sends them to the search engine without tracking successes or failures

application-example.conf
```diff
@@ -89,45 +137,35 @@ kafka {
...
-  # this setting controls
-  #     1) maximum size of the queue of published documents waiting to be processed
-  #     2) maximum size of the queue of completed documents waiting to be indexed
-  # e.g. if it is set to 10 then each each queue can contain up to 10 documents for a total of 20
-  #
-  # attempts to publish beyond this capacity will cause the publisher to block
-  # this setting applies only when running Lucille in local / single-JVM mode using a LocalMessageManager
-  # this setting defaults to 100
-  queueCapacity: 100
+  sourceTopic: # TODO
```
rseitz, line 152: this is the name of the kafka topic that contains Documents generated by the Publisher that are waiting to be picked up by a Worker and sent through the pipeline. Typically, lucille generates this topic "behind the scenes" and names it as <pipeline_name>_source
However, there are cases where Lucille runs without a Runner/Publisher and doesn't create the sourceTopic; rather an existing Kafka topic should be used as sourceTopic and a lucille Worker or WorkerIndexer should read Documents from that existing topic. In that scenario, a user would want to specify kafka.sourceTopic in the config and make sure it's set to whatever topic contains the documents that they want to process

application-example.conf
abijitran: Resolved

application-example.conf
abijitran: Thanks! Addressed in new commit

application-example.conf
abijitran: Thanks! Addressed in new commit

application-example.conf
```diff
@@ -1,80 +1,130 @@
...
+  # maximum number of times across all workers that an attempt should be made to process any given document;
+  # when this property is not provided, retries will not be tracked and no limit will be imposed;
+  # this property requires that zookeeper is available at the specified zookeeper.connectString
+  maxRetries: 2
+
+  # number of worker threads to start for each pipeline when running lucille in local / single-JVM mode
+  threads: 2
+
+  # maximum time to spend processing a message, after which the worker will assume a problem and shut down
+  # NOTE: not supported while com.kmwllc.lucille.core.Worker and PipelineWorker are being reconciled
```
rseitz, line 119: this "NOTE:" is obsolete & should be removed -- line 119 only

but on line 118 where it says "assume a problem and shut down," need to clarify that shut down only happens if worker.exitOnTimeout=true

as discussed, also want to mention worker.heartbeat and worker.period

abijitran: Addressed in new commit. However, if you could take another look at the comments and explanation I provided, I just want to make sure that it is accurate

rseitz (approved):

meghanboyd (approved):

PR #64, opened 2023-11-07
LC-385 Improved Naming

rseitz:
All interfaces and classes that used the phrase "MessageManager" now use "Messenger."
The goal of this change is to 1) increase brevity, 2) increase clarity re: what the class actually does.
PersistingLocalMessageManager which is used widely throughout the tests is now shortened to TestMessenger.
The use of Test in the name indicates that this class is intended only for test use; however, it has not been moved into the test tree in this PR because Runner depends on it.


abijitran (approved):
Looks great!

kyle-kmw (approved):

PR #66, opened 2023-11-08
adding a check to verify that data exists before trying to fetch/parse it.

kwatters:


lucille-plugins/lucille-parquet/src/main/java/com/kmwllc/lucille/parquet/connector/ParquetConnector.java
```diff
@@ -151,6 +151,10 @@ public void execute(Publisher publisher) throws ConnectorException {
 
   private static void setDocField(Document doc, Type field, SimpleGroup simpleGroup, int j) {
 
+    // check if we have a field value before setting it.
+    if (simpleGroup.getFieldRepetitionCount(j) == 0)
```
rseitz, line 155: We're following the convention of always using braces with an if statement -- based on the Google Java Style Guide here: https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

rseitz (approved):
Approving but have requested a minor change to match coding conventions

PR #67, opened 2023-11-09
LC-387 add framework for public documentation site

rseitz:
- docs live as a tree of markdown files under docs/site
- Hugo static site generation tool is used to convert markdown into html
- "Docsy" theme for Hugo is used to determine look/feel and various features of site
- github workflow is set up to run Hugo and deploy html to Github Pages on any change to main
- current docs are only placeholders/examples, this PR is just to set up infrastructure


akul-sethi (approved):
looks good to me

PR #73, opened 2023-11-16
Address error handling when SolrIndexer sends batch of documents to multiple collections

kyle-kmw:
The SolrIndexer when configured to use a collection field will send documents to multiple collections as determined by the collection field on each document. If one of the collections is down, or doesn't exist, or some other error occurs we do not want this to prevent the other documents in the batch to the collections that they are routed to. This PR defers the throwing of any exceptions from calling sendAddUpdateBatch or sendDeletionBatch by collecting them in a MultiException and throwing the error(s) collectively at the end of batch processing.

This PR does not address false reporting of failures. When an error occurs while processing a batch it is still the case that every document in the batch will be marked as a failure.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
@@ -22,6 +22,7 @@ import org.apache.solr.client.solrj.response.SolrPingResponse;
 import org.apache.solr.common.SolrInputDocument;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.SimpleOrderedMap;
+import org.eclipse.jetty.util.MultiException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import sun.misc.Signal;
@@ -110,6 +111,10 @@ public class SolrIndexer extends Indexer {
 
   @Override
   protected void sendToIndex(List<Document> documents) throws Exception {
+    //when sending updates/deletes to separate collections we do not want an error sending to a particular collection to
+    //prevent us from sending to other collections in the batch. collect any errors that occur in this multiexception and
+    //then use it to throw the error(s) at the end.
+    MultiException errorsSending = new MultiException();
 
     if (solrClient == null) {
       log.debug("sendToSolr bypassed for documents: " + documents);
@@ -146,7 +151,12 @@ public class SolrIndexer extends Indexer {
             && doc.has(deleteByFieldField)
             && deleteByFieldValue != null
             && doc.has(deleteByFieldValue))) {
-          sendAddUpdateBatch(collection, solrDocRequests.getAddUpdateDocs());
+          try {
+            sendAddUpdateBatch(collection, solrDocRequests.getAddUpdateDocs());
+          } catch (Exception e) {
+            errorsSending.add(e);
+          }
+
           solrDocRequests.resetAddUpdates();
         }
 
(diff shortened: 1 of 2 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -22,6 +22,7 @@
 import org.apache.solr.common.SolrInputDocument;
 import org.apache.solr.common.util.NamedList;
 import org.apache.solr.common.util.SimpleOrderedMap;
+import org.eclipse.jetty.util.MultiException;
```
rseitz, line 25: org.eclipse.jetty:jetty-util is a transitive dependency of solrj...
Should this be referenced explicitly in the lucille-core pom, or just leave it as a transitive dependency?
Seems like we might want to move the SolrIndexer to its own module at some point, in which case we'd have to deal transitive dependencies that we're relying on without declaring... Thoughts on a general policy for this kind of situation?

lucille-core/src/test/java/com/kmwllc/lucille/indexer/SolrIndexerIntegrationTest.java
```diff
@@ -339,6 +342,86 @@ public void testValidateConnection() throws SolrServerException, IOException {
...
+      Document doc1 = Document.create("id_1");
+      doc1.setField("foo", "bar");
+      doc1.setField("collection", COL);
+
+      // Send one document to a collection that doesn't exist.
+      Document doc2 = Document.create("id_2");
+      doc2.setField("foo", "baz");
+      doc2.setField("collection", "this-collection-does-not-exist");
+
+      // Send a second  document to the test collection.
```
rseitz, line 385: // send a second document to a different collection that doesn't exist

lucille-core/src/test/java/com/kmwllc/lucille/indexer/SolrIndexerIntegrationTest.java
```diff
@@ -339,6 +342,86 @@ public void testValidateConnection() throws SolrServerException, IOException {
...
+      Document doc3 = Document.create("id_3");
+      doc3.setField("foo", "bar");
+      doc3.setField("collection", "this-collection-also-does-not-exist");
+
+      Throwable exception = assertThrows(SolrException.class, () -> indexer.sendToIndex(List.of(doc1, doc2)));
```
rseitz, line 390: how come we are sending doc1 and doc2 but NOT doc3 here?

lucille-core/src/test/java/com/kmwllc/lucille/indexer/SolrIndexerIntegrationTest.java
```diff
@@ -339,6 +342,86 @@ public void testValidateConnection() throws SolrServerException, IOException {
...
+      cluster.getSolrClient().commit(COL);
+
+      assertEquals(
+          "The index should be empty.",
+          0,
+          cluster.getSolrClient().query(COL, qr).getResults().size());
+
+      //run the same test again with docs targeting both collections that do not exist with doc1 last.
+
+      exception = assertThrows(MultiException.class, () -> indexer.sendToIndex(List.of(doc3, doc2, doc1)));
```
rseitz, line 409: would there be any added testing value in putting doc1 in the middle?

rseitz:
do we have a ticket for this?

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -110,6 +111,10 @@ public void closeConnection() {
 
   @Override
   protected void sendToIndex(List<Document> documents) throws Exception {
+    //when sending updates/deletes to separate collections we do not want an error sending to a particular collection to
```
rseitz, line 114: super-minor, but whitespace after the // would be in keeping with the convention

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -164,22 +174,35 @@ protected void sendToIndex(List<Document> documents) throws Exception {
...
           solrDocRequests.resetDeletes();
         }
         solrDocRequests.addDocForAddUpdate(solrDoc);
       }
     }
     for (String collection : solrDocRequestsByCollection.keySet()) {
-      sendAddUpdateBatch(
-          collection, solrDocRequestsByCollection.get(collection).getAddUpdateDocs());
-      sendDeletionBatch(collection, solrDocRequestsByCollection.get(collection));
+      try {
```
rseitz, line 193: so, this PR has a bigger consequence than just making sure that failures against one collection don't prevent us from sending to different collections...

it also means that if we have an certain doc ID and there's an update followed by a delete, we might get a failure from the update but still go on to try the delete. Likewise, if there's a delete followed by an update, we might get a failure from the delete, but still go on to try the update. Thinking about this, maybe this new behavior is fine. The important thing is that we never swap an update and a delete, and that's not happening here. But, I wanted to leave this comment just to bring attention to the change -- is this something you considered and also concluded was OK? A code comment recording our conclusion might be helpful for the future.

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
kyle-kmw: I think you are right, we should declare dependencies that we directly use explicitly and not rely on transitive dependencies from a 3rd party. That said I actually think we can just have our own exception for this (which might be refactored into a batch response / batchlet given other work).

abijitran (approved):
Nice!

meghanboyd:
> do we have a ticket for this PR?

I created LC-395 for this issue because there was not a ticket. @kyle-kmw could you please add LC-395 to the PR title? Thanks.

PR #74, opened 2023-11-20
update Indexer.sendToIndex to return a list of failed docs

rseitz:
Draft PR to sketch out a way make error reporting in Indexer more granular. Currently, if any document in a batch fails, all documents are treated as having failed. We want to allow some docs to fail while others succeed. 

- the abstract Indexer.sendToIndex(List<Document) method now returns a list of documents that failed, which may be null if there were no failures. Eventually, a dedicated class like BatchResult would be a better return value -- the current code is a sketch
- an indexer implementation may still throw an exception from sendToIndex, in which case all docs in the batch will treated as failed (this is the old/existing behavior)
- an indexer implementation may now choose to NOT throw an exception from sendToIndex but instead return a list of the specific documents in the batch that failed (assuming all other documents succeeded). In this case, the indexer base class will only generate error events for the specific documents that are known to have failed, while all other documents in the batch will be treated as having succeeded
- TODO: change the return value of sendToIndex to a BatchResult that contains a list of succeeded documents and a list of failed documents along with specific exceptions encountered. Note that there may be some documents in the batch that don't fall into either list because an attempt was not made to send them (i.e. after an error was encountered with a previous document so indexing was aborted).

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
@@ -123,7 +123,7 @@ public abstract class Indexer implements Runnable {
    * call to the batch API provided by the search engine client, if available, as opposed to sending
    * each document individually.
    */
-  protected abstract void sendToIndex(List<Document> documents) throws Exception;
+  protected abstract List<Document> sendToIndex(List<Document> documents) throws Exception;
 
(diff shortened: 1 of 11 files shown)


rseitz:
closing this draft PR was this work was carried forward in https://github.com/kmwtechnology/lucille/pull/292

PR #79, opened 2024-01-23
LC-332 runner with -validate flag printing config invalid for valid configs

akul-sethi:
there was a formatting issue so I reverted the original commit


rseitz:
please put "LC-332" at the beginning of the PR title
you don't need to use words like "Fixed:" because it's implied that the PR fixes the issue described int he ticket

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -164,23 +164,26 @@ public static Map<String, TestMessenger> runInTestMode(String configName) throws
 
   private static Map<String, List<Exception>> logValidation(Config config) throws Exception {
     Map<String, List<Exception>> exceptions = validatePipelines(config);
-    if (exceptions.isEmpty()) {
```
rseitz, line 167: exceptions is going to be a map from connector name to a list of exceptions for that connector
we want to print "Configuration is valid" if all connectors have an empty list of exceptions
so, instead of doing exceptions.isEmpty() we need a condition that checks isEmpty() for every value in the map

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -164,23 +164,26 @@ public static Map<String, TestMessenger> runInTestMode(String configName) throws
...
-        for (Exception e : entry.getValue()) {
-          message.append("\t\tException ").append(i++).append(": ")
-              .append(e.getMessage()).append("\n");
+
+    for (Map.Entry<String, List<Exception>> entry : exceptions.entrySet()) {
+      if (!entry.getValue().isEmpty()) {
+        StringBuilder message =
+            new StringBuilder("Configuration is invalid. " + "Printing the list of exceptions for each pipeline\n");
+
+        for (Map.Entry<String, List<Exception>> entryInner : exceptions.entrySet()) {
```
rseitz, line 173: let's try to implement what we want without nested for loops over the same set

akul-sethi: done

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -165,25 +165,24 @@ public static Map<String, TestMessenger> runInTestMode(String configName) throws
...
+          message.append("\t\tException ").append(i++).append(": ").append(e.getMessage()).append("\n");
         }
-
-        log.error(message.delete(message.length() - 1, message.length()).toString());
-        return exceptions; 
       }
+
+      log.error(message.delete(message.length() - 1, message.length()).toString());
+
+    } else {
```
rseitz, line 183: when you have an if/else where one clause is really simple, it can help readability to put that first (as part of the the if)

In this case, you could take the ! out of `if(!exceptions.entrySet(...` and put the "Configuration is valid" message there

also note that per the java conventions we follow, there should be a space between an "if" and the open paren

akul-sethi:
test coverage for this will be coming in LC-401. Discussed with Rudi.

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -164,21 +164,22 @@ public static Map<String, TestMessenger> runInTestMode(String configName) throws
...
     Map<String, List<Exception>> exceptions = validatePipelines(config);
-    if (exceptions.isEmpty()) {
+
+    if (exceptions.entrySet().stream().allMatch(e -> e.getValue().isEmpty())) {
       log.info("Configuration is valid");
     } else {
-      StringBuilder message = new StringBuilder("Configuration is invalid. " +
-          "Printing the list of exceptions for each pipeline\n");
+      StringBuilder message =
+          new StringBuilder("Configuration is invalid. " + "Printing the list of exceptions for each pipeline\n");
```
rseitz, line 172: this is on one line now, so we no longer need the +

rseitz (approved):

PR #81, opened 2024-01-24
LC-278 ComputeFieldSize stage

akul-sethi:
one of the tests logs an "error accessing bye[] field" but that is intentional. The JsonDocument implementation of Document getBytes() method logs this to notify users that they are incorrectly interpreting a field as a byte[], but allows the program to keep running by returning null.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
@@ -26,10 +26,10 @@ public class ComputeFieldSize extends Stage {
   private final String destination;
 
   public ComputeFieldSize(Config config) {
-    super(config, new StageSpec().withRequiredProperties("source", "destination"));
+    super(config, new StageSpec().withRequiredProperties("source", "dest"));
 
     this.source = config.getString("source");
-    this.destination = config.getString("destination");
+    this.destination = config.getString("dest");
   }
 
   /**
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
@@ -5,6 +5,7 @@ import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.Stage;
 import com.kmwllc.lucille.core.StageException;
 import com.typesafe.config.Config;
+import com.amazonaws.services.s3.model.SelectObjectContentEvent.ContinuationEvent;
 import com.kmwllc.lucille.core.*;
 
 /**
@@ -41,10 +42,14 @@ public class ComputeFieldSize extends Stage {
   @Override
   public Iterator<Document> processDocument(Document doc) throws StageException {
     if (!doc.has(source)) {
-      throw new StageException(String.format("Document does not have source field: %s", source));
+      return null;
     }
 
-    doc.update(destination, UpdateMode.OVERWRITE, doc.getBytes(source).length);
+    try {
+      doc.update(destination, UpdateMode.OVERWRITE, doc.getBytes(source).length);
+    } catch (NullPointerException e) {
+      throw new StageException(String.format("Field: {} is not a byte array", source));
+    }
     return null;
   }
 }
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
@@ -1,0 +1,45 @@
+package com.kmwllc.lucille.stage;
+
+import java.util.Iterator;
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.Stage;
+import com.kmwllc.lucille.core.StageException;
+import com.typesafe.config.Config;
+import com.kmwllc.lucille.core.*;
+
+/**
+ * Computes sizes of given byte array field and adds a field with that size
+ *
+ * Config Paramters:
+ *   - sourceField (String) : The field containing the byte array 
+ *   - outputField (String) : The name of the field which will be added to store the size. Will overwrite value if field already 
+ *                            exists
+ */
+public class ComputeFieldSize extends Stage {
+
+  private final String sourceField;
+  private final String outputField;
+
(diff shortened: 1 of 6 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
```diff
@@ -0,0 +1,45 @@
...
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.Stage;
+import com.kmwllc.lucille.core.StageException;
+import com.typesafe.config.Config;
+import com.kmwllc.lucille.core.*;
+
+/**
+ * Computes sizes of given byte array field and adds a field with that size
+ *
+ * Config Paramters:
```
rseitz, line 13: Formatting like this in javadoc generally requires html-like markup to render properly.

lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class ComputeFieldSize extends Stage {
+
+  private final String sourceField;
+  private final String outputField;
+
+  public ComputeFieldSize(Config config) {
+    super(config, new StageSpec().withRequiredProperties("sourceField", "outputField"));
+
+    this.sourceField = config.getString("sourceField");
```
rseitz, line 26: Other stages that take a source field generally refer to it as "source", not "sourceField". Let's try to be consistent with any implicit conventions like this. Should also check on "outputField" -- possibly rename to "destination"

lucille-core/src/test/java/com/kmwllc/lucille/stage/ComputeFieldSizeTest.java
```diff
@@ -0,0 +1,79 @@
...
+    assertThrows(StageException.class, () -> {
+      factory.get("ComputeFieldSizeTest/no-source.conf");
+    });
+    assertThrows(StageException.class, () -> {
+      factory.get("ComputeFieldSizeTest/no-destination.conf");
+    });
+  }
+
+  @Test
+  public void testSourceFieldDoesNotExist() throws StageException {
```
rseitz, line 27: In general we want stages to be tolerant of a source field not existing and simply skip processing in that case. Lots of different kinds of documents could be flowing through a lucille pipeline and the goal is to "get as many documents through" as possible.  You can see this approach in stages like CopyFields and DeleteFields, there's a conditional like 

      if (!doc.has(sourceField)) {
        continue;
      }

Let's take the same approach here.

lucille-core/src/test/java/com/kmwllc/lucille/stage/ComputeFieldSizeTest.java
```diff
@@ -0,0 +1,79 @@
...
+    });
+  }
+
+  @Test
+  public void testSourceNotByteArray() throws StageException {
+    Stage stage = factory.get("ComputeFieldSizeTest/basic.conf");
+    Document doc1 = Document.create("doc1");
+    doc1.update("source", UpdateMode.OVERWRITE, "string");
+
+    assertThrows(NullPointerException.class, () -> {
```
rseitz, line 42: It's good that this test documents the current behavior -- but NPE's aren't great exceptions for stages to throw in general. To refine the behavior in this case, we have a choice between 1) skipping the doc, 2) adding something to the destination field to indicate the error (could be -1 to be consistent with a number going there, or could be a text error), 3) throwing a StageException. What do you think is best here? I'm leaning toward 3.

lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
```diff
@@ -0,0 +1,55 @@
...
+   */
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+    if (!doc.has(source)) {
+      return null;
+    }
+
+    try {
+      doc.update(destination, UpdateMode.OVERWRITE, doc.getBytes(source).length);
+    } catch (NullPointerException e) {
```
rseitz, line 50: It's generally better to check for null explicitly rather than generate an NPE and catch it. Here I would save doc.getBytes(source) in a variable, check if the variable is null and throw the stage exception if so. This makes it easier for the reader to see where specifically in the code you're worried about the null arising.

lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
```diff
@@ -0,0 +1,55 @@
...
+   * @throws NullPointerException If {@link ComputeFieldSize#source} is not a byte array in document
+   */
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+    if (!doc.has(source)) {
+      return null;
+    }
+
+    byte[] bytes = doc.getBytes(source);
+    if(bytes == null) {
```
rseitz, line 48: space between if and (

lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
```diff
@@ -0,0 +1,55 @@
...
+      throw new StageException(String.format("Field: {} is not a byte array", source));
+    }
+
+    doc.update(destination, UpdateMode.OVERWRITE, doc.getBytes(source).length);
```
rseitz, line 52: don't need to call getBytes() a second time -- we have the var named bytes
doc.getBytes is one of the getters on Document that could possibly be expensive

lucille-core/src/main/java/com/kmwllc/lucille/stage/ComputeFieldSize.java
```diff
@@ -0,0 +1,55 @@
...
+ * exists
+ * </p>
+ */
+public class ComputeFieldSize extends Stage {
+
+  private final String source;
+  private final String destination;
+
+  public ComputeFieldSize(Config config) {
+    super(config, new StageSpec().withRequiredProperties("source", "destination"));
```
rseitz, line 29: There's an implicit convention in the stage package where a destination field is called "dest" and not "destination"
Let's follow that convention there.
I found one exception -- the Timestamp stage calls it "dest_field" but we should change that.

rseitz (approved):

PR #82, opened 2024-01-25
LC-401 runner validates all pipelines once

akul-sethi:
Originally, runInValidationMode returned a map of connector names to a list of exceptions caused by its pipeline. Since the goal of this method is to validate all the pipelines once (even if they are used by multiple or zero connectors), I think it makes more sense to return a map of pipelines to exceptions and directly validate each pipeline in the config. This automatically fixes both issues and only required one test to be changed. 

When support for validating connectors and indexers is added they can each get their own section while logging so users can specifically see which connectors are failing and if it is due to its pipeline.


lucille-core/src/test/java/com/kmwllc/lucille/stage/ConfigValidationTest.java
```diff
@@ -57,33 +58,80 @@ public void testNonValidationModeException() throws Exception {
...
+    Map<String, List<Exception>> exceptions = Runner.runInValidationMode(addPath("no-used-pipelines.conf"));
+    assertEquals(1, exceptions.size());
+
+    List<Exception> exceptions1 = exceptions.get("pipeline1");
+    assertEquals(2, exceptions1.size());
 
-    // TODO note that for the following two exceptions, the fields are retrieved before
-    //  the config validation is called
+    testException(exceptions1.get(0), IllegalArgumentException.class,
+        "com.kmwllc.lucille.stage.NoopStage: " + "Stage config contains unknown property invalid_property");
```
rseitz, line 85: this is on one line so we don't need the plus. same issue occurs in one or two other places

lucille-core/src/test/java/com/kmwllc/lucille/stage/ConfigValidationTest.java
```diff
@@ -57,33 +58,80 @@ public void testNonValidationModeException() throws Exception {
...
 
-    testException(exceptions1.get(1), IllegalArgumentException.class,
-        "Stage config must contain property fields");
+    testException(exceptions1.get(1), IllegalArgumentException.class, "Stage config must contain property fields");
+  }
 
-    testException(exceptions2.get(0), IllegalArgumentException.class,
-        "Stage config must contain property dest");
+  @Test
+  public void testStringifyValidationExceptions() {
```
rseitz, line 91: Runner.validatePipelines uses a LinkedHashMap instead of a plain HashMap so key entry order is preserved. When we stringify the exception map, we should respect that order, and tests should confirm this. The goal is that the user running the validation command should see the output for each pipeline in the same order the pipelines occur in the config file. e.g. if the config has:

pipelines: [{name: "pipeline1", stages: [{}]}, {name: "pipeline2", stages: [{}]}]

we should see the validation output for pipeline1, then pipeline2

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -162,43 +162,54 @@ public static Map<String, TestMessenger> runInTestMode(String configName) throws
     return result.getHistory();
   }
 
+  // Returns a mapping from pipeline names to the list of exceptions produced when validating them.
   private static Map<String, List<Exception>> logValidation(Config config) throws Exception {
```
rseitz, line 166: Suggest that this method should be typed as `public static void logValidation(Map<String, List<Exception>> exceptionMap)` and be dedicated only to logging (and stringifying in order to do logging). Some changes would be needed in a few places where logValidation is called, to get the validation result and pass it to this method.

lucille-core/src/test/resources/ConfigValidationTest/duplicate-pipeline.conf
```diff
@@ -0,0 +1,42 @@
...
+  {
+    name: "connector2",
+    class: "com.kmwllc.lucille.connector.CSVConnector",
+    pipeline: "pipeline1",
+    idField: "id",
+    path: "classpath:RunnerTest/singleDoc.csv",
+  }
+]
+
+pipelines: [
```
rseitz, line 18: comment about the filename "duplicate-pipeline.conf"
this is really an example of a "reused" pipeline -- i.e. a pipeline that is used or shared across multiple connectors
"duplicate pipeline" suggests two pipelines definitions with the same name

rseitz (approved):
approving with suggestion to rename "duplicate pipeline" conf file and corresponding test as "reused pipeline" or "shared pipeline"

PR #83, opened 2024-01-29
LC-403 lucille stage which extracts raw text from selected html elements

akul-sethi:
Selects elements from an HTML document and extracts the raw text or attributes from them into destination fields. Supported selectors are shown here: [CSS selectors](https://jsoup.org/cookbook/extracting-data/selector-syntax). 

When testing, I chose not to mock Jsoup and rather texted it directly as it makes the usage of the stage clearer.


lucille-core/src/main/java/com/kmwllc/lucille/stage/JSoupExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+    if (this.filePathField != null && this.byteArrayField != null) {
+      throw new StageException("Stage cannot have both filePathField and byteArrayField specified");
+    }
+
+    for (Map.Entry<String, Object> entry : destinationFields.entrySet()) {
+      Object value = entry.getValue();
+      if (value instanceof Map) {
+        Map<String, String> map = (Map<String, String>) value;
+        if (map.keySet().size() != 2)
+          throw new StageException("map must contain two fields");
```
rseitz, line 65: our convention is to always use braces around the body of an if statement, even it it's only one line

lucille-core/src/main/java/com/kmwllc/lucille/stage/JSoupExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+        throw new StageException("destinationFields mapping can only have strings or a map as its values");
+      }
+    }
+  }
+
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+    String field = filePathField != null ? filePathField : byteArrayField;
+
+    if (doc.has(field)) {
```
rseitz, line 78: suggest a shorter if statement here

if (!doc.has(field) { 
  return null; 
}

then the main logic can come after

lucille-core/src/main/java/com/kmwllc/lucille/stage/JSoupExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+            doc.update(entry.getKey(), UpdateMode.OVERWRITE, jsoupDoc.select((String) value).eachText().toArray(new String[0]));
+          } else {
+            // if the user specified a `destination -> {selector: selector, attribute: attribute}` mapping
+            Map<String, String> config = (Map<String, String>) value;
+            doc.update(entry.getKey(), UpdateMode.OVERWRITE, jsoupDoc.select(config.get("selector")).eachAttr(config.get("attribute")).toArray(new String[0]));
+          }
+
+        }
+      } catch (IOException e) {
+        throw new StageException("File parse failed: " + e.getMessage());
```
rseitz, line 97: about the exception message -- we weren't necessarily parsing a file... for example, we could be parsing a byte[] that was set by a FetchUri stage

so... message should say something like "HTML parsing failed"

lucille-core/src/main/java/com/kmwllc/lucille/stage/JSoupExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+    String field = filePathField != null ? filePathField : byteArrayField;
+
+    if (doc.has(field)) {
+      try {
+        org.jsoup.nodes.Document jsoupDoc = filePathField != null ? Jsoup.parse(new File(doc.getString(filePathField)))
+            : Jsoup.parse(new String(doc.getBytes(byteArrayField), StandardCharsets.UTF_8));
```
rseitz, line 81: we should probably support a stringField along with byteArrayField... and when the user specifies byteArrayField there should be a way to override the default charset

lucille-core/src/main/java/com/kmwllc/lucille/stage/JSoupExtractor.java
```diff
@@ -0,0 +1,103 @@
...
+ * not modified
+ * </p>
+ * <b>destinationFields</b> (Map<String, String>) : defines a mapping from css selectors to the destination fields 
+ * in the processed document. If they already exist they are overwritten. Otherwise, they are created. If a selector returns 
+ * multiple elements the destination field receives a list of processed elements. If a selector returns no elements for a document then the 
+ * destination field is not created
+ * @see <a href="https://jsoup.org/cookbook/extracting-data/selector-syntax"> CSS selectors </a> for information on supported selectors
+ * </p>
+ */
+public class JSoupExtractor extends Stage {
```
rseitz, line 39: most stage names are phrased as verbs / actions 
ExtractHTML conforms to that approach, but you received a suggestion to include JSoup in the name because it is so well known
So... how about ApplyJSoup?

lucille-core/src/main/java/com/kmwllc/lucille/stage/JSoupExtractor.java
akul-sethi: The  IOException is only thrown when the file parsing fails. Jsoup.parse for a String does not throw any declared exceptions

rseitz: ok, I'd suggest to put the try/catch only around the smaller block of code where an exception can be thrown -- only the line where we're calling Jsoup.parse. Also add a brief comment saying the IOException is only thrown when parsing a file

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJSoup.java
```diff
@@ -0,0 +1,137 @@
...
+        if (!map.containsKey("attribute") || !map.containsKey("selector")) {
+          throw new StageException("map must contain a attribute and selector fields");
+        }
+      } else if (!(value instanceof String)) {
+        throw new StageException("destinationFields mapping can only have strings or a map as its values");
+      }
+    }
+  }
+
+  private String activeField() {
```
rseitz, line 89: We can determine and store the active field in the constructor of this stage -- I don't think we need to determine it dynamically on each call to processDocument

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJSoup.java
```diff
@@ -0,0 +1,182 @@
...
+
+  public ApplyJSoup(Config config) throws StageException {
+    super(config, new StageSpec().withOptionalProperties("filePathField", "byteArrayField", "stringField", "charset")
+        .withRequiredParents("destinationFields"));
+
+    this.destinationFields = config.getConfig("destinationFields").root().unwrapped();
+    this.charset = ConfigUtils.getOrDefault(config, "charset", null);
+
+    if (!((config.hasPath("filePathField") ^ config.hasPath("byteArrayField") ^ config.hasPath("stringField")
+        ^ (config.hasPath("filePathField") && config.hasPath("byteArrayField") && config.hasPath("stringField"))))) {
```
rseitz, line 82: what's the purpose of the l82?

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJSoup.java
```diff
@@ -0,0 +1,182 @@
...
+ * </p>
+ */
+public class ApplyJSoup extends Stage {
+
+  private final Map<String, Object> destinationFields;
+  private final String charset;
+  private final String activeField;
+  private final Mode mode;
+
+  private enum Mode {
```
rseitz, line 70: suggest renaming to InputType or SourceType

PR #84, opened 2024-01-31
LC-404 pinecone indexer sends to the default namespace if none are provided

akul-sethi:
Changed the stage so that documents are indexed to the default namespace if none is provided in the config. If no namespace property is provided a "defaultEmbeddingField" property must be provided so that the vectors to be indexed can be found.

 Additionally, I changed the way namespaces are indicated. Since there is a one-to-one mapping between namespaces and embedding fields, rather than having a list for each I have a map from namespaces to their respective fields. If an empty namespace map is provided the program throws an error rather than doing nothing or sending documents to the default namespace. The former is behavior that someone would rarely want and the latter is inconsistent with when namespaces are found in the map. It suggests that vectors will be indexed to the default namespace in addition to the namespaces that are provided which we also do not want.


lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -14,39 +22,38 @@
...
-        .withApiKey(config.getString("pinecone.apiKey"))
-        .withEnvironment(config.getString("pinecone.environment"))
-        .withProjectName(config.getString("pinecone.projectName"))
+    this.defaultEmbeddingField = ConfigUtils.getOrDefault(config, "pinecone.defaultEmbeddingField", null);
+    PineconeClientConfig configuration = new PineconeClientConfig().withApiKey(config.getString("pinecone.apiKey"))
+        .withEnvironment(config.getString("pinecone.environment")).withProjectName(config.getString("pinecone.projectName"))
         .withServerSideTimeoutSec(config.getInt("pinecone.timeout"));
     this.client = new PineconeClient(configuration);
+
+    if(namespaces == null && defaultEmbeddingField == null) {
```
rseitz, line 49: space after if

kyle-kmw (approved):

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -14,39 +22,38 @@
...
-        .withProjectName(config.getString("pinecone.projectName"))
+    this.defaultEmbeddingField = ConfigUtils.getOrDefault(config, "pinecone.defaultEmbeddingField", null);
+    PineconeClientConfig configuration = new PineconeClientConfig().withApiKey(config.getString("pinecone.apiKey"))
+        .withEnvironment(config.getString("pinecone.environment")).withProjectName(config.getString("pinecone.projectName"))
         .withServerSideTimeoutSec(config.getInt("pinecone.timeout"));
     this.client = new PineconeClient(configuration);
+
+    if (namespaces == null && defaultEmbeddingField == null) {
+      throw new IllegalArgumentException(
+          "either a mapping of namespaces to embedding fields or a defaultEmbeddingField must be provided");
```
rseitz, line 51: "at least one of a defaultEmbeddingField or a non-empty namespaces mapping is required"

rseitz (approved):

PR #88, opened 2024-02-12
LC-412 restore ability to start Indexer in fully distributed mode

rseitz:
This PR removes the duplicated main() methods from some of the Indexer subclasses and places it back on the Indexer base class. The aim is that a user wishing to run Lucille in fully-distributed mode should be able to run com.kmwllc.lucille.core.Indexer at the command line and have the appropriate kind of indexer be started based on the configuration.

Note: We don't yet have test coverage for the main() methods on Indexer, WorkerIndexer, or Worker. I tested this PR manually. The manual test involved starting kafka+zookeeper, starting a lucille Indexer, Worker, and then launching a run via Runner. I confirmed that the run succeeded. In my case, I was using a config file that specified a CSVIndexer. I checked that the csv file that the run should have generated was indeed created and properly populated after the run.


akul-sethi (approved):
looks good to me

PR #90, opened 2024-02-27
LC-399 make sure Worker and WorkerIndexer terminate when a startup exception is encountered

rseitz:
When starting Worker or WorkerIndexer at the command line, the java process would not terminate if there was an error in certain parts of the config file, including the pipeline section and/or the kafka section. This was happening because at least one thread had been started before attempting to parse those sections of the config. When the parse error was encountered, there was no logic in place to cleanly stop all threads that might have been started. The lingering threads were preventing the process from terminating. This PR addresses that issue by catching exceptions encountered while starting Worker and WorkerIndexer threads and, in case an exception is encountered, calling stop() on the corresponding thread pool.

Update: after a closer look, it turned out the issue with WorkerIndexer not terminating was being caused by a lingering thread that got created by instantiating a java.util.Timer as an instance variable, and in some error, cases not calling cancel() on the timer. This problem was fixed by delaying timer instantiation to the WorkerIndexer.start() method and passing a flag to make the timer threads "daeomon threads" that won't prevent JVM termination. The earlier code changes in this PR to call stop() in error scenarios are still beneficial.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerIndexerPool.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerIndexerPool.java
@@ -56,11 +56,11 @@ public class WorkerIndexerPool {
         workerIndexer.start(config, pipelineName, bypassSearchEngine, idSet);
         workerIndexers.add(workerIndexer);
       } catch (Exception e) {
-        log.error("Exception caught when starting WorkerIndexer thread {}; aborting", i+1, e);
+        log.error("Exception caught when starting WorkerIndexer thread {}; aborting", i+1);
         try {
           stop();
         } catch (Exception e2) {
-          log.error("Exception caught when attempting to stop WorkerIndexer threads because of a startup problem", e);
+          log.error("Exception caught when attempting to stop WorkerIndexer threads because of a startup problem", e2);
         }
         throw e;
       }
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerIndexerPool.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerIndexerPool.java
@@ -84,7 +84,9 @@ public class WorkerIndexerPool {
 
   public void stop() throws Exception {
     log.debug("Stopping " + workerIndexers.size() + " worker threads");
-    logTimer.cancel();
+    if (logTimer != null) {
+      logTimer.cancel();
+    }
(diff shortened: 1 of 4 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerIndexerPool.java
```diff
@@ -49,12 +49,26 @@ public void start() throws Exception {
...
+      try {
+        WorkerIndexer workerIndexer = new WorkerIndexer();
+        workerIndexer.start(config, pipelineName, bypassSearchEngine, idSet);
+        workerIndexers.add(workerIndexer);
+      } catch (Exception e) {
+        log.error("Exception caught when starting WorkerIndexer thread {}; aborting", i+1, e);
+        try {
+          stop();
+        } catch (Exception e2) {
+          log.error("Exception caught when attempting to stop WorkerIndexer threads because of a startup problem", e);
```
akul-sethi, line 63: I believe this should be e2

lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerPool.java
```diff
@@ -53,11 +53,25 @@ public void start() throws Exception {
...
-      threads.add(Worker.startThread(config, messenger, pipelineName, metricsPrefix));
+      try {
+        WorkerMessenger messenger = workerMessengerFactory.create();
+        threads.add(Worker.startThread(config, messenger, pipelineName, metricsPrefix));
+      } catch (Exception e) {
+        log.error("Exception caught when starting Worker thread {}; aborting", i+1);
+        try {
+          stop();
+        } catch (Exception e2) {
+          log.error("Exception caught when attempting to stop Worker threads because of a startup problem", e);
```
akul-sethi, line 66: i believe it should be e2

akul-sethi (approved):
looks good to me!

lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerIndexerPool.java
rseitz: thx -- fixed

lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerPool.java
rseitz: thx -- fixed

PR #91, opened 2024-02-29
LC-428 added misc tests

akul-sethi:
added some tests that were needed from looking at the jacaco code coverage reports. 

a few bugs were found in SolrUtils.toDocument(Tuple tuple)

1. An exception was not thrown when the tuple did not contain Document.ID_FIELD since tuple.getString() does not return null when the field does not exist. 
2. It only worked when the tuple only contained strings because values were always casted to strings before adding to document
3. It re added all fields including the id field to the new document which is not allowed.

I fixed all the above errors and updated the Javadoc to reflect it. The only type from tuples which is not supported is maps since Lucille documents do not have a concept of nesting


lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
```diff
@@ -117,21 +119,58 @@ public static List<String> getSolrUrls(Config config) {
...
 
     for (Entry<String, Object> e : tuple.getFields().entrySet()) {
-      d.setOrAdd((String) e.getKey(), (String) e.getValue());
+      Object val = e.getValue();
+      String key = e.getKey();
+
+      if (key.equals(Document.ID_FIELD)) {
+        continue;
+      }
+      DocumentException exception = new DocumentException(
```
rseitz, line 145: suggest we don't instantiate this exception unless we're actually going to throw it
I would just write this out on line 168 and 172 -- the little bit of duplication that would be introduced between those lines seems ok to me, but if you really wanted you could avoid it with a private static getException helper method.

lucille-core/src/test/java/com/kmwllc/lucille/core/WorkerTest.java
```diff
@@ -0,0 +1,77 @@
...
+public class WorkerTest {
+
+  @Test
+  public void testMain() {
+    Map<WorkerPool, List<Object>> args = new HashMap<>();
+    Config config = ConfigFactory.load("WorkerTest/config.conf");
+    WorkerMessengerFactory mockFactory = new WorkerMessengerFactory() {
+      @Override
+      public WorkerMessenger create() {
+        // TODO Auto-generated method stub
```
rseitz, line 27: can remove this TODO and the one on l34 since no action is needed

lucille-core/src/test/java/com/kmwllc/lucille/core/WorkerTest.java
```diff
@@ -0,0 +1,75 @@
...
+      factory.when(() -> {
+        WorkerMessengerFactory.getKafkaFactory(config, "foo2");
+      }).thenReturn(mockFactory2);
+
+      try {
+        Worker.main(new String[0]);
+        Worker.main(new String[] {"foo"});
+      } catch (Exception e) {
+        // we do not want an exception to be thrown
+        assertTrue(false);
```
rseitz, line 57: the test will automatically fail if an exception is thrown, so we don't really need the try catch
if there's a reason why you want to be explicit, could use fail() instead of assertTrue(false)
could also use assertDoesNotThrow (need to check if it's available in the junit version we're on)
suggest the implicit approach of just letting the exception bubble up, but could add a comment with any info about what exception might possibly be thrown that we don't want

lucille-core/src/test/java/com/kmwllc/lucille/message/KafkaMessengerTest.java
```diff
@@ -0,0 +1,168 @@
...
+        KafkaUtils.getEventTopicName("foo2", doc.getRunId());
+      }).thenReturn("foo2_event_id");
+      KafkaWorkerMessenger messenger = new KafkaWorkerMessenger(config, "foo");
+      KafkaWorkerMessenger messenger2 = new KafkaWorkerMessenger(config2, "foo2");;
+      try {
+        messenger.sendEvent(doc, "message", Event.Type.FAIL);
+        assertTrue(true);
+      } catch (Exception e) {
+        // we want this to not throw an expection
+        assertTrue(false);
```
rseitz, line 150: same comment as elsewhere -- don't need the try/catch, just let any exceptions bubble up and fail the test

lucille-core/src/test/java/com/kmwllc/lucille/core/WorkerTest.java
rseitz: Looks like this has been addressed in KafkaMessengerTest -- thx -- but not here?

lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
```diff
@@ -117,21 +119,58 @@ public static List<String> getSolrUrls(Config config) {
...
     for (Entry<String, Object> e : tuple.getFields().entrySet()) {
-      d.setOrAdd((String) e.getKey(), (String) e.getValue());
+      Object val = e.getValue();
+      String key = e.getKey();
+
+      if (key.equals(Document.ID_FIELD)) {
+        continue;
+      }
+
+      if (val instanceof String) {
```
rseitz, line 146: This code could be expanded to cover the remaining types that Document supports and moved into a method on the Document interface itself:   `default void setOrAdd(String key, Object val) throws DocumentException`

That would be a big enough change to warrant moving this to a separate PR with its own ticket

I am trying to remember if there's a reason we didn't do something like this earlier. In any case, it seems better to have this type inference logic encapsulated in Document rather than living just in SolrUtils. LMK what you think.

rseitz: I was thinking of an expanded version of something like this on Document:

```
  default void setOrAdd(String key, Object val) throws DocumentException {
    if (val instanceof String) {
      setOrAdd(key, (String) val);
    } else if (val instanceof Long) {
      setOrAdd(key, (Long) val);
    } else if (val instanceof Double) {
      setOrAdd(key, (Double) val);
    } else if (val instanceof Boolean) {
      setOrAdd(key, (Boolean) val);
    } else if (val instanceof List<?>) {
      List<?> list = (List<?>) val;
      if (!list.isEmpty()) {
        if (list.get(0) instanceof String) {
          update(key, UpdateMode.APPEND, list.toArray(new String[0]));
        } else if (list.get(0) instanceof Long) {
          update(key, UpdateMode.APPEND, list.toArray(new Long[0]));
        } else if (list.get(0) instanceof Double) {
          update(key, UpdateMode.APPEND, list.toArray(new Double[0]));
        } else if (list.get(0) instanceof Boolean) {
          update(key, UpdateMode.APPEND, list.toArray(new Boolean[0]));
        } else {
          throw new DocumentException(String.format("Class %s is not supported in Document", val.getClass()));
        }
      }
    }
  }
```

lucille-core/src/test/java/com/kmwllc/lucille/core/WorkerTest.java
akul-sethi: done

lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
rseitz: I think the instanceof checks are, plain and simple, what we need here. I'm not thinking of a more elegant approach... The good news is that all this logic would be encapsulated in one method on Document. I was trying to recall whether we had type checking code like this elsewhere in the codebase, because I'm sure this need has arisen before, but I don't think we do. I would say go ahead with adding this as a default method on Document, and we can refactor that method later if we find a better approach. This work will also help us with an existing limitation of the DatabaseConnector -- currently it treats everything as a string. See https://github.com/kmwtechnology/lucille/blob/7418a2423a91b6eef8bdefc8cbd6783ea1d42ae0/lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java#L188

There's a related ticket for DatabaseConnector referencing a returnTypes array but that appears to be stale:
https://kmwllc.atlassian.net/browse/LC-220

akul-sethi: i refactored the SolrUtils.toDocument method to use the generic Document setters which have been added. The method now supports adding all types supported by Document as well as Lists and Arrays of those types.

lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
```diff
@@ -117,21 +117,44 @@ public static List<String> getSolrUrls(Config config) {
...
+
+    if (tuple.getFields().containsKey(Document.ID_FIELD)) {
       d = Document.create(tuple.getString(Document.ID_FIELD));
     } else {
       throw new DocumentException("Unable to create Document from Tuple. No id field present.");
     }
 
     for (Entry<String, Object> e : tuple.getFields().entrySet()) {
-      d.setOrAdd((String) e.getKey(), (String) e.getValue());
+      Object val = e.getValue();
```
rseitz, line 137: minor: move this after the `if (key.equals...` when we know we need it

lucille-core/src/test/java/com/kmwllc/lucille/core/WorkerTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+      factory.when(() -> {
+        WorkerMessengerFactory.getKafkaFactory(config, "foo");
+      }).thenReturn(mockFactory);
+
+      factory.when(() -> {
+        WorkerMessengerFactory.getKafkaFactory(config, "foo2");
+      }).thenReturn(mockFactory2);
+
+      Worker.main(new String[0]);
```
rseitz, line 52: main() is meant as the entry point for launching Worker as a java app/process so it's highly unlikely that main() would ever be called twice in a row like this. I see that this is testing main() in two cases, with and without args. I think this would be more natural as two separate tests... testMainWithArgs() and testMainWithoutArgs()...

PR #93, opened 2024-02-29
LC-419 added tests for ParseFloats stage

akul-sethi:
one "bug" I found was that the stage would throw an exception if the input string could not correctly be parsed to a float list. Since the design philosophy of lucille is to ingest as many documents as possible without crashing, I changed this to just log a warning and skip the document instead. 

NOTE: the stage currently will consider the empty list ("[]") as non-parsable and skip rather than treating it as an empty list of floats and removing the field. Let me know if I should change this behavior. 
FIXED ABOVE: the stage now will parse the empty list and since its behavior is to remove the source field and replace it, it effectively just deletes the source field since there is no way to add an empty list.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
@@ -7,10 +7,8 @@ import com.kmwllc.lucille.core.Stage;
 import com.kmwllc.lucille.core.StageException;
 import com.typesafe.config.Config;
 
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
-import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -31,7 +29,7 @@ public class ParseFloats extends Stage {
       List<Float> floats;
       try {
         ObjectMapper mapper = new ObjectMapper();
-        floats = mapper.readValue(value, new TypeReference<List<Float>>(){});
+        floats = mapper.readValue(value, new TypeReference<List<Float>>() {});
       } catch (Exception e) {
         log.warn("String: {} cannot be parsed.", value, e);
         return null;
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
@@ -1,5 +1,7 @@
 package com.kmwllc.lucille.stage;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.Stage;
 import com.kmwllc.lucille.core.StageException;
@@ -28,8 +30,8 @@ public class ParseFloats extends Stage {
       String value = doc.getString(this.field);
       List<Float> floats;
       try {
-        floats = Arrays.stream(value.replaceAll("\\[", "").replaceAll("\\]", "").replaceAll(",", "").split(" "))
-            .map(v -> Float.parseFloat(v)).collect(Collectors.toList());
+        ObjectMapper mapper = new ObjectMapper();
+        floats = mapper.readValue(value, new TypeReference<List<Float>>(){});
       } catch (Exception e) {
         log.warn("String: {} cannot be parsed.", value, e);
         return null;
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
@@ -16,6 +16,8 @@ public class ParseFloats extends Stage {
 
   private final String field;
   private static final Logger log = LogManager.getLogger(ParseFloats.class);
+  private static final ObjectMapper MAPPER = new ObjectMapper();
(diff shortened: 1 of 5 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
```diff
@@ -23,8 +26,14 @@ public ParseFloats(Config config) {
   public Iterator<Document> processDocument(Document doc) throws StageException {
     if (doc.has(field)) {
       String value = doc.getString(this.field);
-      List<Float> floats = Arrays.stream(value.replaceAll("\\[", "").replaceAll("\\]", "").replaceAll(",", "").split(" "))
-          .map(v -> Float.parseFloat(v)).collect(Collectors.toList());
+      List<Float> floats;
+      try {
+        floats = Arrays.stream(value.replaceAll("\\[", "").replaceAll("\\]", "").replaceAll(",", "").split(" "))
```
rseitz, line 31: I suspect that using Jackson to parse the string as json will be faster than applying these four regexes and creating new strings each time. The error reporting from jackson will probably also be better. Here is how it can be done:

```
    ObjectMapper mapper = new ObjectMapper(); // should be private static final
    String floatString = "[1.2, 3.4, 5.6]";
    Float[] floats = mapper.readValue(floatString, Float[].class); // if we want an array
    // IntelliJ will say TypeReference<List<Float>> can be TypeReference<> but be explicit so jackson can check types for us
    List<Float> floatList = mapper.readValue(floatString, new TypeReference<List<Float>>(){});
```

I know that this PR was meant for tests only
Up to you whether to asses performance of the jackson approach and do it here in this PR, or create a ticket to do it in a followup PR

akul-sethi: i changed it

lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
```diff
@@ -23,8 +26,14 @@ public ParseFloats(Config config) {
   public Iterator<Document> processDocument(Document doc) throws StageException {
     if (doc.has(field)) {
       String value = doc.getString(this.field);
-      List<Float> floats = Arrays.stream(value.replaceAll("\\[", "").replaceAll("\\]", "").replaceAll(",", "").split(" "))
-          .map(v -> Float.parseFloat(v)).collect(Collectors.toList());
+      List<Float> floats;
+      try {
+        ObjectMapper mapper = new ObjectMapper();
```
rseitz, line 31: both the ObjectMapper and the TypeReference should be private static final vars on the class. We don't need to create them with each call to processDocument()

akul-sethi: makes sense

lucille-core/src/test/java/com/kmwllc/lucille/stage/ParseFloatsTest.java
```diff
@@ -0,0 +1,52 @@
...
+    Stage stage = factory.get("ParseFloatsTest/basic.conf");
+
+    Document bad = Document.create("doc1");
+    Document empty = Document.create("doc2");
+    Document simple = Document.create("doc3");
+    Document noField = Document.create("doc4");
+
+    bad.setOrAdd("source", "not a float");
+    empty.setOrAdd("source", "[]");
+    simple.setOrAdd("source", "[1, 2, 3]");
```
rseitz, line 31: for completeness, I'd suggest giving one of the floats a decimal point... like 3.5 instead of 3
and adding a separate doc where the array has mostly floats but one bad value, i.e. [1, 2.3, "a", 4]

lucille-core/src/test/java/com/kmwllc/lucille/stage/ParseFloatsTest.java
```diff
@@ -0,0 +1,66 @@
...
+    Document simple = Document.create("doc3");
+    Document noField = Document.create("doc4");
+    Document floats = Document.create("doc5");
+    Document badList = Document.create("doc6");
+
+    bad.setOrAdd("source", "not a float");
+    empty.setOrAdd("source", "[]");
+    simple.setOrAdd("source", "[1, 2, 3]");
+    floats.setOrAdd("source", "[1.2, 2.5, 3.1]");
+    badList.setOrAdd("source", "[1, a, 3]");
```
rseitz, line 35: suggest putting the a in quotes so the test string is actually valid json
the "bad" case already covers invalid json

lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseFloats.java
```diff
@@ -23,8 +28,13 @@ public ParseFloats(Config config) {
...
-      List<Float> floats = Arrays.stream(value.replaceAll("\\[", "").replaceAll("\\]", "").replaceAll(",", "").split(" "))
-          .map(v -> Float.parseFloat(v)).collect(Collectors.toList());
+      List<Float> floats;
+      try {
+        floats = MAPPER.readValue(value, TYPE_REFERENCE);
+      } catch (Exception e) {
+        log.warn("String: {} cannot be parsed.", value, e);
+        return null;
+      }
       doc.removeField(this.field);
```
rseitz, line 38: would you please make a ticket to update this stage to use a "source" and "dest" approach like other stages.
It seems overly limit that this stage only supports overwriting the source field.
This work can be done later, in a separate PR.

rseitz (approved):
approving with two minor comments -- tweak to make test case use valid json, and request to open a ticket for making this stage accept a dest field which can be different from source

PR #94, opened 2024-03-01
LC-418 added tests for ElasticsearchLookup stage

akul-sethi:
Changes made: 
1. there appears to be a bug in the StageSpec involving requiredParents so I modified the stage spec passed to this stage for now so that the tests could page. I will look further into this.
2. the stage used to crash if the result from elastic search did not have the given key. I changed it to now skip that field. 

points which should be reviewed:
1. If destination fields and source fields do not have the same size the stage will crash during execution and there currently is no validation for this. 
2. If there is an exception thrown while getting from elastic search the stage will crash. We may not want this? 
3. NOTE-- this point is for Document code: if two documents are created from the same ObjectNode they will share changes made to both of them separately. We may want to change this.


lucille-core/src/test/java/com/kmwllc/lucille/stage/ElasticsearchLookupTest.java
```diff
@@ -0,0 +1,208 @@
...
+      mockedUtils.when(() -> ElasticsearchUtils.getElasticsearchOfficialClient(config6)).thenReturn(foundClient);
+
+      assertThrows(StageException.class, () -> {
+      new ElasticsearchLookup(config).processDocument(doc);
+      });
+      new ElasticsearchLookup(config2).processDocument(doc2);
+      new ElasticsearchLookup(config3).processDocument(doc3);
+      new ElasticsearchLookup(config4).processDocument(doc4);
+      new ElasticsearchLookup(config5).processDocument(doc5);
+      new ElasticsearchLookup(config6).processDocument(doc6);
```
rseitz, line 183: Difficult to read. Six different configs. Six different docs. Naming of configs and docs is numerical and doesn't give any clues about what cases they represent. To understand the setup, you have to go look in multiple different places above. No comments saying what the different cases are.

Could you reorganize this test and the others so that the setup logic for each case we're testing is grouped together and maybe there's a code comment saying what the case is trying to test? I think this could be done without any new code, just by reorganizing the code you've already written here.

Looks like very strong & comprehensive test coverage here... but we should make tests as easy to read as possible.

lucille-core/src/test/java/com/kmwllc/lucille/stage/ElasticsearchLookupTest.java
```diff
@@ -0,0 +1,212 @@
...
+    Config four = ConfigFactory.load("ElasticSearchLookupTest/noIndex.conf");
+    assertThrows(IllegalArgumentException.class, () -> new ElasticsearchLookup(one));
+    assertThrows(IllegalArgumentException.class, () -> new ElasticsearchLookup(two));
+    assertThrows(IllegalArgumentException.class, () -> new ElasticsearchLookup(three));
+    assertThrows(IllegalArgumentException.class, () -> new ElasticsearchLookup(four));
+  }
+
+  @Test
+  public void testStart() throws Exception {
+    Config config = ConfigFactory.parseResourcesAnySyntax("ElasticsearchLookupTest/append.conf");
```
rseitz, line 60: could you give the configs descriptive names here like you did in testStop(), testProcessDocument(), etc.?

lucille-core/src/test/java/com/kmwllc/lucille/stage/ElasticsearchLookupTest.java
```diff
@@ -0,0 +1,212 @@
...
+      fail("Exception not thrown");
+    } catch (StageException e) {
+      assertEquals(message, e.getMessage());
+    }
+
+  }
+
+  @Test
+  public void testMalformedConfigs() throws Exception {
+    Config one = ConfigFactory.load("ElasticSearchLookupTest/noSource.conf");
```
rseitz, line 48: could you give the configs descriptive names here like you did in testStop(), testProcessDocument(), etc.?

PR #95, opened 2024-03-02
LC-429 generic setField, addToField, update, and setOrAdd added to Document

akul-sethi:
I added all updater fields rather than just setOrAdd. I had to use the reflection api in order to maintain one set of type checking logic. 

QUESTION: Why do we not have default implementations of the update methods on the Document interface. All implementations of update do not require object state and only use interface methods so it is doable. My generic implementation of update is on the Document class and I could add the rest there as well.

NOTE: Currently all the generic methods throw a check Exception if they fail but I can change this to a RuntimeException or a more specific checked exception if you think that is better. Additionally, update currently modifies the document until it encounters an argument which it cannot support. This means that the method has side effects when an exception is thrown (I currently have this reflected in testUpdateFailsInMiddle). I think it makes sense to make sure that all variadic arguments are of the same type at runtime before updating the document to avoid this issue. This will also stop the issue of having lists of mixed types. If you agree that this is a good idea i will make the change.


lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -51,6 +53,67 @@ public interface Document {
 
   void update(String name, UpdateMode mode, byte[]... values);
 
+  /**
+   * @throws IllegalArgumentException if object is not supported*/
+  public default void update(String name, UpdateMode mode, Object... values) throws Exception {
```
rseitz, line 58: All places where we throw an exception on these new methods, let's make it a DocumentException. It's ok for it to be a checked exception in this case. This is a different scenario from the other setters that take a specific type. We wanted to prevent those type-specific setters from throwing checked exceptions because we didn't want people to have to wrap code in try/catch just to add data to a Document. That's why validateFieldNames throws IllegalArgumentException, a RuntimeException. But in the case where you're setting a generic Object, it's reasonable to expect something could go wrong and to catch a checked exception.

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -51,6 +53,67 @@ public interface Document {
...
+      clazz = Integer.class;
+    } else if (value instanceof Instant) {
+      clazz = Instant.class;
+    } else if (value instanceof byte[]) {
+      clazz = byte[].class;
+    } else {
+      throw new IllegalArgumentException(String.format("Type of Object: %s is not supported", value.toString()));
+    }
+
+    Document.class.getMethod(method, String.class, clazz).invoke(this, name, value);
```
rseitz, line 114: Java reflection has a performance cost. It may be getting faster with more recent JVMs but the cost is still there. Some pointers:

https://blogs.oracle.com/javamagazine/post/java-reflection-performance
https://www.baeldung.com/java-reflection-benefits-drawbacks#1-performance-overhead

All of the setters on Document should be considered performance-critical. They could be called in a loop. And there could be millions or billions of Documents processed in a run. So we should avoid reflection here. If there's a choice between reflection and code duplication, let's accept the duplication in favor of better performance.

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
akul-sethi: sounds good. What do you think about the other parts of the description? should I add default implementations for the update methods and also check to make sure that all objects to be added are of the same type?

rseitz: If you find that default update methods can be added to Document in a way that works for both implementations (JsonDocument and HashMapDocument), then yes, let's try out that refactor

yes to checking that objects are of the same type, let's try it -- I'd like to see the change

please note my other comment about avoiding reflection

akul-sethi:
I made the changes we talked about. I also made the exceptions IllegalArgumentException rather than a checked exception. I did not really see a benefit to a checked exception and adding it required me to make a custom ThrowingConsumer interface so I didn't do it

rseitz (approved):

kyle-kmw:
looks good to me.

PR #96, opened 2024-03-04
LC-415 when creating kafka event topic, don't fail if it already exists

rseitz:
https://kmwllc.atlassian.net/browse/LC-415


akul-sethi (approved):
looks good to me

PR #98, opened 2024-03-05
LC-431 Remove background image

rseitz:
This PR removes the current background image from the doc site landing page, as we don't want to redistribute this image.
As of this change, the landing page of the doc site will have a solid dark gray background and no image.
We can add a new background image later if desired.

Also removing the photo of the porridge bowl that comes with Docsy and is shown on the About page.
There is one remaining image in blog/news/first-post/features-sunset-get.png that is left in place for now since it is part of an example that shows how to use resources in Hugo.


meghanboyd (approved):
Approved, thank you Rudi!

PR #103, opened 2024-03-14
LC-423 added tests for ZKRetryCounter

akul-sethi:
found one "bug" which is not really a functional bug but seems to be unintended behavior.

for non-kafka documents, the "counterPath" which is generated has two backslashes in it rather than one (check line 116 of the tests). It seems as though this is supposed to be a path-like representation in which case the double backslash would not typically be intended.

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/core/ZKRetryCounterTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/core/ZKRetryCounterTest.java
@@ -47,12 +47,13 @@ public class ZKRetryCounterTest {
 
       KafkaDocument kafkaDoc = spy(new KafkaDocument(Document.create("kafkaDoc"), "topic", 1, 2, "key"));
       when(kafkaDoc.getRunId()).thenReturn("runId");
-      JsonDocument jsonDoc = new JsonDocument("jsonDoc", "runId2");
+      Document doc = Document.create("jsonDoc", "runId2");
+      assertTrue(!(doc instanceof KafkaDocument));
 
       ZKRetryCounter retry3 = new ZKRetryCounter(ConfigFactory.load("ZKRetryCounterTest/config.conf"));
       ArgumentCaptor<String> captor = ArgumentCaptor.forClass(String.class);
 
-      retry3.remove(jsonDoc);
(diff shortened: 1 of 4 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/core/ZKRetryCounterTest.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+  }
+
+  @Test 
+  public void testConstructor() {
+    try (MockedStatic<CuratorFrameworkFactory> curatorStatic = mockStatic(CuratorFrameworkFactory.class)) {
+      CuratorFramework curatorFramework = mock(CuratorFramework.class);
+      curatorStatic.when(() -> CuratorFrameworkFactory.newClient(eq("foo"), any())).thenReturn(curatorFramework);
+      new ZKRetryCounter(ConfigFactory.load("ZKRetryCounterTest/config.conf"));
+      verify(curatorFramework, times(1)).start();
```
rseitz, line 71: mocking lets us test things that aren't really part of the public contract of the method or constructor
in this case we're testing that the ZKRetryCounter constructor calls start on CuratorFramework but that's not really a guarantee that ZKRetryCounter aims to make. It would be reasonable for start to be called somewhere else or even for ZKRetryCounter to interact with zookeeper in a different way, not using CuratorFramework. There's a tradeoff here between improving our code coverage vs. making the tests too intrusive into implementation details that aren't part of what the method guarantees

lucille-core/src/test/java/com/kmwllc/lucille/core/ZKRetryCounterTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.mockito.ArgumentCaptor;
+import org.mockito.MockedConstruction;
+import org.mockito.MockedStatic;
+import com.typesafe.config.ConfigFactory;
+
+public class ZKRetryCounterTest {
+
+  @Test
+  public void testRemove() throws Exception {
+    String jsonCounterString = "/LucilleCounters/bar//NON_KAFKA/runId2/jsonDoc";
```
rseitz, line 35: do we need to change the // to / here?

PR #104, opened 2024-03-15
LC-433 changed database connector to support all object types

akul-sethi:


lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -185,10 +185,13 @@ public void execute(Publisher publisher) throws ConnectorException {
...
 
-          String fieldValue = rs.getString(i);
-          // log.info("Add Field {} Value {} -- Doc {}", fieldName, fieldValue, doc);
+          Object fieldValue = rs.getObject(i);
           if (fieldValue != null) {
-            doc.setOrAdd(fieldName, fieldValue);
+            try {
+              doc.setOrAdd(fieldName, fieldValue);
+            } catch (IllegalArgumentException e) {
+              log.warn("Error encountered while adding database object to Lucille document: {}");
```
rseitz, line 193: seems to be missing an argument to bind to the "{}" -- also should include e as last argument so the log event includes the full stacktrace

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -185,10 +185,13 @@ public void execute(Publisher publisher) throws ConnectorException {
...
 
-          String fieldValue = rs.getString(i);
-          // log.info("Add Field {} Value {} -- Doc {}", fieldName, fieldValue, doc);
+          Object fieldValue = rs.getObject(i);
           if (fieldValue != null) {
-            doc.setOrAdd(fieldName, fieldValue);
+            try {
+              doc.setOrAdd(fieldName, fieldValue);
+            } catch (IllegalArgumentException e) {
+              log.warn("Error encountered while adding database object to Lucille document: {}", e);
```
rseitz, line 193: the logger will automatically include e in the output, so it is not necessary to reference it via {}

use {} only when there a specific string or other value you want to include in the primary message

lucille-core/src/test/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnectorTest.java
```diff
@@ -44,7 +43,48 @@ public void initTestMode() throws Exception {
...
+
+    // start the connector
+    connector.execute(publisher);
+
+    // Confirm there were 3 results.
+    List<Document> docsSentForProcessing = messenger.getDocsSentForProcessing();
+    assertEquals(2, docsSentForProcessing.size());
+
+    assertEquals("1", docsSentForProcessing.get(0).getId());
+    assertEquals((Integer)3, docsSentForProcessing.get(0).getInt("int_field"));
```
rseitz, line 76: does this test break without the code change to DatabaseConnector (i.e. we'd like it to break but I'm wondering if it actually would). Calling getInt() on a JsonDocument field that had been set as a String should do an automatic conversions:

here's a test (not committed) that shows this:

```
  @Test
  public void testTypeConversion() {
    Document doc = createDocument("id1");
    doc.setField("field1", "1");
    assertEquals(Integer.valueOf(1), doc.getInt("field1"));
    assertEquals(String.class, doc.asMap().get("field1").getClass());
  }
```

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -185,10 +185,13 @@ public void execute(Publisher publisher) throws ConnectorException {
...
 
-          String fieldValue = rs.getString(i);
-          // log.info("Add Field {} Value {} -- Doc {}", fieldName, fieldValue, doc);
+          Object fieldValue = rs.getObject(i);
           if (fieldValue != null) {
-            doc.setOrAdd(fieldName, fieldValue);
+            try {
+              doc.setOrAdd(fieldName, fieldValue);
+            } catch (IllegalArgumentException e) {
+              log.warn("Error encountered while adding database object to Lucille document: ", e);
```
rseitz, line 193: the ": " at the end of the message isn't necessary

rseitz (approved):

PR #105, opened 2024-03-15
LC-434 csv indexer now flushes after send

akul-sethi:


lucille-core/src/test/java/com/kmwllc/lucille/indexer/CSVIndexerTest.java
```diff
@@ -68,4 +68,27 @@ public void testCSVIndexer() throws Exception {
...
+    Config config = ConfigFactory.load("CSVIndexerTest/config.conf");
+
+    Document doc = Document.create("doc1", "test_run");
+    doc.setField("f1","123");
+    doc.setField("f2","abc");
+    Document doc2 = Document.create("doc2", "test_run");
+    doc2.setField("f1","456");
+    doc2.setField("f2","def");
+
+    CSVIndexer indexer = new CSVIndexer(config, messenger, false,"testing");
```
rseitz, line 86: should probably call close() on the indexer after we've done Files.readAllLines. So lines 86 through 89 could be wrapped in a try/finally.  The Indexer.run() method that's used in the other test calls close, but sendToIndex doesn't. Should leave some kind of comment explaining how this test is different from the other -- otherwise someone could see them and think they're basically the same.

akul-sethi: done

rseitz: thx
indexer.closeConnection() should be in a finally block
looks like this PR got closed by mistake?

rseitz (approved):

PR #107, opened 2024-03-19
LC-417 Adds github action config to deploy to sonatype.

kiratraynor:


kwatters (approved):
looks good to me.

.github/workflows/release-to-sonatype.yml
```diff
@@ -0,0 +1,27 @@
...
+        required: true
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Maven Central Repository
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11'
```
rseitz, line 18: in .github/workflows/maven.yaml we specify java-version 17
should we use 17 here too?

kwatters: my preference is that we keep with java 11 compliance for now.. let's create another issue to track updating to a newer version of java.

meghanboyd: Created LC-442 to track the task of evaluating Lucille for Java 17 upgrade.

rseitz (approved):

PR #110, opened 2024-03-25
LC-445 and LC-444 Create action for releases to sonatype

kiratraynor:
- Added the sources jar to be built as part of the release.
- Changes `EventTest.testCreateEqualsHashcodeDeserialize` due to failure because of timing.
- LC-445: Edits release-to-sonatype.yml to create release that can be pushed to maven central with the pgp signing.
     - description will be updated with another ticket
- LC-444: Adds release-snapshot.yml that will run everytime something is merged into master. This will create a snapshot release. 
- LC-452: Make sure `mvn install` does not use gpg signing profile.


lucille-core/pom.xml
```diff
@@ -258,12 +258,6 @@
       <version>0.6.3</version>
     </dependency>
 
-    <dependency>
```
rseitz, line 261: Just curious to know any background on why this is being removed?

rseitz (approved):

PR #118, opened 2024-05-08
LC-462: stage conditions paradigm asserts that provided operator is valid

akul-sethi:
The Conditions constructor now requires the provided operator to be valid and throws a runtime exception otherwise. Since the condition object is created at validation time (before a run) this validates the conditions clauses "statically"

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Condition.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Condition.java
@@ -11,15 +11,27 @@ public class Condition implements Predicate<Document> {
 
   private final List<String> fields;
   private final Set<String> values;
-  private final String operator;
+  private final Operator operator;
+
+  private enum Operator {
+    MUST, MUST_NOT;
+
+    public static Operator getOperator(String op) {
+      switch(op) {
+        case "must": return Operator.MUST;
+        case "must_not": return Operator.MUST_NOT;
+        default: throw new IllegalArgumentException(op + " is not a legal value for operator");
+      }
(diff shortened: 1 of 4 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/Condition.java
```diff
@@ -11,15 +11,27 @@ public class Condition implements Predicate<Document> {
 
   private final List<String> fields;
   private final Set<String> values;
-  private final String operator;
+  private final Operator operator;
+
+  private enum Operator {
+    MUST, MUST_NOT;
+
+    public static Operator getOperator(String op) {
```
rseitz, line 19: Suggest renaming this to get()

Also, see this pattern where each enum value stores its string name: https://howtodoinjava.com/java/enum/java-enum-string-example/#5-reverse-lookup-get-enum-name-from-value

rseitz:
this test failed: StageTest.testProcessMultipleConditions
that's actually good
I believe it's because multipleConditions.conf has `operator = "must not"` which should now be illegal

this PR needs some dedicated test coverage for operator validation

also, we want to make sure that when someone runs lucille in validation mode, if there's an illegal operator in a stage condition, they still get useful output -- the validation error should say what's wrong

lucille-core/src/main/java/com/kmwllc/lucille/core/Condition.java
akul-sethi: did that

rseitz (approved):

PR #119, opened 2024-05-10
LC-463: Conditions framework now does not need values to be provided

akul-sethi:
If no values list is provided, then only the existence of the fields is used with the "must" or "must_not" operators. For example, with a fields list of ["foo", "bar"], all documents must have both fields with the "must" operator and must not have either field with the "must_not" operator.


rseitz (changes requested):
Is there a test for what happens when an invalid operator is provided?
Have you done a manual test of running lucille in validation mode when the the config includes an invalid condition operator -- what does the output look like?
Does this PR 119 supersede PR 118?

akul-sethi:
> Is there a test for what happens when an invalid operator is provided? Have you done a manual test of running lucille in validation mode when the the config includes an invalid condition operator -- what does the output look like? Does this PR 119 supersede PR 118?

akul: this PR should be merged after 118. There is a test in 18 which fails since the operator is not accepted. It should also throw an error in validation mode before an actual run starts.

rseitz:
What is the test that parses a stage config with an unrecognized operator and checks that an exception is thrown? File and line number? I'm not seeing it.

akul-sethi:
> What is the test that parses a stage config with an unrecognized operator and checks that an exception is thrown? File and line number? I'm not seeing it.

It is now on line 28 of ConfigValidationTest.java

akul-sethi:
> Is there a test for what happens when an invalid operator is provided? Have you done a manual test of running lucille in validation mode when the the config includes an invalid condition operator -- what does the output look like? Does this PR 119 supersede PR 118?

this is the output that is seen after running in validation mode with an invalid operator: 
```
24/05/13 13:56:43 INFO Runner: Configuration is invalid. Printing the list of exceptions for each pipeline
        Pipeline: simple_pipeline       Error count: 1
                Exception 1: operator is invalid. It must take the value of `must` or `must_not`
```

rseitz (approved):

PR #122, opened 2024-05-20
LC-314: add Document support for JSON getters and setters

rseitz:
The Document interface has a single-value json setter, but not corresponding getters or updaters.
There's `void setField(String name, JsonNode value)` but no corresponding `JsonNode getJson(String name)`
This PR fleshes out Document's support for JSON values by adding the standard getters and updaters.

The complication with Json values is that JsonDocument internally stores multivalued fields as JsonArrays. For example, if you add 1, 2, 3, 4 to a multivalued int field, JsonDocument will represent this as a JsonArray containing those ints. But what should happen if someone _adds_ a JsonArray containing ints 1, 2, 3,4 via setField(String name, JsonNode value)? Should JsonDocument treat this as a multivalued field containing 4 ints, or should it treat this as a single-valued field containing one JsonArray? 

The key decisions in this PR are that 
   1) getJson(name) will always return the actual json that JsonDocument keeps internally to represent that field value or values, without doing any transformation to figure out whether there's one value, or multiple values -- getJson just gives you back whatever's there in the JsonDocument's internal storage
  2) if JsonDocument document is storing a JsonArray in a field, it will always report true for isMultiValued, no matter whether that JsonArray got added directly (via a json setter) or indirectly (by calling addToField several times to store multiple values in the field)

One thing to note is that the test coverage in this PR has been added in JsonDocumentTest, not in the generic DocumentTest that runs with both a JsonDocument and a HashMapDocument. The semantics of json-handling in HashMapDocument are different as of this PR, and not yet tested. HashMapDocument isn't currently used, and we have to decide what to do with it...

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/core/JsonDocumentTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/core/JsonDocumentTest.java
@@ -1,6 +1,8 @@
 package com.kmwllc.lucille.core;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.time.Instant;
 import org.junit.Test;
@@ -11,7 +13,9 @@ import java.util.function.UnaryOperator;
 
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
 
 public class JsonDocumentTest extends DocumentTest.NodeDocumentTest {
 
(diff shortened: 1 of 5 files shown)


akul-sethi (commented):
typo: should be arrayOfOjbect1AndObject2

lucille-core/src/test/java/com/kmwllc/lucille/core/JsonDocumentTest.java
```diff
@@ -125,4 +138,66 @@ public void testTypeConversion() {
...
+
+
+    d.addToField("jsonObject1", jsonObject2);
+    assertTrue(d.isMultiValued("jsonObject1"));
+
+    JsonNode jsonArray1WithObject2Added = mapper.readTree("[{\"a\":1}, {\"b\": 2}, {\"c\":3, \"d\": 4}]");
+    d.addToField("jsonArray1", jsonObject2);
+    assertEquals(jsonArray1WithObject2Added, d.getJson("jsonArray1"));
+
+    JsonNode arrayOfOjbect1AndObject2 = mapper.readTree("[{\"a\":1, \"b\": 2}, {\"c\":3, \"d\": 4}]");
```
akul-sethi, line 179: should be arrayOfOjbect1andObject2

meghanboyd (dismissed):
Approving since I see that the primary feedback was to resolve a typo which Rudi has since fixed.

kiratraynor (approved):

akul-sethi (approved):
looks good to me

PR #125, opened 2024-05-28
LC-436: added smoketest to lucille

akul-sethi:
This smoketest uses the failsafe plugin to run an integration test. The integration test simply goes through a full lucille run in file to file mode similar to the file-to-file-example, but it does so on the packaged jar rather than compiled class files


lucille-core/src/test/resources/MainIT/source.ndjson
```diff
@@ -0,0 +1,3 @@
+{"id":"3", "name":"Burger", "price":30, "country":"USA"}
```
rseitz, line 1: it looks like .jsonl is a more standard/common file extension for newline-delimited json than .ndjson
so let's rename this file to source.jsonl
there are a few other .ndjson files in the project that will need to be renamed but that can be done in a separate ticket

lucille-core/src/test/java/com/kmwllc/lucille/core/MainIT.java
```diff
@@ -0,0 +1,32 @@
...
+    assertEquals("\"my_name\",\"my_country\",\"my_price\"", reader.readLine());
+    assertEquals("\"Carbonara\",\"Italy\",\"30\"", reader.readLine());
+    assertEquals("\"Pizza\",\"Italy\",\"10\"", reader.readLine());
+    assertEquals("\"Tofu Soup\",\"Korea\",\"12\"", reader.readLine());
+    assertEquals("\"my_name\",\"my_country\",\"my_price\"", reader.readLine());
+    assertEquals("\"Burger\",\"USA\",\"30\"", reader.readLine());
+    assertEquals("\"Salad\",\"USA\",\"10\"", reader.readLine());
+    assertEquals("\"Wine\",\"France\",\"12\"", reader.readLine());
+    reader.close();
+    assertTrue(true);
```
rseitz, line 30: we don't need this assertion
if reader.close() throws an exception the test will fail

lucille-core/pom.xml
```diff
@@ -305,6 +305,42 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
```
rseitz, line 308: looks like this still needs the config added to make sure we're using the jar

       <configuration>
          <classesDirectory>target/fake</classesDirectory>

etc.

lucille-core/src/test/resources/MainIT/source.ndjson
meghanboyd: @rseitz @akul-sethi FYI ticket to rename all .ndjson files to .jsonl is LC-476

rseitz:
I ran `mvn verify` and it worked! But it looks like the lucille-core/output/dest.csv file that got generated was never cleaned up. So, after running `mvn verify` I get that file showing up as a new file in my git changelist. Could you update this PR to make sure the file is cleaned up?

akul-sethi:
> I ran `mvn verify` and it worked! But it looks like the lucille-core/output/dest.csv file that got generated was never cleaned up. So, after running `mvn verify` I get that file showing up as a new file in my git changelist. Could you update this PR to make sure the file is cleaned up?

done

rseitz:
tried it again but still had `lucille-core/output/dest.csv`  left over after `mvn clean verify` run at top level of repo

lucille-core/src/test/java/com/kmwllc/lucille/core/FileToFileIngestIT.java
```diff
@@ -0,0 +1,33 @@
...
+import com.typesafe.config.ConfigFactory;
+
+public class FileToFileIngestIT {
+
+  @Test
+  public void fullIT() throws Exception {
+    File f = null;
+    try {
+      Config config = ConfigFactory.parseFile(new File("src/test/resources/FileToFileIngestIT/file-to-file-example.conf"));
+      Runner.runInTestMode(config);
```
rseitz, line 18: we're cleaning up the file at the end of the test, which is nice and self-contained
why not also delete the file at the beginning in case it exists, rather than relying on a maven pre-integration phase?

rseitz (approved):

PR #126, opened 2024-05-31
LC-421: misc test improvement

akul-sethi:
added and improved tests for the following stages: Print, ReplacePatterns, and QueryDatabase


lucille-core/src/test/java/com/kmwllc/lucille/stage/ReplacePatternsTest.java
```diff
@@ -38,6 +38,46 @@ public void testReplacePatterns() throws Exception {
...
+    stage.processDocument(doc);
+    assertEquals("The term REPLACED should be replaced.", doc.getStringList("output1").get(0));
+  }
+
+  @Test 
+  public void testMultiline() throws Exception {
+    Stage stage = factory.get("ReplacePatternsTest/two.conf");
+
+    Document doc = Document.create("doc");
+    doc.setField("input1", "The term fAlse should be replaced,\n but not this false");
```
rseitz, line 56: This test is focusing on multiline handline not case-insensitivity so I'd suggest changing the first "fAlse" to "false" to simplify/clarify the scenario. Given that there's a separate test for case insensitivity. Please give the conf files descriptive names if possible, not one.conf, two.conf, three.conf etc.

rseitz (approved):

PR #127, opened 2024-06-03
LC-427: improved tests for IndexerFactory

akul-sethi:


rseitz:
seems like this PR also has the LC-421 changes?

PR #128, opened 2024-06-03
LC-479 ParseJson stage should not delete the src field

rseitz:
ParseJson currently deletes the src field after parsing the json contained in it. If a user wants to remove src, they can do it with a DeleteFields stage. This PR updates ParseJson so it leaves the src field in place. This leaves the choice up to the user re: whether to delete src later via DeleteFields.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseJson.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ParseJson.java
@@ -62,10 +62,6 @@ public class ParseJson extends Stage {
     this.jsonParseCtx = JsonPath.using(this.jsonPathConf);
   }
 
-  /**
-   * @param doc
-   * @return
-   */
   @Override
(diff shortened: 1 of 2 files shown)


kiratraynor:
It looks like this keeps the src field in all cases. Not sure if there was a specific reason it was deleted before, ie. if there was a case that we would actually want to remove the src field, should we add that as a config parameter? Or if there isn't really a case for that then I think this is good.

rseitz:
> It looks like this keeps the src field in all cases. Not sure if there was a specific reason it was deleted before, ie. if there was a case that we would actually want to remove the src field, should we add that as a config parameter? Or if there isn't really a case for that then I think this is good.

If a user needs to delete the src field, they can use a DeleteFields stage. If we add an option to delete src here, we get into having to add similar options consistently elsewhere; easiest to let the user add DeleteFields later in the pipeline if they want.

kiratraynor (approved):

PR #129, opened 2024-06-03
LC-480 update FetchURI stage to accept headers specified in stage config

rseitz:


lucille-core/src/test/resources/FetchUriTest/headers.conf
```diff
@@ -0,0 +1,10 @@
+{
+  name = "FetchUriTest"
+  class = "com.kmwllc.lucille.stage.FetchUri"
+  source = "url"
+  dest = "url_data"
+  headers {
```
sartian, line 6: what file format is this? would expect a : after headers but there isn't any, see = in one place and : in other places... is this yaml?

sartian: If this is HOCON, I think this should be... 
```
{
  name = "FetchUriTest"
  class = "com.kmwllc.lucille.stage.FetchUri"
  source = "url"
  dest = "url_data"
  headers = {
    header1-name = "header1-value"
    header2-name = "header2-value"
  }
}
```

rseitz: All lucille *.conf files are HOCON, yes.
Both of our versions are valid HOCON, and my unit test that uses this conf passes.
Your version is more consistent with using "=" everywhere, but we do it similar to mine in several other tests. 
I guess that's what one gets with a super-lenient format, lots of stuff is legal: https://github.com/lightbend/config?tab=readme-ov-file#examples-of-hocon
I'll put the equals in as it's less surprising that way

kiratraynor (approved):

PR #132, opened 2024-06-07
LC-472 - Remove lucille.sh.

kiratraynor:
[LC-472](https://kmwllc.atlassian.net/browse/LC-472): The lucille.sh script no longer works properly so we just want to remove to from the codebase for now. In the future, there may be some benefit to adding another script to show how you could configure a run.


kwatters (approved):

akul-sethi (approved):
looks good to me

rseitz (approved):

PR #133, opened 2024-06-07
LC-254: CleanText Stage

akul-sethi:
This stage removes diacritics and accents from strings


lucille-core/src/main/java/com/kmwllc/lucille/stage/RemoveDiacritics.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Iterator;
+import com.kmwllc.lucille.core.ConfigUtils;
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.Stage;
+import com.kmwllc.lucille.core.StageException;
+import com.typesafe.config.Config;
+
+
+/**
+ * Adds random data to a document field given parameters
```
rseitz, line 13: comment seems to be copy-pasted from another stage?

rseitz (approved):

PR #135, opened 2024-06-13
LC-486 Add retries to FetchUri Stage

kiratraynor:
[LC-486](https://kmwllc.atlassian.net/browse/LC-486)

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/ConfigUtils.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/ConfigUtils.java
@@ -1,8 +1,11 @@
 package com.kmwllc.lucille.core;
 
 import com.typesafe.config.Config;
-import com.typesafe.config.ConfigException;
-import com.typesafe.config.ConfigFactory;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.apache.http.Header;
+import org.apache.http.message.BasicHeader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -30,4 +33,17 @@ public class ConfigUtils {
     return fallback;
   }
 
+  /**
+   * Creates an array of org.apache.http.Headers given a map of header keys and values which can be used for http requests.
+   *
+   * @param map map containing the header values
+   * @return array of Header objects
+   */
+  public static Header[] createHeaderArray(Map<String, Object> map) {
+    List<Header> headerList = new ArrayList<>();
+    for (Map.Entry<String, Object> entry : map.entrySet()) {
+      headerList.add(new BasicHeader(entry.getKey(), (String) entry.getValue()));
+    }
+    return headerList.toArray(new Header[0]);
+  }
 }
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/ExponentialBackoffRetryHandler.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/ExponentialBackoffRetryHandler.java
@@ -1,0 +1,39 @@
+package com.kmwllc.lucille.core;
+
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import org.apache.http.client.HttpRequestRetryHandler;
+import org.apache.http.protocol.HttpContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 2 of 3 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/core/ExponentialBackoffRetryHandlerTest.java
```diff
@@ -0,0 +1,72 @@
...
+        .create()
+        .addInterceptorFirst((HttpRequestInterceptor) (request, context) -> requestCounter++)
+        .setRetryHandler(new ExponentialBackoffRetryHandler(2, 500, 10000))
+        .addInterceptorLast((HttpResponseInterceptor) (response, context) -> { throw new IOException(); })
+        .build();
+  }
+
+  @Test
+  public void test2retries() {
+    createFailingHttpClient();
```
rseitz, line 40: Since we only call createFailingHttpClient() once, can we just do the creation of the HttpClient inline here and not have a instance variable that multiple tests methods update? Can we use a local counter variable rather than having multiple test methods share the same counter?

lucille-core/src/test/java/com/kmwllc/lucille/core/ExponentialBackoffRetryHandlerTest.java
```diff
@@ -0,0 +1,72 @@
...
+    this.httpClient = HttpClientBuilder
+        .create()
+        .addInterceptorFirst((HttpRequestInterceptor) (request, context) -> requestCounter++)
+        .setRetryHandler(new ExponentialBackoffRetryHandler(0, 1000, 10000))
+        .build();
+  }
+
+  @Test
+  public void testNoRetries() throws IOException {
+    createHttpClientNoRetries();
```
rseitz, line 55: Similar comment to above -- can we create the HttpClient inline here rather than having a separate method that updates an instance variable?

lucille-core/src/main/java/com/kmwllc/lucille/core/ExponentialBackoffRetryHandler.java
```diff
@@ -0,0 +1,39 @@
...
+    this.maxExpiry = maxExpiry;
+  }
+
+  @Override
+  public boolean retryRequest(IOException exception, int numRetries, HttpContext httpContext) {
+    if (numRetries > this.maxNumRetries) {
+      return false;
+    } else {
+      long delay = (long) Math.min(this.initialExpiry * Math.pow(2, numRetries - 1), this.maxExpiry);
+      log.info("Retrying request........waiting {}ms", delay);
```
rseitz, line 29: Should we find a way to include the URL in the log message? I'm thinking that if someone sees the current message in the logs, they'll have to do some piecing-together to figure out what request is actually being retried here, if that would even be possible.

lucille-core/src/main/java/com/kmwllc/lucille/core/ConfigUtils.java
```diff
@@ -30,4 +33,17 @@ public static <T> T getOrDefault(Config config, String setting, T fallback) {
     return fallback;
   }
 
+  /**
+   * Creates an array of org.apache.http.Headers given a map of header keys and values which can be used for http requests.
+   *
+   * @param map map containing the header values
+   * @return array of Header objects
+   */
+  public static Header[] createHeaderArray(Map<String, Object> map) {
```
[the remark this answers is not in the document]
kiratraynor: Okay that makes sense, do you think it's worth keeping the original createHeaderArray(Map) at all or I can just replace with what signature you have above createHeaderArra(config, name)?

rseitz (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/FetchUri.java
```diff
@@ -75,7 +74,10 @@ void setClient(CloseableHttpClient client) {
 
   @Override
   public void start() throws StageException {
-    client = HttpClients.createDefault();
+    client = HttpClientBuilder
+        .create()
+        .setRetryHandler(new ExponentialBackoffRetryHandler(this.maxNumRetries, this.initialExpiry, this.maxExpiry))
```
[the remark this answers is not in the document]
meghanboyd: @rseitz @kiratraynor we will follow up on this in LC-502

rseitz:
Note: PR was merged despite build failure; integration test failure was examined, determined unrelated to code changes in this PR, and being addressed separately

PR #140, opened 2024-06-18
LC-436: fixes for smoketest

akul-sethi:


lucille-core/src/test/java/com/kmwllc/lucille/core/FileToFileIngestIT.java
```diff
@@ -30,6 +33,7 @@ public void fullIT() throws Exception {
       assertEquals("\"Wine\",\"France\",\"12\"", lines.get(7));
     } finally {
       f.delete();
```
rseitz, line 35: f could be null if we fail on an earlier line (like we aren't able to create the directory)

rseitz (approved):

PR #143, opened 2024-06-21
LC-422: ElasticsearchUtilsTest

akul-sethi:
This ticket has not been completed as mockito currently does not allow for argument capture for static methods. The tests can still be made using mockito's ```mockStatic``` method but it will most likely be less precise.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/util/ElasticsearchUtils.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/util/ElasticsearchUtils.java
@@ -11,10 +11,7 @@ import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
 import org.apache.http.impl.client.BasicCredentialsProvider;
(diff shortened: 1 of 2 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/util/ElasticsearchUtils.java
```diff
@@ -23,58 +20,6 @@
  */
 public class ElasticsearchUtils {
```
Jozurf, line 22: Removed getElasticsearchRestClient as RestHighLevelClient is deprecated and method is not being used anywhere else in Lucille.

lucille-core/src/test/java/com/kmwllc/lucille/util/ElasticsearchUtilsTest.java
```diff
@@ -0,0 +1,180 @@
+package com.kmwllc.lucille.util;
+
+import static com.kmwllc.lucille.util.ElasticsearchUtils.getAllowInvalidCert;
```
rseitz, line 3: lets remove this static import

rseitz (approved):

PR #145, opened 2024-07-15
LC-469: Lucille Admin API POC

spencersolomon6:
- Implemented a first pass of the Lucille Admin API with endpoints for readiness, liveness and starting a lucille run and viewing its status
- Included unit testing for the API Endpoints
- Still some open questions around unit testing and testing the asynchronous processes


lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -0,0 +1,56 @@
...
+  protected static boolean isRunning = false;
+
+  private RunnerManager() {}
+
+  public static RunnerManager getInstance() {
+    return instance;
+  }
+
+  public boolean isRunning() {
+    return isRunning;
```
rseitz, line 26: Do we need the isRunning variable?
Could we instead return (future != null && !future.isDone()) here?

lucille-admin/README.md
```diff
@@ -0,0 +1,51 @@
+# Lucille Admin API
+
+## Running the API
+
+1. Navigate into the `lucille-admin` module
+2. Run `mvn clean install` to generate the admin jar
+
+The API can be ran locally or in a Docker container:
```
rseitz, line 8: can be run

lucille-admin/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+              <groupId>org.slf4j</groupId>
+              <artifactId>log4j-over-slf4j</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+<!--            <plugin>
```
rseitz, line 78: is there a reason why these are commented out (but not deleted)?

rseitz:
build error seen twice:

>  Error:    The project com.kmwllc:lucille-admin:0.2.3-SNAPSHOT (/Users/runner/work/lucille/lucille/lucille-admin/pom.xml) has 1 error
> Error:      Non-resolvable parent POM for com.kmwllc:lucille-admin:0.2.3-SNAPSHOT: The following artifacts could not be resolved: com.kmwllc:lucille-parent:pom:0.2.3-SNAPSHOT (absent): Could not find artifact com.kmwllc:lucille-parent:pom:0.2.3-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 9, column 13 -> [Help 2]

lucille-admin/pom.xml
rseitz, line 1: I see this as an optional component of lucille that depends on lucille-core and can be used if needed... in other words, it seems like a plugin to me, so I'd suggest putting it under lucille-plugins

lucille-admin/src/main/java/com/kmwllc/lucille/objects/RunStatus.java
```diff
@@ -0,0 +1,43 @@
...
+
+  @Override
+  public boolean equals(Object o) {
+    if (!(o instanceof RunStatus)) {
+      return false;
+    }
+
+    RunStatus status = (RunStatus) o;
+
+    return this.isRunning == status.isRunning && this.runId.equals(status.runId);
```
rseitz, line 35: if runId can ever be null (i.e. via new RunStatus(null, false)) this would give a nullpointerexception at this.runId.equals()

lucille-admin/src/test/java/endpoints/LucilleAdminResourceTest.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Test
+  public void testRun() {
+    LucilleAdminResource admin = new LucilleAdminResource(runnerManager);
+
+    Response status = Response.status(404).build();
+    try {
+      status = admin.startRun();
+    } catch (Exception e) {
```
rseitz, line 43: don't use try/catch here. let the exception bubble up (and fail the test if it occurs)

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -218,6 +192,46 @@ public static Map<String, List<Exception>> validatePipelines(Config config) thro
     return exceptionMap;
   }
 
+  /**
+   * Derives the RunType for the new run from the 'useKafka' and 'local' parameters.
+   */
+  protected static RunType getRunType(boolean useKafka, boolean local) {
```
rseitz, line 198: can be package-private (leave off protected keyword)

spencersolomon6:
Change the name of the module to lucille-api and remove old references

lucille-admin/src/test/java/endpoints/LucilleAdminResourceTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class LucilleAdminResourceTest {
+
+  private final RunnerManager runnerManager = RunnerManager.getInstance();
+
+  @Before
+  public void setUp() {
+    try {
+      Thread.sleep(500);
```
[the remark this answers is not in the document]
spencersolomon6: This is necessary to ensure that previous runs have finished before checking the status or kicking off a new one. The pipeline being run is basically an empty pipeline, so I think 500ms should be plenty of time for it to complete in any environment.

lucille-core/src/test/java/com/kmwllc/lucille/indexer/NoOpIndexer.java
```diff
@@ -0,0 +1,27 @@
+package com.kmwllc.lucille.indexer;
+
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.Indexer;
+import com.kmwllc.lucille.message.IndexerMessenger;
+import com.typesafe.config.Config;
+import java.util.List;
+
+public class NoOpIndexer extends Indexer {
```
rseitz, line 9: I'm actually thinking we should put this in the main tree as opposed to test
there are enough cases where someone wants to run an ingest without actually sending to a search engine
we say to configure a regular indexer and set sendEnabled=false, but I'm coming around to thinking a NoOpIndexer might be a better way. We have a stage in the test tree called NoopStage without second o in caps, should probably be consistent and call this NoopIndexer. I think I actually prefer NopIndexer and NopStage -- thoughts?

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -0,0 +1,81 @@
...
+  synchronized public boolean isRunning() {
+    return !future.isDone();
+  }
+
+  /**
+   * Blocks the calling thread until the current lucille run is completed. This method is primarily intended to be used for testing,
+   * but has been made public to facilitate testing in other modules, namely lucille-plugins/lucille-api. NOTE: This method will block
+   * all other synchronized methods in the class.
+   */
+  synchronized public void waitForRunCompletion() throws ExecutionException, InterruptedException {
```
kyle-kmw, line 39: Does this need to be synchronized? I think it shouldn't be.

kyle-kmw (approved):

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -0,0 +1,53 @@
...
+    RunnerManager runnerManager = RunnerManager.getInstance();
+    Config config = ConfigFactory.load("RunnerManagerTest/sleep.conf");
+
+    // Ensure no lucille run is running at the start of the test
+    assertFalse(runnerManager.isRunning());
+
+    // Kick off a lucille run and ensure it is not skipped
+    assertTrue(runnerManager.runWithConfig(config));
+
+    // While we are sleeping, ensure lucille isRunning and a new run is skipped
```
rseitz, line 26: update comment -- we're not sleeping here

lucille-plugins/lucille-api/README.md
```diff
@@ -0,0 +1,51 @@
...
+## Running the API
+
+1. Navigate into the `lucille-admin` module
+2. Run `mvn clean install` to generate the admin jar
+
+The API can be run locally or in a Docker container:
+
+### Local Run:
+1. Replace the placeholders with your config file names or export them as environment variables. See [Configuration](#Configuration) for more information
+2. `java -Dconfig.file=${LUCILLE_CONF} -jar target/lucille-api-0.2.3-SNAPSHOT.jar server ${DROPWIZARD_CONF}`
```
rseitz, line 12: should probably replace the 0.2.3 in the jar name with a placeholder like "<version>" because this is going to go out of date and we don't want to have to update it with each release

lucille-plugins/lucille-api/src/test/java/APIIntegrationTest.java
```diff
@@ -0,0 +1,47 @@
...
+import io.dropwizard.testing.junit.DropwizardAppRule;
+import jakarta.ws.rs.client.Client;
+import jakarta.ws.rs.core.Response;
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class APIIntegrationTest {
+  // TODO : Improve Integration testing with better mocking
```
rseitz, line 15: is there something specific we want to do here re: improvement? if so, let's create a ticket. 
suggest removing the TODO comment

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -0,0 +1,53 @@
...
+    assertTrue(runnerManager.isRunning());
+    assertFalse(runnerManager.runWithConfig(config));
+
+    // Sleep until the run is over
+    runnerManager.waitForRunCompletion();
+
+    // Ensure lucille is not running and make sure we can now kick off a new run
+    assertFalse(runnerManager.isRunning());
+    assertTrue(runnerManager.runWithConfig(config));
```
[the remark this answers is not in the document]
rseitz: confirming that I think we need to do this. there are other tests elsewhere that assume no lucille threads are running when the test starts, and we don't want to break that assumption

lucille-plugins/lucille-api/src/test/java/APIIntegrationTest.java
spencersolomon6: I don't think there are any further improvements required anymore since the tests were switched to use a test config. The TODO has been removed.

rseitz (approved):

PR #153, opened 2024-07-24
LC-503: added ignoreFields functionality in Elasticsearch and OpenSearch Indexer

Jozurf:
- edited testRouting method in ElasticsearchTest and OpenSearchTest. Previously routing field was not deleted due to ignoreFields not implemented in both indexers.
- ignoreFields now allows routing to be appropriately set even though routing is in ignoreFields. The final document will not contain the routing field.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
@@ -99,7 +99,9 @@ public class OpenSearchIndexer extends Indexer {
     BulkRequest.Builder br = new BulkRequest.Builder();
 
     for (Document doc : documents) {
-      Map<String, Object> indexerDoc = doc.asMap();
+
+      // removing the fields mentioned in the ignoreFields setting in configurations
+      Map<String, Object> indexerDoc = getIndexerDoc(doc);
 
       // remove children documents field from indexer doc (processed from doc by addChildren method call below)
       indexerDoc.remove(Document.CHILDREN_FIELD);
@@ -110,8 +112,6 @@ public class OpenSearchIndexer extends Indexer {
 
       // handle special operations required to add children documents
       addChildren(doc, indexerDoc);
-
-      String routing = doc.getString(routingField);
       Long versionNum = (versionType == VersionType.External || versionType == VersionType.ExternalGte)
           ? ((KafkaDocument) doc).getOffset()
           : null;
@@ -121,7 +121,7 @@ public class OpenSearchIndexer extends Indexer {
             .update((up) -> {
               up.index(index).id(docId);
               if (routingField != null) {
-                up.routing(routing);
+                up.routing(doc.getString(routingField));
               }
               if (versionNum != null) {
                 up.versionType(versionType).version(versionNum);
@@ -133,7 +133,7 @@ public class OpenSearchIndexer extends Indexer {
             .index((up) -> {
               up.index(index).id(docId);
               if (routingField != null) {
-                up.routing(routing);
+                up.routing(doc.getString(routingField));
               }
               if (versionNum != null) {
                 up.versionType(versionType).version(versionNum);
--- a/lucille-core/src/test/java/com/kmwllc/lucille/indexer/ElasticsearchIndexerTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/indexer/ElasticsearchIndexerTest.java
@@ -34,6 +34,7 @@ import java.util.List;
 import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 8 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/indexer/OpenSearchIndexerTest.java
```diff
@@ -290,8 +293,12 @@ public void testRouting() throws Exception {
     Map<String, Object> map = (Map<String, Object>) indexRequest.document();
 
     assertEquals("doc1", indexRequest.id());
+
+    // routing has been set appropriately even though routing field has been deleted by ignoreFields
+    // note that id is still in document even though it was in ignoreFields
+    // as the id is later added back to the document
```
rseitz, line 299: when interacting with a live opensearch index, does the id show up as a document field in the search engine itself, or are we only seeing the id in the return value of indexRequest.document() in this particular testing context?

Jozurf: When documents are indexed into live opensearch index, both id & _id (same value) still shows up as a document field in the search engine itself.

lucille-core/src/test/resources/OpenSearchIndexerTest/ignoreFieldsWithOverride2.conf
```diff
@@ -0,0 +1,15 @@
+indexer {
+  type : "OpenSearch"
+  batchSize : 1
+  batchTimeout : 1000
+  logRate : 1000
+  ignoreFields : ["id",]
```
rseitz, line 6: trailing comma can be removed

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -121,7 +130,7 @@ protected void sendToIndex(List<Document> documents) throws Exception {
             .update((up) -> {
               up.index(index).id(docId);
               if (routingField != null) {
-                up.routing(routing);
+                up.routing(doc.getString(routingField));
```
spencersolomon6, line 133: Is there a specific reason to remove the routing variable here?

lucille-core/src/test/java/com/kmwllc/lucille/indexer/ElasticsearchIndexerTest.java
```diff
@@ -416,6 +420,149 @@ public void testBulkResponseErroring() throws Exception {
...
+    Document doc = Document.create("doc1");
+    doc.setField("ignoreField1", "value1");
+    doc.setField("ignoreField2", "value2");
+    doc.setField("normalField", "normalValue");
+
+    // checking that document fields has been set
+    assertNotNull(doc.getString("id"));
+    assertNotNull(doc.getString("ignoreField1"));
+    assertNotNull(doc.getString("ignoreField2"));
+    assertNotNull(doc.getString("normalField"));
```
spencersolomon6, line 443: Since setField is (hopefully!) being tested elsewhere in the codebase, I don't think its necessary to assert these fields exist after you set them.

If you agree we don't need these just remember to remove them for all the tests.

spencersolomon6 (approved):
Everything looks really good, especially regarding testing. Just left a couple minor notes.

spencersolomon6 (approved):
Everything looks really good, especially regarding testing. Just left a couple minor notes.

rseitz (approved):

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
Jozurf: Not really, to be honest. Previously the routing variable was defined before checking if routingField was null, and so my thought process was that it would be cleaner(?) to only be using the routingField once we knew it wasnt null, even though the previous iteration would still work. Looking at it again we are referencing the routing variable twice so maybe it would be better to revert back by defining the routing variable. What do you think?

lucille-core/src/test/java/com/kmwllc/lucille/indexer/ElasticsearchIndexerTest.java
Jozurf: yes those were tested in DocumentTest. Will remove those.

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
spencersolomon6: It's not a huge deal either way, especially since its only used twice. I think the change you made is definitely more readable, so you're good to leave it as is. Was mostly just curious if there was a specific reason for it.

PR #156, opened 2024-07-25
LC-504: delete foo file created from test

Jozurf:


rseitz:
Great. I ran mvn clean install locally using this branch and confirmed that the foo file was not left over after the run.

rseitz (approved):

PR #157, opened 2024-07-25
 LC-527: update jacoco version

Jozurf:


rseitz:
Noting that the github passed running java 17, and I got a clean local build with openjdk version "19.0.1" 2022-10-18. Jacoco docs suggest that earlier java versions remain supported, so this upgrade looks safe to me. https://www.jacoco.org/jacoco/trunk/doc/faq.html

rseitz (approved):

PR #158, opened 2024-07-26
LC-532: fix leak of threads in TextExtractor stage

Jozurf:
- closing of inputStream is in a separate branch LC-528. This branch changes how it got inputStream from filepath
- changed test for TextExtractor to use absolute path rather than relative path to reflect the same procedure. (doc.getString(filePathField) would return an absolute path)

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
@@ -68,7 +68,8 @@ public class FileUtils {
   /**
    * Takes a path and fsManager to retrieve the FileObject
    *
-   * @param path the path as a String
+   * @param path the path to file as a String
+   * @param fsManager the fileSystem manager instance
    * @return the FileObject to retrieve content from
    */
   public static FileObject getFileObject(String path, FileSystemManager fsManager) {
--- a/lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
@@ -8,9 +8,16 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.stream.Stream;
+import org.apache.commons.vfs2.FileContent;
+import org.apache.commons.vfs2.FileObject;
+import org.apache.commons.vfs2.FileSystemException;
+import org.apache.commons.vfs2.FileSystemManager;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public class FileUtils {
 
+  private static final Logger log = LogManager.getLogger(FileUtils.class);
   /**
    * Returns a Reader for the file at the given path. If the path begins with "classpath:" the prefix will be removed
    * and the file will be read from the classpath. If the path appears to be a URI, it will be accessed using VFS.
@@ -64,11 +71,47 @@ public class FileUtils {
    * @param path the path as a String
    * @return the correct InputStream
    */
-  public static InputStream getInputStream(String path) throws IOException {
+  public static InputStream getInputStream(String path, FileSystemManager fsManager) {
+    InputStream is;
+
+    // setting up VFS
     if (isValidURI(path)) {
-      return VFSInputStream.open(path);
+      FileObject file;
+      try {
+        file = fsManager.resolveFile(path);
+
+        if (file == null) {
+          log.warn("File object is null for path: {}", path);
+          return null;
+        }
+
+        if (!file.exists()) {
+          log.warn("File not found at: {}", path);
+          return null;
+        }
+
+        FileContent content = file.getContent();
+        try {
+          is = content.getInputStream();
+          return is;
+        } catch (FileSystemException e) {
+          log.warn("Error opening input stream for file: {}", path, e);
+          return null;
+        }
+      } catch (FileSystemException e) {
+        log.warn("Error retrieving file contents: {}", path, e);
+        return null;
+      }
     }
-    return new FileInputStream(path);
+
+    // setting up local file system manager
+    try {
+      is = new FileInputStream(path);
+      return is;
+    } catch (FileNotFoundException e) {
+      log.warn("File not found at: {}", path, e);
+      return null;
+    }
   }
 
   /**
--- a/lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
@@ -66,50 +66,20 @@ public class FileUtils {
   }
 
   /**
-   * Takes a path and returns the correct InputStream depending on whether the path is a URI or URL.
+   * Takes a path and fsManager to retrieve the FileObject
    *
    * @param path the path as a String
-   * @return the correct InputStream
+   * @return the FileObject to retrieve content from
    */
-  public static InputStream getInputStream(String path, FileSystemManager fsManager) {
-    InputStream is;
-
-    // setting up VFS
-    if (isValidURI(path)) {
-      FileObject file;
-      try {
-        file = fsManager.resolveFile(path);
-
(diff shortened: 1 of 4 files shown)


lucille-plugins/lucille-tika/src/test/java/com/kmwllc/lucille/tika/stage/TextExtractorTest.java
```diff
@@ -90,7 +110,13 @@ public void testExcel() throws StageException {
   public void testCustomTikaConfig() throws StageException {
     Stage stage = factory.get("TextExtractorTest/tika-config.conf");
     Document doc = Document.create("doc1");
-    doc.setField("path", "src/test/resources/TextExtractorTest/tika.txt");
+
+    // get current directory absolute path
+    Path currentRelativePath = Paths.get("");
```
rseitz, line 115: try creating a Path from the relative path string ("src/test/...") and then calling toAbsolutePath() on it -- can do this in one line

Jozurf: resolved in new commit

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -106,16 +125,44 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+      if (inputStream == null) {
+        log.warn("Error getting inputStream for file at : {}", filePath);
+        return null;
       }
+      parseInputStream(doc, inputStream);
     }
     return null;
   }
 
+  private InputStream getFileInputStream(String filePath) {
```
rseitz, line 138: The original FileUtils.getInputStream had extra functionality that we're not yet capturing here. If the filePath wasn't a valid URI it would treat it as a regular file path and open a FileInputStream...

lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
```diff
@@ -64,11 +71,47 @@ public static boolean isValidURI(String uriString) {
    * @param path the path as a String
```
rseitz, line 71: add new parameter

lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
```diff
@@ -64,11 +71,47 @@ public static boolean isValidURI(String uriString) {
...
+        if (!file.exists()) {
+          log.warn("File not found at: {}", path);
+          return null;
+        }
+
+        FileContent content = file.getContent();
+        try {
+          is = content.getInputStream();
+          return is;
+        } catch (FileSystemException e) {
```
rseitz, line 97: consider removing this inner try/catch since the outer try/catch is for the same exception type

lucille-core/src/main/java/com/kmwllc/lucille/util/FileUtils.java
```diff
@@ -64,11 +71,47 @@ public static boolean isValidURI(String uriString) {
...
+      FileObject file;
+      try {
+        file = fsManager.resolveFile(path);
+
+        if (file == null) {
+          log.warn("File object is null for path: {}", path);
+          return null;
+        }
+
+        if (!file.exists()) {
```
rseitz, line 88: is this existence test necessary, or if the file doesn't exist and we call file.getContent() will we get a FileSystemException that will be caught and logged

rseitz: looking at docs for the individual methods we're calling here, can we simplify the code as follows?

`      try {
        file = fsManager.resolveFile(path); // never returns null
        FileContent content = file.getContent(); // can be called even if !file.exists()
        return content.getInputStream(); // should throw FileSystemException if !file.exists()
      } catch (FileSystemException e) {
        log.warn("Error retrieving file contents: {}", path, e);
        return null;
      }`

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -103,14 +122,31 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+
+      // if file null, don't process document
+      if (file == null) {
+        return null;
+      }
+
+      // getting inputStream and catching errors gracefully
+      InputStream inputStream;
+      try {
+        inputStream = file.getContent().getInputStream();
```
rseitz, line 137: suggest adding comment that file.getContent() never returns null; can include link to vfs doc where it says this

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -103,14 +122,31 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+
+      parseInputStream(doc, inputStream);
 
+      // close file after, do not need to check if file is null as we have already checked on line 130
       try {
-        InputStream inputStream = FileUtils.getInputStream(filePath);
-        parseInputStream(doc, inputStream);
-      } catch (IOException e) {
-        throw new StageException("InputStream cannot be parsed or created", e);
+        file.close();
```
rseitz, line 147: add comment saying parseInputStream would close the InputStream itself
add comment saying file.close() will close FileContent and InputStream as well?

rseitz (approved):
approving but please wait for second approval from kyle before merging

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -103,14 +122,34 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+      parseInputStream(doc, inputStream);
 
+      // close file after, do not need to check if file is null as we have already checked above
       try {
-        InputStream inputStream = FileUtils.getInputStream(filePath);
-        parseInputStream(doc, inputStream);
-      } catch (IOException e) {
-        throw new StageException("InputStream cannot be parsed or created", e);
+        // closes the file, and its content, including any open stream
+        file.close();
```
kyle-kmw, line 150: If an exception is thrown. file won't close. Better to wrap the creation of FileObject file in a try-with-resources and then we don't have to call close at all and we know that try will close the resource no matter what happens e.g.

try (FileObject file = FileUtils.getFileObject(filePath, fsManager)) {
 ...
}

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -103,14 +122,34 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+      FileObject file = FileUtils.getFileObject(filePath, fsManager);
+
+      // if file null, don't process document
+      if (file == null) {
+        return null;
+      }
+
+      // getting inputStream and catching errors gracefully
+      InputStream inputStream;
+      try {
```
kyle-kmw, line 136: recommend popping the inputstream into the try

try (InputStream inputStream = file.getContent().getInputStream()) {
 ...
}

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
kyle-kmw: I suggested above creating the inputStream in a try-with-resources. The close() contract is a no-op if it is called multiple times. If some day parseInputStream is refactored and the implementation changes so that it doesn't close the inputStream (which would be a reasonable change IMO) then we at least know that the resource we created here is closed before we leave.

kyle-kmw (approved):

PR #159, opened 2024-08-02
LC-530: remove use of parallel streams and unordered iteration from VFSConnector

Jozurf:


lucille-core/src/main/java/com/kmwllc/lucille/connector/VFSConnector.java
```diff
@@ -64,8 +64,8 @@ public void execute(Publisher publisher) throws ConnectorException {
 
       fsManager.init();
 
-      // locate all valid files within the provided VFS path
-      traverseFiles(fsManager, vfsPath).forEach(fo -> {
+      // locate all valid files within the provided VFS path in order
+      traverseFiles(fsManager, vfsPath).forEachOrdered(fo -> {
```
kwatters, line 68: if you have a very large directory listing, will asking for it to be "ordered" here have a performance impact?

Jozurf: I believe this specific change calling forEachOrdered would not have much of a performance impact as we have changed from using a parallel to a sequential stream. The change made from parallelStream to stream, however, would have a negative performance impact. https://www.geeksforgeeks.org/foreach-loop-vs-stream-foreach-vs-parallel-stream-foreach/
https://stackoverflow.com/questions/32797579/foreach-vs-foreachordered-in-java-8-stream

Jozurf: From this extensive article, it states that there are no guarantees that executing a parallel stream vs a sequential stream and lists factors that affect performance. https://blogs.oracle.com/javamagazine/post/java-parallel-streams-performance-benchmark

rseitz: The stream we changed from parallel to sequential is a tiny stream of the exclude patterns, not the possibly long stream of all the files, which was already sequential. Since the stream of files was already sequential, I doubt that using forEachOrdered would have much of a performance effect. Note that this is NOT sorting the stream, it's just saying to respect the existing stream order when iterating over stream elements.

rseitz: Also note that unfortunately, this PR doesn't make VFS connector fully deterministic and we really can't do that without sorting, because the Apache VFS FileObject's getChildren returns an array without any ordering guarantee. What we're doing in this PR is just making sure that our code doesn't add any further nondeterminism on top of what the VFS API already introduces.

rseitz (approved):

PR #161, opened 2024-08-05
LC-522: add render flag for runner

Jozurf:
- need to create runner tests, but manual testing works so far. Tried normal testing and when giving non existent config path

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
@@ -13,7 +13,6 @@ import com.kmwllc.lucille.util.LogUtils;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
```diff
@@ -787,4 +787,10 @@ public void testMetrics() throws Exception {
     // 1) other metrics collected by Stage beyond the counts in the processDocumentTime timer
     // 2) metrics collected by Indexer, Worker, Publisher
   }
+
+  @Test
+  public void testRenderFlag() throws Exception {
```
rseitz, line 792: suggest adding an additional test for the case where the config has a substitution that can't be resolved
check that Runner.renderConfig throws ConfigException (which is a kind of RuntimeException)
this new test would use a different config that references an environment variable named something like LUCILLE_TEST_CONFIG_RENDER_UNDEFINED

rseitz (approved):

PR #162, opened 2024-08-09
LC-535: added ability to ignore "id" field in SolrIndexer

Jozurf:
- now allows user to place "id" field in ignoreFields config.
- ignoreFields will now apply to Children and Parent Document(s)
- Changed SolrInputRequest to use SolrId instead of Document.ID_FIELD

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
@@ -296,7 +296,9 @@ public class SolrIndexer extends Indexer {
       return;
     }
     for (Document child : children) {
-      Map<String, Object> map = child.asMap();
+      // remove key:value pair mappings if they appear in ignoreFields
+      Map<String, Object> map = getIndexerDoc(child);
+
       SolrInputDocument solrChild = new SolrInputDocument();
       for (String key : map.keySet()) {
         // we don't support children that contain nested children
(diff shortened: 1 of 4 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -135,6 +139,11 @@ protected void sendToIndex(List<Document> documents) throws Exception {
       SolrDocRequests solrDocRequests = solrDocRequestsByCollection.get(collection);
       String solrId = idOverride != null ? idOverride : doc.getId();
 
+      // if user provided uniqueKey is in document use that as id, else continue with the original id
+      if (uniqueKey != null) {
```
rseitz, line 143: consider 

if (uniqueKey != null && doc.has(uniqueKey)) {
  solrId = doc.getString(uniqueKey);
}

rseitz:
We already support an idOverrideField which allows the user to specify an alternate field like "myId" to use as the value of the "id" to be sent to Solr.
But what if the user has specified "myId" as the uniqueKey in their Solr schema and they don't want an "id" field to be included on documents in the index.
In this case we just want to allow them to specify "id" in the list of ignoreFields and make sure SolrIndexer removes "id" before sending the SolrDocument to solr.
I don't think there's a need for the user to have to specify the "uniqueKey" explicitly in the indexer config. In this use case, they would specify idOverrideField=myId and ignoreFields=["id"]. SolrIndexer doesn't need to know about the Solr schema and what uniqueKey is used there, it just needs to send documents with "myId" populated, which can be achieved via pipeline stages that set that field appropriately, and it needs to know not to send "id".

rseitz (approved):
approving but please get @kyle-kmw 's review before merging

PR #164, opened 2024-08-19
LC-536: added JDBCUtils for JDBC type handling

Jozurf:


lucille-core/src/main/java/com/kmwllc/lucille/util/JDBCUtils.java
```diff
@@ -0,0 +1,131 @@
...
+
+  public static void parseResultToDoc(Document doc, ResultSet rs, String fieldName, int columnIndex)
+      throws SQLException {
+    ResultSetMetaData metadata = rs.getMetaData();
+    typeHandling(doc, rs, fieldName, columnIndex, metadata.getColumnType(columnIndex));
+  }
+
+  private static void typeHandling(Document doc, ResultSet rs, String fieldName, int columnIndex, int columnType)
+      throws SQLException {
+      switch (columnType) {
```
spencersolomon6, line 42: Great use of the switch statement and the type mappings from SQL -> Java all look good to me. Nice work!

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/connector/jdbc/DatabaseConnector.java
```diff
@@ -185,13 +186,13 @@ public void execute(Publisher publisher) throws ConnectorException {
...
-            } catch (IllegalArgumentException e) {
-              log.warn("Error encountered while adding database object to Lucille document", e);
-            }
+          try {
+            // parse result into document
+            // can throw SQL Exception if column cannot be found or resultSet is closed
+            // will not add to document if fieldValue is null or if field value is unsupported type
+            JDBCUtils.parseResultToDoc(doc, rs, fieldName, i);
+          } catch(SQLException e) {
+            log.warn("Error encountered while processing resultSet");
```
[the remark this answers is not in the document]
Jozurf: done

lucille-core/src/main/java/com/kmwllc/lucille/util/JDBCUtils.java
```diff
@@ -0,0 +1,131 @@
...
+        case Types.BINARY:
+        case Types.VARBINARY:
+        case Types.LONGVARBINARY:
+          byte[] bytes = rs.getBytes(columnIndex);
+          if (bytes != null) {
+            doc.setOrAdd(fieldName, bytes);
+          }
+          break;
+        default:
+          log.warn("SQL Type {} not supported", columnType);
```
kiratraynor, line 128: I think it might be nice to include the actual sql Type name rather than just the number in this log. I think a way to do this would be something like:
`log.warn("SQL Type {} ({}) not supported", JDBCType.valueOf(columnType).getName(), columnType);`
Not super necessary though because we can look up the type easily from the number.
Other than this, everything looks good to me!

kiratraynor (approved):

lucille-core/src/main/java/com/kmwllc/lucille/util/JDBCUtils.java
Jozurf: done!

PR #168, opened 2024-08-20
LC-538 - Update connector timeout error message to better describe where the timeout is coming from

kiratraynor:


lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
```diff
@@ -214,7 +214,7 @@ public PublisherResult waitForCompletion(ConnectorThread thread, int timeout) th
 
       if (ChronoUnit.MILLIS.between(start, Instant.now()) > timeout) {
         log.error("Exiting run with " + numPending() + " pending documents; publisher timed out (" + timeout + "ms)");
```
rseitz, line 216: still says "publisher timed out" here

rseitz (approved):

PR #170, opened 2024-08-22
LC-542: increase default queue capacity

Jozurf:
- increased to 10000
- added some comments

### DIFF
--- a/lucille-examples/lucille-vector-ingest-example/conf/10M-parquet-pinecone.conf
+++ b/lucille-examples/lucille-vector-ingest-example/conf/10M-parquet-pinecone.conf
@@ -64,8 +64,10 @@ publisher {
   #
   # attempts to publish beyond this capacity will cause the publisher to block
   # this setting applies only when running Lucille in local / single-JVM mode using a LocalMessageManager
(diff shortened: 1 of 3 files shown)


lucille-examples/lucille-vector-ingest-example/conf/10M-parquet-pinecone.conf
```diff
@@ -64,8 +64,10 @@ publisher {
   #
   # attempts to publish beyond this capacity will cause the publisher to block
   # this setting applies only when running Lucille in local / single-JVM mode using a LocalMessageManager
-  # this setting defaults to 100
-  queueCapacity: 100
+  # this setting affects Lucille's memory footprint as it determines how many in-flight documents can be held in memory at once
```
rseitz, line 67: this setting only takes effect when running in local mode
the number of inflight documents can be roughly double the value of this setting, because two queues are used in local mode and this setting is applied separately to each queue. There is a queue of published documents waiting to be run through the pipeline, and a queue of processed documents waiting to be indexed.

rseitz (approved):

PR #171, opened 2024-08-22
LC-465: create OpenAIEmbed Stage

Jozurf:
- currently retrieving API key from config file
- truncating if token limit is hit


lucille-core/src/main/java/com/kmwllc/lucille/stage/OpenAiEmbedding.java
```diff
@@ -0,0 +1,120 @@
...
+  private void sendForEmbedding(Document doc) throws StageException {
+    if (!doc.has(textField)) {
+      log.warn("{} does not exist, skipping document: {}", textField, doc);
+      return;
+    }
+
+    String content = doc.getString(textField);
+    content = ensureContentIsWithinLimit(content);
+    Response<Embedding> response = model.embed(content);
+    Embedding embedding = response.content();
```
Jozurf, line 104: Need to check response status to make sure it was successful before adding.

lucille-core/src/main/java/com/kmwllc/lucille/stage/OpenAiEmbedding.java
```diff
@@ -0,0 +1,135 @@
...
+    String content = doc.getString(textField);
+    content = ensureContentIsWithinLimit(content);
+    Response<Embedding> response;
+
+    // model.embed will retry 3 times by default
+    // model.embed might throw runtimeException due to invalid API key and other errors.
+    try {
+      response = model.embed(content);
+    } catch (RuntimeException e) {
+      log.warn("failed to get embedding", e);
```
spencersolomon6, line 115: First, this log message should be improved to be more informative. maybe include the id of the erroring document?

Second, I think we should throw a StageException here instead of just logging. This error would mean that none of our encodings are being properly created, so we would likely not want to just skip this step and continue the run if this is not working.

lucille-core/src/main/java/com/kmwllc/lucille/stage/OpenAiEmbedding.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  private String ensureContentIsWithinLimit(String content) {
+    int tokenUsage = enc.countTokens(content);
+    String truncatedContent = null;
+    if (tokenUsage > DEFAULT_OPENAI_TOKEN_LIMIT) {
+      EncodingResult encoded = enc.encode(content, DEFAULT_OPENAI_TOKEN_LIMIT);
+      truncatedContent = enc.decode(encoded.getTokens());
+    }
+    return truncatedContent == null ? content : truncatedContent;
```
spencersolomon6, line 133: This can be simplified by removing the truncatedContent variable. In the if statement, just `return enc.decode(encoded.getTokens())` and outside of the if just `return content`

spencersolomon6 (changes requested):

lucille-core/src/main/java/com/kmwllc/lucille/stage/OpenAiEmbedding.java
```diff
@@ -0,0 +1,135 @@
...
+    Response<Embedding> response;
+
+    // model.embed will retry 3 times by default
+    // model.embed might throw runtimeException due to invalid API key and other errors.
+    try {
+      response = model.embed(content);
+    } catch (RuntimeException e) {
+      log.warn("failed to get embedding", e);
+      return;
+    }
```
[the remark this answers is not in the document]
spencersolomon6: Response is not AutoCloseable so this cannot be put in try-resources. Resolving comment.

lucille-core/src/main/java/com/kmwllc/lucille/stage/OpenAIEmbedding.java
```diff
@@ -0,0 +1,194 @@
...
+  private final Integer dimensions;
+  private Encoding enc;
+
+  private static final Logger log = LoggerFactory.getLogger(OpenAIEmbedding.class);
+
+  public OpenAIEmbedding(Config config) throws StageException {
+    super(config, new StageSpec()
+        .withRequiredProperties("text_field", "embed_document", "embed_children", "api_key")
+        .withOptionalProperties("embedding_field", "model_name", "dimensions"));
+    textField = config.getString("text_field");
```
rseitz, line 63: convention is to use "this." prefix when setting instance variables inside a constructor

Jozurf: resolved in commit 0c8bc21

spencersolomon6 (approved):

rseitz (approved):

lucille-core/src/main/java/com/kmwllc/lucille/core/OpenAIEmbeddingModel.java
```diff
@@ -0,0 +1,37 @@
+package com.kmwllc.lucille.core;
+
+
+import com.typesafe.config.Config;
+
+public enum OpenAIEmbeddingModel {
```
rseitz, line 6: should go in com.kmwllc.lucille.stage.util package

lucille-core/src/main/java/com/kmwllc/lucille/stage/OpenAIEmbedding.java
```diff
@@ -0,0 +1,195 @@
...
+ * - model_name (String, Optional) : the name of the OpenAI embedding model to use, set default to text-embedding-3-small
+ *    1. text-embedding-3-small
+ *    2. text-embedding-3-large
+ *    3. text-embedding-ada-002
+ *    more details: <a href="https://platform.openai.com/docs/guides/embeddings/embedding-models">...</a>
+ * - dimensions (Integer, Optional) : number of dimensions the resulting embedding should have. Only supported in text-embedding-3
+ * and later models. Default set to null, which will call the model's default dimensions.
+ */
+
+public class OpenAIEmbedding extends Stage {
```
rseitz, line 43: most stage names are based on verbs
so I'd suggest OpenAIEmbed or EmbedForOpenAI

Jozurf: changed to OpenAIEmbed in [feeaa04](https://github.com/kmwtechnology/lucille/pull/171/commits/feeaa045fb3aa12e3a184b1bfa7564aafdd21d1d)

lucille-core/src/main/java/com/kmwllc/lucille/core/OpenAIEmbeddingModel.java
Jozurf: resolved in [feeaa04](https://github.com/kmwtechnology/lucille/pull/171/commits/feeaa045fb3aa12e3a184b1bfa7564aafdd21d1d)

PR #177, opened 2024-09-16
LC-506: support deletion for PineconeIndexer

Jozurf:
- added support for deletion using Indexer config deletionMarkerField & deletionMarkerFieldValue
- namespace will now be "default" if no namespaces defined (deletion by prefix needs non empty namespace)
- added additional checks on other operations to increase robustness
- upgraded netty & grpc to latest version to avoid error in internal grpc calls
- added javadocs
- added local testing

### DIFF
--- a/lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
+++ b/lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
@@ -97,8 +97,15 @@ public class PineconeIndexer extends Indexer {
     List<Document> uploadList = new ArrayList<>();
     for (Document doc : documents) {
       if (isDeletion(doc)) {
-        deleteList.add(doc);
+        // if doc is already in uploadList and is now marked for deletion, then remove that document in upload
+        if (uploadList.contains(doc)) {
+          uploadList.remove(doc);
+        } else {
+          deleteList.add(doc);
+        }
       } else {
+        // if doc is already in deleteList, then continue adding to uploadList as uploading is performed after deletion
+        // note: updating a non-existent ID will not throw an error in Pinecone
         uploadList.add(doc);
       }
     }
--- a/lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
+++ b/lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
@@ -97,15 +97,18 @@ public class PineconeIndexer extends Indexer {
     List<Document> uploadList = new ArrayList<>();
     for (Document doc : documents) {
       if (isDeletion(doc)) {
-        // if doc is already in uploadList and is now marked for deletion, then remove that document in upload
-        if (uploadList.contains(doc)) {
-          uploadList.remove(doc);
+        // if doc is in uploadList and is now marked for deletion, then remove that document in uploadList and add it to deleteList
+        // if deleteByPrefix is true, also delete all instances of those in uploadList
+        if (deleteByPrefix) {
+          uploadList = uploadList.stream().filter(other -> !other.getId().startsWith(doc.getId())).collect(
+              Collectors.toList());
         } else {
-          deleteList.add(doc);
+          uploadList = uploadList.stream().filter(other -> !other.getId().equals(doc.getId())).collect(Collectors.toList());
         }
+        deleteList.add(doc);
       } else {
         // if doc is already in deleteList, then continue adding to uploadList as uploading is performed after deletion
-        // note: updating a non-existent ID will not throw an error in Pinecone
+        // note: updating/deleting a non-existent ID will return an OK response
         uploadList.add(doc);
       }
     }
--- a/lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
+++ b/lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
@@ -1,7 +1,6 @@
 package com.kmwllc.lucille.pinecone.indexer;
 
 import com.kmwllc.lucille.core.IndexerException;
-import io.pinecone.clients.Pinecone.Builder;
 import io.pinecone.proto.ListItem;
 import io.pinecone.proto.ListResponse;
 import io.pinecone.proto.UpsertResponse;
@@ -10,10 +9,8 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.apache.commons.lang3.StringUtils;
 import org.openapitools.control.client.model.IndexModelStatus.StateEnum;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -37,8 +34,8 @@ import static io.pinecone.commons.IndexInterface.buildUpsertVectorWithUnsignedIn
  * - mode (String, Optional) : the type of request you want PineconeIndexer to send to your index
  *    1. upsert (will replace if vector id exists in namespace or index if no namespace is given)
  *    2. update (will only update embeddings)
- *    3. delete (deletes all vectors given document ids)
- *      - will only delete the exact id of documents that go through the pipeline, for id prefix deletion, look at deletionPrefix
+ *    for deletion requests, take a look at application-example.conf under Indexer configs. Only support
+ *    deletionMarkerField, and deletionMarkerFieldValue as serverless index currently does not support delete via query/metadata.
  * - defaultEmbeddingField (String, required if namespaces is not set) : the field to retrieve embeddings from
  * - deleteByPrefix (boolean, optional) : sends a deletion by prefix request instead of a usual request
  * - addDeletionPrefix (String, optional) : String that adds on to all document ids used for prefix deletion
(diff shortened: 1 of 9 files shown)


lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -78,7 +101,8 @@ private void uploadDocuments(List<Document> documents, String embeddingField, St
 
     if (mode.equalsIgnoreCase("upsert")) {
       List<VectorWithUnsignedIndices> upsertVectors = documents.stream()
-          .filter(doc -> doc.hasNonNull(embeddingField)) // buildUpsertVector would throw an error if embeddings is null
+          // buildUpsertVector would throw an error if embeddings is null,
+          // should add dropDocument stage with stage conditions in pipeline before indexing
```
Jozurf, line 105: Previously, we had a filter here to filter out documents with no embeddings, but now opted to remove filter to standardize the use case of Indexer, and instead extract the filtering logic to a dropDocument stage with stage conditions.

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -90,24 +114,29 @@ private void uploadDocuments(List<Document> documents, String embeddingField, St
...
-    // larger dimensions will mean smaller batch size limit
-    if (upsertVectors.size() > 1000) {
-      throw new IndexerException("max upsert size of each batch is 1000, reduce the batch size indexer configuration.");
+  private void deleteDocuments(List<Document> documents, String namespace, String additionalPrefix) throws IndexerException {
+    List<String> idsToDelete = documents.stream()
+        .map(doc -> doc.getId() + additionalPrefix)
+        .collect(Collectors.toList());
+
+    try {
+      index.deleteByIds(idsToDelete, namespace);
```
Jozurf, line 131: In the serverless Pinecone, there are 3 ways to delete documents-- id, id prefix, namespaces. Deleting namespaces is similar to Solr deleting a collection, so not supporting that. Deletion by id & id prefix have the same syntax. In the case where user wants to delete record id "100", this will not delete records of id "1004", "10005", etc.

Jozurf: outdated

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -22,6 +28,22 @@
...
+ * - apiKey (String) : API key used for requests
+ * - mode (String, Optional) : the type of request you want PineconeIndexer to send to your index
+ *    1. upsert (will replace if vector id exists in namespace or index if no namespace is given)
+ *    2. update (will only update embeddings)
+ *    3. delete (deletes all vectors given document ids)
+ *      - will only delete the exact id of documents that go through the pipeline, for id prefix deletion, look at deletionPrefix
+ * - defaultEmbeddingField (String, required if namespaces is not set) : the field to retrieve embeddings from
+ * - deleteByPrefix (boolean, optional) : sends a deletion by prefix request instead of a usual request
+ * - addDeletionPrefix (String, optional) : String that adds on to all document ids used for prefix deletion
+ *  - e.g. if id of doc is "doc1" & deletionPrefix is "-", then all vectors containing "doc1-" in id as prefix will be deleted.
```
Jozurf, line 45: this is useful if there are multiple versions of documents in a namespace. e.g.  doc1-v1-chunk1 vs doc1-v2-chunk1. To delete all version 1 of doc1, here the user would put "-v1" in the addDeletionPrefix

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -33,24 +55,33 @@ public class PineconeIndexer extends Indexer {
...
+    this.metadataFields = config.hasPath("pinecone.metadataFields")
+        ? new HashSet<>(config.getStringList("pinecone.metadataFields")) : new HashSet<>();
     this.mode = config.hasPath("pinecone.mode") ? config.getString("pinecone.mode") : "upsert";
     this.defaultEmbeddingField = ConfigUtils.getOrDefault(config, "pinecone.defaultEmbeddingField", null);
     this.client = new Pinecone.Builder(config.getString("pinecone.apiKey")).build();
     this.index = this.client.getIndexConnection(this.indexName);
-    if (namespaces == null && defaultEmbeddingField == null) {
+    this.addDeletionPrefix = config.hasPath("pinecone.addDeletionPrefix") ? config.getString("pinecone.addDeletionPrefix") : "";
+    this.deleteByPrefix = config.hasPath("pinecone.deleteByPrefix") ? config.getBoolean("pinecone.deleteByPrefix") : false;
+    if (namespaces == null && defaultEmbeddingField == null && !mode.equalsIgnoreCase("delete")) {
```
Jozurf, line 73: in "delete" mode, it is not necessary to have namespaces or defaultEmbeddingField. This is because we allow upserting and updating when namespaces is null and defaultEmbeddingField is set. In this case, vectors will be uploaded to "default" namespace. "delete" mode will then delete from "default" if namespaces is null

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -70,15 +101,16 @@ protected void sendToIndex(List<Document> documents) throws IndexerException {
         uploadDocuments(documents, (String) entry.getValue(), entry.getKey());
       }
     } else {
-      uploadDocuments(documents, defaultEmbeddingField, "");
+      uploadDocuments(documents, defaultEmbeddingField, "default");
```
Jozurf, line 104: adding "default" as the default namespaces to allow delete by prefix to work. Index.list requires a non empty string for namespace to list vectors by prefix -- an operation required and recommended by pinecone to delete by prefix.
More info here: https://docs.pinecone.io/guides/data/manage-rag-documents#delete-all-records-for-a-parent-document

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -90,24 +114,46 @@ private void uploadDocuments(List<Document> documents, String embeddingField, St
...
           .collect(Collectors.toList());
-
       upsertDocuments(upsertVectors, namespace);
     }
 
     if (mode.equalsIgnoreCase("update")) {
       updateDocuments(documents, embeddingField, namespace);
     }
+
+    if (mode.equalsIgnoreCase("delete")) {
```
kyle-kmw: I am trying to consider whether we want to have "delete" as a mode. Or rather something that is indicated by a document e.g. a delete document. That would be closer to the SolrIndexer delete feature. It seems like by using a delete mode we wouldn't be able to set up a system that handles updates and deletes within a single pipeline.

kyle-kmw: Let's remove the "delete" mode and replace it with support for a deletionMarkerField and deletionMarkerFieldValue. Similar to the SolrIndexer.

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -33,24 +55,33 @@ public class PineconeIndexer extends Indexer {
...
+  private final String addDeletionPrefix;
+  private final boolean deleteByPrefix;
 
   public PineconeIndexer(Config config, IndexerMessenger messenger, String metricsPrefix) {
     super(config, messenger, metricsPrefix);
     this.indexName = config.getString("pinecone.index");
     this.namespaces = config.hasPath("pinecone.namespaces") ? config.getConfig("pinecone.namespaces").root().unwrapped() : null;
-    this.metadataFields = new HashSet<>(config.getStringList("pinecone.metadataFields"));
+    this.metadataFields = config.hasPath("pinecone.metadataFields")
+        ? new HashSet<>(config.getStringList("pinecone.metadataFields")) : new HashSet<>();
```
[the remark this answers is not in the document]
Jozurf: No more "delete" mode but still applies if user wants indexer to perform deletion via Indexer configs.

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
Jozurf: no more "delete" mode. Instead moved validation part to when there is a need to upload documents

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -65,20 +92,61 @@ public boolean validateConnection() {
...
-    if (namespaces != null) {
-      for (Map.Entry<String, Object> entry : namespaces.entrySet()) {
-        uploadDocuments(documents, (String) entry.getValue(), entry.getKey());
+    // retrieve documents to delete & upload
+    List<Document> deleteList = new ArrayList<>();
+    List<Document> uploadList = new ArrayList<>();
+    for (Document doc : documents) {
+      if (isDeletion(doc)) {
+        // if doc is already in uploadList and is now marked for deletion, then remove that document in upload
+        if (uploadList.contains(doc)) {
```
kyle-kmw, line 101: The contains(doc), remove(doc) won't work as expected here due to relying on Object.equals.

kyle-kmw: You can use something like uploadList.filter(other -> other.id.equals(doc.id));

kyle-kmw (approved):

PR #178, opened 2024-09-16
LC-543: Refactor Document Interface to Support Timestamp & Date

Jozurf:
- also added checking instance of Float type in setField, setOrAdd, addToField object methods


lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
```diff
@@ -477,6 +490,69 @@ public JsonNode getJson(String name) {
...
 
+  @Override
+  public Date getDate(String name) {
+    if (!data.has(name)) {
+      return null;
+    }
+
+    JsonNode node = getSingleNode(name);
+
+    long dateLong = node.asLong();
```
spencersolomon6, line 501: Will this return an error if node is null? If so, we can just remove the dateLong variable and move node.asLong() directly into the Date constructor.

I don't think node should ever be null, since we first check if it exists on `data` but since we're checking node.isNull() in the return statement I guess it can be null?

spencersolomon6 (changes requested):

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
Jozurf: changed implementation to convert Date/Timestamp to Instant, and then delegating to the setters/getters of instant type methods

spencersolomon6 (approved):

rseitz (approved):

PR #181, opened 2024-09-18
LC-549: reduce watcher interval

Jozurf:


lucille-core/src/main/java/com/kmwllc/lucille/core/WorkerPool.java
```diff
@@ -28,7 +28,7 @@ public class WorkerPool {
   public static final int DEFAULT_POOL_SIZE = 1;
 
   private static final Logger log = LoggerFactory.getLogger(WorkerPool.class);
-  private static final long WATCHER_PERIOD_SECONDS = 5;
+  private static final long WATCHER_PERIOD_MILLISECONDS = 500;
```
Jozurf, line 31: tested by ingesting Enron dataset into Solr. No noticeable slowdown from lowering watcher period

rseitz (approved):

PR #182, opened 2024-09-20
LC-161: add connection retry for dbConnector

Jozurf:


rseitz (changes requested):
requesting to make retry pause a milliseconds value
in the test set it to something like 10 milliseconds so we don't add significantly to the duration of the test suite

rseitz (approved):

PR #187, opened 2024-09-26
LC-550: create EmitDocsToDeleteByPrefix Pinecone stage

Jozurf:
- Created PineconeUtils class to reuse code
- Created EmitDocsToDeletebyPrefix class. Only process documents that have deleted marker field.
- Refactored PineconeIndexer to remove previous prefix deletion logic
- Updated tests for PineconeIndexer
- Added tests for PineconeUtils, EmitDocsToDeletebyPrefix


lucille-plugins/lucille-pinecone/pom.xml
```diff
@@ -24,6 +24,12 @@
       <scope>provided</scope>
     </dependency>
 
+    <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>2.17.0</version>
+    </dependency>
```
Jozurf, line 31: For some reason, I am unable to create documents by calling Document.create('id') without it throwing error that it is unable to find class ObjectNode. So for temporary measures adding the dependency here.

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/stage/EmitDeletedByPrefix.java
```diff
@@ -0,0 +1,130 @@
...
+  private final String deletionMarkerFieldValue;
+
+  private static final Logger log = LoggerFactory.getLogger(EmitDeletedByPrefix.class);
+
+  public EmitDeletedByPrefix(Config config) {
+    super(config, new StageSpec()
+        .withOptionalProperties("dropOriginal", "addPrefix")
+        .withOptionalParents("namespaces")
+        .withRequiredProperties("apiKey", "deletionMarkerField", "deletionMarkerFieldValue", "index"));
+    this.pineconeManager = PineconeManager.getInstance(config.getString("apiKey"), config.getString("index"));
```
Jozurf, line 47: How to "share" config values? It would be nice for stage to retrieve configs from other parts of the configs.

kwatters: config files can reference environment variables or system properties that are passed in.. so you the same config value can be used in multiple places...

rseitz: Along with referencing an env var or system property, you can define a variable in the conf file itself and then reference it in multiple places using substitutions. The syntax is explained here: https://github.com/lightbend/config?tab=readme-ov-file#uses-of-substitutions

So, if you set myVal=1 and you want to pass myVal into multiple stages, you would include myProp=${myVal} in the config for each stage

lucille-plugins/lucille-pinecone/pom.xml
Jozurf: found that this was due to my testing environment not containing the class.

Jozurf: Still unable to remove dependency. Github build fails with same error when dependency not included. So far only passes when included or when scope for lucille-core dependency is set to compile.

`<dependency>
      <groupId>com.kmwllc</groupId>
      <artifactId>lucille-core</artifactId>
      <version>${project.version}</version>
      <scope>compile</scope>
    </dependency>
`

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/stage/EmitDeletedByPrefix.java
```diff
@@ -0,0 +1,146 @@
...
+  private final String deletionMarkerFieldValue;
+
+  private static final Logger log = LoggerFactory.getLogger(EmitDeletedByPrefix.class);
+
+  public EmitDeletedByPrefix(Config config) {
+    super(config, new StageSpec()
+        .withOptionalProperties("dropOriginal", "addPrefix")
+        .withOptionalParents("namespaces")
+        .withRequiredProperties("apiKey", "deletionMarkerField", "deletionMarkerFieldValue", "index"));
+    this.client = PineconeManager.getClient(config.getString("apiKey"));
```
kyle-marcotte-tfs, line 58: The creation of the client should be moved to the start method. Same for anything that relies on calling externally using the client.

In the lifecycle of a stage. The constructor should never fail. The start method should fail if there are issues setting up anything that it depends on e.g. client / index connection and if theres an issue in startup that would prevent it from working throw a stage exception.

lucille-plugins/lucille-pinecone/pom.xml
kyle-marcotte-tfs: This is fine to add the dependency. Do not specify the version here because the project as a whole is already importing the jackson bom in lucille-parent.

So it should be

```
    <dependency>
        <groupId>com.fasterxml.jackson.core</groupId>
        <artifactId>jackson-databind</artifactId>
    </dependency>```

And it will always pull in the same version that lucille is using.

Comment (Jozurf) [reply lucille-plugins/lucille-pinecone/pom.xml:30]: resolved in [807c2fb](https://github.com/kmwtechnology/lucille/pull/187/commits/807c2fb3b69ce3736420f2909902e90958310386)

File: lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/util/PineconeManager.java [hunk 4]
```diff
@@ -0,0 +1,26 @@
...
+
+import io.pinecone.clients.Pinecone;
+import java.lang.invoke.MethodHandles;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PineconeManager {
+  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
rseitz, line 11: seems to be unused

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -95,52 +97,60 @@ protected void sendToIndex(List<Document> documents) throws IndexerException {
...
-        // if deleteByPrefix is true, also delete all instances of those in uploadList
-        if (deleteByPrefix) {
-          uploadList = uploadList.stream().filter(other -> !other.getId().startsWith(doc.getId())).collect(
-              Collectors.toList());
-        } else {
-          uploadList = uploadList.stream().filter(other -> !other.getId().equals(doc.getId())).collect(Collectors.toList());
-        }
+        // always add to deleteList;
+        // in the case where doc is in uploadList and is now marked for deletion, then remove that document in uploadList
+        uploadList.removeIf(x -> x.getId().equals(doc.getId()));
```
spencersolomon6, line 102: How would a document marked for deletion ever end up in the upload list? Don't we only add documents to the uploadList if isDeletion() is false?

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -95,52 +97,60 @@ protected void sendToIndex(List<Document> documents) throws IndexerException {
...
         deleteList.add(doc);
       } else {
-        // if doc is already in deleteList, then continue adding to uploadList as uploading is performed after deletion
-        // note: updating/deleting a non-existent ID will return an OK response
+        // always add non deletion documents to uploadList;
+        // in the case that doc is also in deleteList, we remove it from the deleteList to avoid sending unnecessary deletion request
+        deleteList.removeIf(x -> x.getId().equals(doc.getId()));
```
spencersolomon6, line 107: Same question as above re deleteList

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -95,52 +97,60 @@ protected void sendToIndex(List<Document> documents) throws IndexerException {
...
           "At least one of a defaultEmbeddingField or a non-empty namespaces mapping is required when uploading documents.");
     }
+    if (uploadListSize > MAX_PINECONE_BATCH_SIZE) {
+      throw new IndexerException(
+          "Pinecone maximum batch size is " + MAX_PINECONE_BATCH_SIZE + ", lower batchSize in indexer configs.");
+    }
+  }
+
+  private void validateDeleteRequirements(int deleteListSize) throws IndexerException {
+    if (deleteListSize > MAX_PINECONE_BATCH_SIZE) {
```
spencersolomon6, line 150: This code can be abstracted to a new method validateBatchSizeRequirements() or similar. This could then be called in validateUploadRequirements and be called directly on deleteList.size() to avoid copied code.

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/stage/EmitDocsToDeleteByPrefix.java
```diff
@@ -0,0 +1,149 @@
...
+      index = client.getIndexConnection(indexName);
+    } catch (Exception e) {
+      throw new StageException("Unable to get client or index connection", e);
+    }
+
+    StateEnum state = client.describeIndex(indexName).getStatus().getState();
+    boolean isStable = state != StateEnum.INITIALIZING &&
+        state != StateEnum.INITIALIZATIONFAILED &&
+        state != StateEnum.TERMINATING;
+    if (!isStable) {
```
spencersolomon6, line 77: Since this is the same check as is performed in PineconeIndexer, can we abstract these out to a method on the PineconeManager boolean isStable() (or similar) which returns the status of this. That way if Pinecone adds some new state down the line we only have to fix in one place.

lucille-plugins/lucille-pinecone/src/test/java/com/kmwllc/lucille/pinecone/util/PineconeManagerTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Test
+  public void testReuseClient() throws Exception {
+    Pinecone mockClient = mock(Pinecone.class);
+    Pinecone differentClient = mock(Pinecone.class);
+    try(MockedConstruction<Pinecone.Builder> builder = Mockito.mockConstruction(Pinecone.Builder.class,(mock,context)-> {
+      // will return mockClient first time is called, then return a different client when called a second time
+      when(mock.build()).thenReturn(mockClient).thenReturn(differentClient);
+    })) {
+      // test that "sameClient" is retrieving the same client object, and that .build was only called trice
```
spencersolomon6, line 21: Should be "thrice" in the above comment but maybe we should change it to "three times". Only suggesting because I misread it as twice at first and was surprised when the test asserted it was called 3 times.

spencersolomon6 (changes requested):

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/stage/EmitDocsToDeleteByPrefix.java
```diff
@@ -0,0 +1,149 @@
...
+    if (namespaces != null) {
+      for (Map.Entry<String, Object> entry : namespaces.entrySet()) {
+        idsToDelete.addAll(getVectorsByPrefix(doc, addPrefix, entry.getKey()));
+      }
+    } else {
+      idsToDelete.addAll(getVectorsByPrefix(doc, addPrefix, PineconeManager.getDefaultNamespace()));
+    }
+
+    // drop original document if set to true
+    doc.setDropped(dropOriginal);
```
rseitz, line 100: very minor comment but I'd suggest slightly more verbose

`if (dropOriginal) { doc.setDropped(true);}`

because it's a bit more specific and explicit about the intent. If there were other code somewhere above that marked the doc as dropped (I know there isn't, but there could be) then doc.setDropped(dropOriginal) would actually un-drop it if dropOriginal is false.

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
```diff
@@ -32,49 +31,45 @@
...
-      throw new IllegalArgumentException("namespaces mapping must be non-empty if provided");
+      throw new IllegalArgumentException("Namespaces mapping must be non-empty if provided.");
     }
-    // max upsertSize is 2MB or 1000 records, whichever is reached first, so stopping lucille run if batch size set to more than 1000
+    // max upload batch is 2MB or 1000 records, whichever is reached first, so stopping lucille run if batch size set to more than 1000
     // larger dimensions will mean smaller batch size limit, letting API throw the error if encountered.
-    if (mode.equalsIgnoreCase("upsert") && DEFAULT_BATCH_SIZE > 1000) {
-      throw new IllegalArgumentException("maximum batch size for upsert is 1000, and lower if vectors have higher dimensions");
+    if (DEFAULT_BATCH_SIZE > MAX_PINECONE_BATCH_SIZE) {
+      log.warn(
```
rseitz, line 70: as discussed, this should check what the batch size is actually set to (which could happen via indexer.batchSize or via the default) and log the warning only if it's too high

Jozurf: Ticket [created](https://kmwllc.atlassian.net/browse/LC-554?atlOrigin=eyJpIjoiOWYxYzM1ZjZkNGZlNDJlMGI5NGZlMDY2YTQ2NjVlNWMiLCJwIjoiaiJ9) to enable retrieval of indexer batch size.

spencersolomon6 (approved):

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/indexer/PineconeIndexer.java
Jozurf: replaced in [f745a9e](https://github.com/kmwtechnology/lucille/pull/187/commits/f745a9e5b69d49fdd8b91e8728ea27d4dc3406de) to check for actual batch size

lucille-plugins/lucille-pinecone/src/main/java/com/kmwllc/lucille/pinecone/util/PineconeManager.java
```diff
@@ -0,0 +1,34 @@
+package com.kmwllc.lucille.pinecone.util;
+
+import io.pinecone.clients.Pinecone;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import org.openapitools.control.client.model.IndexModelStatus.StateEnum;
+
+public class PineconeManager {
+  private static final String DEFAULT_NAMESPACE = "default";
+  private static final Map<String, Pinecone> clients = new ConcurrentHashMap<>();
```
rseitz, line 10: when might we need or want to close these clients?

one scenario is if the lucille config includes multiple connectors/pipelines. the first connector might be pinecone, the second might be something not relating to pinecone. Do we want to leave the pinecone clients open while the second connector is executing?

Jozurf: Discussed this on call. Currently, the only places where a Pinecone client is used are stages & Indexers, both have methods that are called after it is done executing (stop() for stage, closeConnection() for Indexer). So in both cases where we have created a client, we have ending methods to close the connection. The Pinecone Client class has a package private closeConnection method, and only called when the Pinecone Index class calls close(). So to make sure that we are not leaving pinecone clients open unnecessary, both Stage and Indexer will call index.close() in their closing methods.

Jozurf: After discussion, found out that pinecone clients are not used for connectors, only for stages and indexers. Brought up a good question in closing clients appropriately and edge cases like concurrent lucille runs in the same JVM. In the end, decided it is safer to not share clients and have each stage and indexer have their own instances of their connection, such that they could close the connection in their respective end methods (stop() in stages and closeConnection() in indexer).

rseitz (approved):

PR #188, opened 2024-10-01
LC-553 Support Configurable Event Topic Name

spencersolomon6:
update the getEventTopicName method to grab a value from the config, if it exists. Updated tests to match


rseitz:
Looks good! One thing is we should add this property to application-example.conf and include a comment explaining when it can be used and when it shouldn't. i.e. it can be used in a streaming use case where a Worker or WorkerIndexer is consuming directly from Kafka and the Runner is not being used to launch batch ingests.

rseitz (approved):

PR #190, opened 2024-10-02
LC-554: add getBatchCapacity to Indexer

Jozurf:


lucille-core/src/main/java/com/kmwllc/lucille/core/Batch.java
```diff
@@ -36,4 +36,10 @@ public interface Batch {
    * regardless of whether the batch is expired.
    */
   public List<Document> flush();
+
+  /**
+   * Retrieves the capacity of a batch. Returns default capacity of Indexer
```
rseitz, line 41: suggest removing the second sentence here. This method should return the actual capacity of the Batch, regardless of the process by which it was set. A Batch can be created programatically, outside the context of Indexer.

rseitz (approved):
approving with a minor suggestion to update the javadoc for Batch.getCapacity()

lucille-core/src/main/java/com/kmwllc/lucille/core/Batch.java
Jozurf: resolved in [906c0bd](https://github.com/kmwtechnology/lucille/pull/190/commits/906c0bd410445e494e8f208f561f040e922f00ba)

PR #203, opened 2024-10-21
LC-501: standardize stage properties to camelCase

Jozurf:


lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,125 @@
...
+    } catch (Exception e) {
+      throw new IOException("Error getting stage properties from config: " + fileName, e);
+    }
+
+    return stageProperties;
+  }
+
+  private static void validateArgument(String[] args) {
+    if (args.length != 1) {
+      log.error("Usage: java ConfigProcessor <filepath/To/Conf>");
```
Jozurf, line 104: need to change 'ConfigProcessor' to 'CamelCaseConfigConverter'

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  public static void main(String[] args) {
+    validateArgument(args);
+
+    try {
+      processFile(args[0]);
+    } catch (Exception e) {
+      log.error("Error processing file: {}", e.getMessage());
```
rseitz, line 121: could remove the {} from the string and pass e directly, so the full stacktrace will be printed

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+      }
+    } catch (Exception e) {
+      throw new IOException("Error getting stage properties from config: " + fileName, e);
+    }
+
+    return stageProperties;
+  }
+
+  private static void validateArgument(String[] args) {
```
rseitz, line 102: minor -- I would call this validateArguments (plural)
even though only one arg is expected, more could be passed, and we might explicitly support more at some point

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,125 @@
...
+    sb.append(result[0]).append(snakeToCamel(stageProperty)).append(result[1]);
+
+    return sb.toString();
+  }
+
+  private static void processFile(String filePath) throws Exception {
+    File file = new File(filePath);
+    String directory = file.getParent();
+    String fileName = file.getName();
+    String nameWithoutExt = fileName.substring(0, fileName.lastIndexOf('.'));
```
rseitz, line 67: look at org.apache.commons.io.FilenameUtils to see if it gives a way to do this

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,125 @@
...
+        sb.append(line).append("\n");
+      }
+    } catch (Exception e) {
+      throw new IOException("Error processing file: " + file.getName(), e);
+    }
+
+    return sb.toString();
+  }
+
+  private static String processLine(String line, String stageProperty) throws Exception {
```
rseitz, line 52: what about String.replaceFirst()
if you want to test that there isn't a second instance of the stageProperty to replace, you could check that String.indexOf(stageProperty) is -1

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,125 @@
...
+      log.info("Written new config to {}", filePath);
+    } catch (Exception e) {
+      throw new IOException("Error writing new config to file path: " + filePath, e);
+    }
+  }
+
+  private static String applyCamelCase(File file, List<String> stageProperties) throws IOException {
+    StringBuilder sb = new StringBuilder();
+
+    try (BufferedReader br = new BufferedReader(new FileReader(file))) {
```
rseitz, line 34: explore whether we can replace this code with the "Simple Code" example in this StackOverflow answer: https://stackoverflow.com/a/40836618

basically, we'd be doing StringUtils.replaceEach( text, keys, values )
using org.apache.commons.lang3.StringUtils

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
Jozurf: resolved in [f30c7ff](https://github.com/kmwtechnology/lucille/pull/203/commits/f30c7ffa4e3a1854f698e2c2f46d753cec0c4323)

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
Jozurf: resolved in [f30c7ff](https://github.com/kmwtechnology/lucille/pull/203/commits/f30c7ffa4e3a1854f698e2c2f46d753cec0c4323)

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
Jozurf: resolved in [f30c7ff](https://github.com/kmwtechnology/lucille/pull/203/commits/f30c7ffa4e3a1854f698e2c2f46d753cec0c4323)

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
Jozurf: added comment in [f90fc30](https://github.com/kmwtechnology/lucille/pull/203/commits/f90fc3031cc2f50349bddcef03f60cea5629a064)

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    String filePath = args[0];
+    if (!filePath.endsWith(".conf")) {
+      log.error("Invalid file extension, must be used on a config file.");
+      System.exit(1);
+    }
+  }
+
+  public static void main(String[] args) {
```
rseitz, line 104: To run this I had to do mvn clean install at the top level, then go to lucille-core/target and run `java -cp 'lucille.jar:lib/*' com.kmwllc.lucille.util.CamelCaseConfigConverter`
Should document that in the class-level javadoc (or a separate readme.md file)

with no args, the utility did nothing (would be better to print out a warning message)
with a file that didn't exist, the error messages weren't easy to understand -- two cases to consider, the filename has no extension, or it does. If the filename has no extension but also doesn't exist, would be better to warn that it doesn't exist...

$ java -cp 'lucille.jar:lib/*' com.kmwllc.lucille.util.CamelCaseConfigConverter

$ java -cp 'lucille.jar:lib/*' com.kmwllc.lucille.util.CamelCaseConfigConverter 1
24/11/05 12:11:04 ERROR CamelCaseConfigConverter: Invalid file extension, must be used on a config file.

$ java -cp 'lucille.jar:lib/*' com.kmwllc.lucille.util.CamelCaseConfigConverter 1.conf
24/11/05 12:24:20 ERROR CamelCaseConfigConverter: Error processing file: 
java.io.IOException: Error getting stage properties from config.
	at com.kmwllc.lucille.util.CamelCaseConfigConverter.getStagePropertiesFromConfig(CamelCaseConfigConverter.java:86) ~[lucille.jar:0.4.1-SNAPSHOT]
	at com.kmwllc.lucille.util.CamelCaseConfigConverter.processFile(CamelCaseConfigConverter.java:64) ~[lucille.jar:0.4.1-SNAPSHOT]
	at com.kmwllc.lucille.util.CamelCaseConfigConverter.main(CamelCaseConfigConverter.java:108) [lucille.jar:0.4.1-SNAPSHOT]
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because "pipelinesMap" is null
	at com.kmwllc.lucille.util.CamelCaseConfigConverter.getStagePropertiesFromConfig(CamelCaseConfigConverter.java:75) ~[lucille.jar:0.4.1-SNAPSHOT]
	... 2 more

validation-example.conf
```diff
@@ -44,12 +44,12 @@ pipelines: [
...
         class = "com.kmwllc.lucille.stage.Concatenate"
-        format_string = "{city}, {state}, {country}"
+        formatString = "{city}, {state}, {country}"
       },
       {
         class = "com.kmwllc.lucille.stage.Concatenate"
         dest = "dest"
-        format_string = "{city}, {state}, {country}"
+        formatString = "{city}, {state}, {country}"
         default_inputs3 {
```
rseitz, line 53: default_inputs3 still has an underscore

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.apache.commons.text.CaseUtils;
+
+public class CamelCaseConfigConverter {
```
rseitz, line 17: We do need a unit test for this. Can be a simple test that runs the conversion on one input conf file and confirms that the output matches what we expect. The conf file should have some examples of edge cases like multiple instances of the same stage property name on one line (used as a property and as a value) and some other snake case properties in connectors or elsewhere that shouldn't be converted because they're not stage properties

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
Jozurf: updated validateArguments to log error if no arguments were provided and also now checks if file exists before checking file extension in [63261d5](https://github.com/kmwtechnology/lucille/pull/203/commits/63261d5ae1fc5ee23031a4e9f3c14d620c7e13d8)

validation-example.conf
Jozurf: updated in [63261d5](https://github.com/kmwtechnology/lucille/pull/203/commits/63261d5ae1fc5ee23031a4e9f3c14d620c7e13d8)

rseitz (approved):

lucille-core/src/main/java/com/kmwllc/lucille/util/CamelCaseConfigConverter.java
Jozurf: WIP

Jozurf: created test for this in [1134f0a](https://github.com/kmwtechnology/lucille/pull/203/commits/1134f0ab1f7385d55758becbd4ca91cab1c4ba1f)

spencersolomon6 (approved):

kiratraynor:
I'm not sure if we'd want to change these or not, but there are a few examples of stages that have values that are in snake case like `NormalizeText` (with `title_case`, `sentence_case`), `ChunkText` (with the doc fields its adding), `Condition` (with 'must_not'), or just `run_id` in general. It seems like the changes that were made were ensuring that the stage properties themselves were in camel case, but just wondering if we would we want the values like these to be as well so that it's more consistent?

lucille-core/src/main/java/com/kmwllc/lucille/stage/ChunkText.java
```diff
@@ -53,25 +53,25 @@
  *       - "offset" : number of character offset from start of document
  *       - "length" : number of characters in this chunk
  *       - "chunk_number" : chunk number
- *       - "total_chunk_number" : total chunk number produced from parent document
+ *       - "total_chunks" : total chunk number produced from parent document
```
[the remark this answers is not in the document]
Jozurf: @kiratraynor to answer both of your questions. I did have the same thoughts, and short answer is, that we can, but this would make it a nightmare to create a script for conversion. Some external libraries (like Tika for TikaExtractor) produces snake_case for lucille documents. And so any other stages further down the pipeline that takes in the name of the field (like source: "field_produced_by_tika") cannot be changed in the script convertor. Also, there is the case where  CSVConnector, SolrConnector, VFSConnector, DatabaseConnector could produce fields with snake_case, meaning that the script would have to know which fields not to convert in cases like this, especially for CSVConnector and DatabaseConnector, which is dependent on the input.

kiratraynor (approved):

PR #210, opened 2024-11-18
LC-568: update all github actions versions

Jozurf:
Idea to upgrade came about when creating Jlama stage. In the PR, it ended up using maven toolchains for separating the build process, which in documentation, were using github actions v4 api.


.github/workflows/hugo.yml
```diff
@@ -40,12 +40,12 @@ jobs:
...
         run: sudo snap install dart-sass
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           submodules: recursive
       - name: Setup Pages
         id: pages
-        uses: actions/configure-pages@v3
+        uses: actions/configure-pages@v4
```
rseitz, line 48: it looks like each type of action is versioned separately
configure-pages is at v5 -- see here: https://github.com/actions/configure-pages/releases
need to look at each action type we're using and upgrade it separately
there's actually a github action for doing this, but probably overkill for us given that we're not using that many different action types: https://github.com/marketplace/actions/github-actions-version-updater

Jozurf: changed all actions to latest versions in [711d0de](https://github.com/kmwtechnology/lucille/pull/210/commits/711d0debbac0e47d6d4c8d1bc89ae2d281843d85)

rseitz (approved):

PR #212, opened 2024-11-19
LC-482: change value to test action failure

Jozurf:


rseitz:
this PR intentionally makes the distributed mode test fail so we can see whether this failure is properly reported in the context of the github action that runs it. this PR will be reverted after we do this test.

rseitz (approved):

PR #220, opened 2024-12-06
LC-574: update javadoc for ChunkText stage

Jozurf:


lucille-core/src/main/java/com/kmwllc/lucille/stage/ChunkText.java
```diff
@@ -53,15 +56,21 @@
  *       - "offset" : number of character offset from start of document
  *       - "length" : number of characters in this chunk
  *       - "chunk_number" : chunk number
- *       - "total_chunk_number" : total chunk number produced from parent document
+ *       - "total_chunks" : total chunk number produced from parent document
```
Jozurf, line 59: found error in javadocs

PR #222, opened 2024-12-18
LC-582/LC-583 improve AddRandomFields stage

rseitz:
This PR includes several improvements to the AddRandomFields stage:

1) a "concatenate" mode now allows for multiple random values to be set as a single space-separated string instead of as multiple field values. i.e. "14 34 323 54" vs. ["14", "34", "323", "54"]
2) when selecting random values from an integer range, the stage no longer generates a list of all integers in the range
3) when reading data from a file, the stage no longer attempts to uniquify the contents (which can be very time consuming for a large file)

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
@@ -142,7 +142,7 @@ public class AddRandomField extends Stage {
   private static List<String> getRandomIntegerStringList(int length, int rangeSize) {
     List<String> result = new ArrayList();
     for (int i = 0; i < length; i++) {
-      result.add(String.valueOf(ThreadLocalRandom.current().nextInt(0, rangeSize + 1)));
+      result.add(String.valueOf(ThreadLocalRandom.current().nextInt(0, rangeSize)));
     }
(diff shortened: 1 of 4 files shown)


spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomField.java
```diff
@@ -45,31 +41,36 @@
...
   private Integer minNumOfTerms;
   private Integer maxNumOfTerms;
   private final boolean isNested;
-  private List<String> dataArr;
-  private List<String> uniqueValues;
+  private List<String> fileData;
+  private final boolean concatenate;
 
   public AddRandomField(Config config) throws StageException {
     super(config, new StageSpec().withOptionalProperties("input_data_path", "field_name", "range_size", "min_num_of_terms",
```
spencersolomon6, line 64: field_name should be a required field

rseitz: existing behavior is to default to "data" if field_name is not provided:
`    this.fieldName = ConfigUtils.getOrDefault(config, "field_name", "data");`

PR #223, opened 2024-12-19
LC-593 Create new Stages for Generating Random Test Data

spencersolomon6:
Improve our ability to produce randomized test data by adding support for generating fields containing ints, doubles, booleans and dates.


lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomDate.java
```diff
@@ -0,0 +1,80 @@
...
+        Instant.ofEpochMilli(ThreadLocalRandom.current().nextLong(
+        rangeStartDate.getTime(), rangeEndDate.getTime()
+    )));
+
+    doc.setField(fieldName,  randomDate);
+
+    return null;
+  }
+
+  protected Date LocalDateToDate(LocalDate ld) {
```
rseitz, line 75: method name should begin with lowercase l
suggest making this private, not protected

kwatters (approved):
LGTM

PR #225, opened 2024-12-27
LC-575: Add parent field copying to EmitNestedChildren Stage

Jozurf:


lucille-core/src/main/java/com/kmwllc/lucille/stage/EmitNestedChildren.java
```diff
@@ -36,6 +41,26 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+      for (Document child : children) {
+        for (String field : fieldsToCopy) {
+          if (!doc.has(field)) {
+            log.warn("Field '{}' not found in parent document", field);
+            continue;
+          }
+          // handle multivalued fields if needed, we want to be a list if multivalued
+          // Note we are grabbing the String version of the contents
+          if (doc.isMultiValued(field)) {
+            child.update(field, updateMode, doc.getStringList(field).toArray(new String[0]));
```
rseitz, line 56: can we do this in a way that preserves the type of the values and doesn't convert to String?

Jozurf: resolved in [0c58d17](https://github.com/kmwtechnology/lucille/pull/225/commits/0c58d1712c7de03e9789791db3e64eba527e0fe1)

lucille-core/src/main/java/com/kmwllc/lucille/stage/EmitNestedChildren.java
```diff
@@ -15,16 +16,20 @@
...
   private final boolean dropParent;
+  private final List<String> fieldsToCopy;
+  private final UpdateMode updateMode;
   private static final Logger log = LoggerFactory.getLogger(EmitNestedChildren.class);
 
   public EmitNestedChildren(Config config) {
-    super(config, new StageSpec().withOptionalProperties("drop_parent"));
+    super(config, new StageSpec().withOptionalProperties("drop_parent", "fields_to_copy", "update_mode"));
     this.dropParent = config.hasPath("drop_parent") ? config.getBoolean("drop_parent") : false;
+    this.fieldsToCopy = config.hasPath("fields_to_copy") ? config.getStringList("fields_to_copy") : List.of();
```
kwatters, line 31: this should be a map.. not a list.. so you can copy a field from the parent to a different field on the child.

PR #231, opened 2025-01-02
Explicitly putting copy dependencies just after packaging

supertick:


kwatters:
we need to make sure the dependencies are in the .tar.gz that is built in the package phase.  moving this plugin to post-package causes the .tar.gz in the target/ folder to not have the dependencies in it..

lucille-examples/lucille-distributed-example/pom.xml
```diff
@@ -77,7 +77,7 @@
         <executions>
           <execution>
             <id>copy-dependencies</id>
-            <phase>package</phase>
```
kwatters, line 80: try moving this to prepare-package instead.  this issue seems only to be triggered inside eclipse , potentially with the newer versions of maven...

PR #232, opened 2025-01-02
LC-576: S3 examples with requested changes

supertick:


lucille-examples/lucille-s3-ingest-example/scripts/run_ingest.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# run this script from top level lucille-simple-csv-solr-example directory via ./scripts/run_ingest.sh
```
rseitz, line 2: comment references lucille-simple-csv-solr-example but this is lucille-s3-ingest-example

lucille-examples/lucille-s3-ingest-example/conf/s3-opensearch.conf
```diff
@@ -0,0 +1,69 @@
...
+}
+
+opensearch {
+  url: ${INGEST_URL}
+  index: ${INGEST_INDEX}
+  acceptInvalidCert: true
+}
+
+worker {
+  pipeline: "pipeline1"
```
rseitz, line 67: it's only necessary to provide worker.pipeline when starting a Worker in a separate JVM as part of a distributed mode deployment
When running lucille in "local" mode with all components in the same JVM, the Runner will create workers for each pipeline used in the run, and you don't need to specify worker.pipeline

supertick: removed

lucille-examples/lucille-s3-ingest-example/scripts/run_ingest.sh
supertick: updated

rseitz (approved):

PR #234, opened 2025-01-03
LC-579 Lucille as a service

supertick:


lucille-plugins/lucille-api/conf/api.yml
```diff
@@ -9,4 +9,18 @@ logging:
...
\ No newline at end of file
+  enabled: false
+  password: "password"
+
+swagger:
+  resourcePackage: "com.kmwllc.lucille.endpoints"
+  title: Lucille API
+  description: Lucille API
+  version: 0.0.1
+  contact: kmwllc@gmail.com
```
kwatters, line 20: lets update this to be info@kmwllc.com

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
```diff
@@ -36,31 +40,52 @@ public String getName() {
...
+	        protected SwaggerBundleConfiguration getSwaggerBundleConfiguration(LucilleAPIConfiguration configuration) {
+	            return configuration.swaggerBundleConfiguration;
+	        }
+	    });
+
+  }
 
   @Override
   public void run(LucilleAPIConfiguration config, Environment env) throws Exception {
+    System.out.println(String.format("starting lucille-api from %s config %s env %s", System.getProperty("user.dir"), config, env));
```
kwatters, line 56: we shouldn't use system.out , the preference is the logger

supertick: I can change it to a logger, but since this is the entry point for all of it I usually have System.out dump the cmd and env vars since they can affect (sometimes adversely or unexpectedly) the logging ...

gperryatnasuni: all other logging output is done with a logger

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/AuthHandler.java
```diff
@@ -0,0 +1,30 @@
+package com.kmwllc.lucille;
+
+import java.util.Optional;
+
+import io.dropwizard.auth.PrincipalImpl;
+import jakarta.ws.rs.core.Response;
+
+public class AuthHandler {
+
+    private final boolean authEnabled;
```
rseitz, line 10: lucille follows the google java coding conventions -- so block indent is 2 spaces
https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation
same comment applies elsewhere in this PR

gperryatnasuni: currently using formatter @rseitz suggested
image

supertick: updated eclipse formatter

lucille-plugins/lucille-api/src/test/java/endpoints/LucilleResourceTest.java
```diff
@@ -1,49 +1,160 @@
...
+    private RunnerManager runnerManager;  // Real instance
+    private LucilleResource lucilleResource;  // Real instance
+
+    @Mock
+    private AuthHandler authHandler;  // Mock authentication only
+
+    private Optional<PrincipalImpl> mockUser;
+
+    @Before
+    public void setUp() {
```
rseitz, line 37: please fix indentation

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/StandardErrorResponse.java
```diff
@@ -0,0 +1,36 @@
+package com.kmwllc.lucille.endpoints;
+
+import java.util.HashMap;
+import java.util.Map;
+import jakarta.ws.rs.core.Response;
+
+public class StandardErrorResponse {
```
rseitz, line 7: please fix indentation

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/AuthHandler.java
rseitz: this file is still indented with 4 spaces

supertick: updated

lucille-plugins/lucille-api/src/test/java/endpoints/LucilleResourceTest.java
supertick: updated

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/StandardErrorResponse.java
supertick: formatted

lucille-plugins/lucille-api/src/test/java/endpoints/LivenessResourceTest.java
```diff
@@ -16,8 +19,24 @@ public class LivenessResourceTest {
 
   @Test
   public void testLivenessEndpoint() {
+    System.out.println("=======================================================================================================================");
     Optional<PrincipalImpl> user = Optional.of(new PrincipalImpl("test"));
-    Response response = liveness.isAlive(user);
+    Response response = liveness.isAlive();  
+    
+    // Dump relevant response details
+    System.out.println("=== Response Dump ===");
```
rseitz, line 27: please remove System.out.printlns; please use junit assertions and/or the logger

code that useful for development but not necessary for the test itself should either be removed or commented out with a quick explanation of why it's there and why someone would want to uncomment it

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -231,17 +231,21 @@ static RunType getRunType(boolean useKafka, boolean local) {
       return RunType.LOCAL;
     }
   }
+  
+  public static RunResult runWithResultLog(Config config, RunType runType) throws Exception {
+	  return runWithResultLog(config, runType, null);
```
rseitz, line 236: indentation

lucille-plugins/lucille-api/src/test/java/endpoints/LucilleResourceTest.java
```diff
@@ -1,49 +1,161 @@
...
+  @Test
+  public void testGetAllRuns_Empty() {
+    when(authHandler.authenticate(any())).thenReturn(null);
+    Response response = lucilleResource.getAllRuns(mockUser);
 
-    } finally {
-      assertEquals(200, status.getStatus());
-    }
+    assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
+    System.out.println(response.getEntity());
```
rseitz, line 158: remove System.out.println

lucille-plugins/lucille-api/src/test/java/auth/BasicAuthenticatorTest.java
```diff
@@ -32,9 +37,9 @@ public void testAuthFailure() {
      */
     Optional<PrincipalImpl> noUser = Optional.empty();
     RunnerManager runnerManager = RunnerManager.getInstance();
-    LucilleResource api = new LucilleResource(runnerManager);
-
-    Response status = api.getRunStatus(noUser);
-    assertEquals(401, status.getStatus());
+    LucilleResource api = new LucilleResource(runnerManager, authHandler);
+    String runId = "runId";
+//    Response status = api.getRunStatus(noUser, runId);
```
rseitz, line 42: remove commented out code

lucille-plugins/lucille-api/src/test/java/auth/BasicAuthenticatorTest.java
```diff
@@ -19,10 +24,10 @@ public void testAuthSuccess() {
      */
     Optional<PrincipalImpl> user = Optional.of(new PrincipalImpl("test"));
     RunnerManager runnerManager = RunnerManager.getInstance();
-    LucilleResource api = new LucilleResource(runnerManager);
-
-    Response status = api.getRunStatus(user);
-    assertEquals(200, status.getStatus());
+    LucilleResource api = new LucilleResource(runnerManager, authHandler);
+    String runId = "runId";
+    // Response status = api.getRunStatus(user, runId);
```
rseitz, line 29: remove commented out code

lucille-plugins/lucille-api/src/test/java/APIIntegrationTest.java
```diff
@@ -20,12 +24,17 @@ public class APIIntegrationTest {
 
   private final Client client = RULE.client();
   private final String url = String.format("http://localhost:%d/", RULE.getLocalPort());
-  private final String authHeader = "Basic " + Base64.getEncoder().encodeToString("test:test".getBytes());
+  private final String authHeader = "Basic " + Base64.getEncoder().encodeToString("admin:password".getBytes());
+
+//  @BeforeAll
```
rseitz, line 29: remove commented out code

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/ReadinessResource.java
```diff
@@ -1,35 +1,40 @@
...
   public ReadinessResource() {
     super();
   }
 
   @GET
-  public Response isReady(@Auth Optional<PrincipalImpl> user) {
+  @Operation(summary = "Readiness Check", description = "Returns 204 No Content to indicate the service is ready.")
+  public Response isReady() {
+    log.debug("Readiness check endpoint accessed.");
+    // return Response.noContent().build();
```
rseitz, line 37: remove commented out code

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/AuthHandler.java
```diff
@@ -0,0 +1,30 @@
...
+import jakarta.ws.rs.core.Response;
+
+public class AuthHandler {
+
+  private final boolean authEnabled;
+
+  public AuthHandler(boolean authEnabled) {
+    this.authEnabled = authEnabled;
+  }
+
```
rseitz, line 14: remove extra whitespace

.github/workflows/maven.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         distribution: 'temurin'
         cache: maven
     - name: Build with Maven
-      run: mvn -B verify --file pom.xml
+      run: mvn -q -B verify --file pom.xml
```
rseitz, line 33: why is -q being added as part of this PR -- is it needed for Lucille as a service? -- and what is the consequence?

lucille-core/src/main/java/com/kmwllc/lucille/core/ConfigException.java
```diff
@@ -0,0 +1,29 @@
+package com.kmwllc.lucille.core;
+
+/**
+ * Exception for invalid configuration of a Lucille run.
+ */
+public class ConfigException extends Exception {
```
rseitz, line 6: the config library we use has `com.typesafe.config.ConfigException` and most problems with a config will be contained in that exception

does lucille need it's own Config exception separate from com.typesafe.config.ConfigException

in this PR, I only see the new ConfigException being thrown when the config file can't be found. A generic FileNotFoundException would probably suffice in that one case?

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,204 @@
...
+   * @throws Exception if the run throws an exception or is interrupted
+   */
+  public void waitForRunCompletion(String runId) throws Exception {
+    RunDetails details = runDetailsMap.get(runId);
+    if (details != null) {
+      details.getFuture().get();
+    } else {
+      log.error("No such run with ID: " + runId);
+      return;
+      // throw new IllegalArgumentException("No such run with ID: " + runId);
```
rseitz, line 62: remove commented out code

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,204 @@
...
+   * @param runId the unique ID for the lucille run
+   * @param config the configuration to use for the run
+   * @return true if the run was started successfully; false if a run with the given ID is already
+   *         in progress
+   * @throws Exception 
    */
-  synchronized protected boolean runWithConfig(Config config) {
-    if (isRunning()) {
-      log.warn("Skipping new run; previous lucille run is still in progress.");
+  public synchronized boolean runWithConfig(String inRunId, String configId) throws ConfigException {
```
rseitz, line 116: the argument can be called runId and you can reassign it if it is null
there is not a need for a separate inRunId parameter here

lucille-core/src/main/java/com/kmwllc/lucille/core/ConfigException.java
rseitz: ok so it's not being thrown when a file isn't found, rather when the config doesn't have an entry in the map
so probably what we need here is something like a RunnerManagerException for problems that can arise around starting/stopping a run

lucille-plugins/lucille-api/pom.xml
```diff
@@ -51,54 +51,70 @@
...
+            <version>${log4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>${log4j.version}</version>
+        </dependency>
+
+        <!-- Logback for testing only -->
+        <dependency>
```
rseitz, line 112: there are a bunch of diffs in this file (from reformatting?) but it looks like the only changes are the addition of swagger and logback. What tests depend on logback?

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
```diff
@@ -36,31 +39,60 @@ public String getName() {
...
-    env.jersey().register(new LucilleResource(runnerManager));
+    AuthHandler authHandler = new AuthHandler(authEnabled);
+    env.jersey().register(new LucilleResource(runnerManager, authHandler));
     env.jersey().register(new LivenessResource());
     env.jersey().register(new ReadinessResource());
     env.jersey().register(new AuthValueFactoryProvider.Binder<>(PrincipalImpl.class));
   }
 
   public static void main(String[] args) throws Exception {
+    System.out.println(String.format("starting lucille-api from %s args %s",
```
rseitz, line 88: use the logger please
this is a codebase that avoids System.out.println and uses logging

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LucilleResource.java
```diff
@@ -1,61 +1,211 @@
...
+    this.authHandler = authHandler;
+  }
+
+  @POST
+  @Tag(name = "Config")
+  @Path("/config")
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Operation(summary = "Create a new Lucille config to be run later",
+      description = "Creates a new Lucille run with the specified configuration")
+  public Response createRun(@Parameter(hidden = true) @Auth Optional<PrincipalImpl> user,
```
rseitz, line 57: rename method to createConfig

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LucilleResource.java
```diff
@@ -1,61 +1,211 @@
...
+      // Extract configId from the request body
+      String configId = requestBody.get("configId");
+      if (configId == null || configId.isBlank()) {
+        return StandardErrorResponse.buildResponse(Response.Status.BAD_REQUEST,
+            "configId is required in the request body.");
+      }
 
-    if (user.isPresent()) {
-      boolean status = runnerManager.run();
+      String runId = UUID.randomUUID().toString();
```
rseitz, line 152: RunnerManager can generate a runId, right?
If so, we don't need to generate the ID here in API code.
The goal should be to minimize the number of different places where runIds get created and ideally confine it to Runner/RunnerManager only.

supertick: It could, but RunManager gave the same option for something external to supply a runId.
In this case, its necessary because the contract is to get the RunDetails which this api consistently does for all started, and running "runs".   RunDetails are key'ed by RunId.  I didn't feel the RunManager needed to know about RunDetails, so in this case the runId is supplied externally.  Potentially it could be an optional parameter in the api .. I could imagine some test cases where this might be useful for full end-to-end, but didn't propagate it that far up, this pr.

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
supertick: Heh, all other System.out statements in other files (which there were plenty of) were mistakes, but this one wasn't .. for the reason I gave @kwatters .. which was at the very beginning of an app startup, the startup details logged can change logging to hide reasons why its not logging.  But I understand I didn't convince anyone.. I'll update.

supertick: fixed

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
```diff
@@ -36,31 +39,60 @@ public String getName() {
...
+          LucilleAPIConfiguration configuration) {
+        return configuration.swaggerBundleConfiguration;
+      }
+    });
+
+  }
 
   @Override
   public void run(LucilleAPIConfiguration config, Environment env) throws Exception {
+    System.out.println(String.format("starting lucille-api from %s config %s env %s",
```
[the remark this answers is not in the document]
supertick: fixed

.github/workflows/maven.yml
supertick: Trying to quite the 18 Megs of logs during a regular green run ... it didn't work. backing it out

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/AuthHandler.java
supertick: fixed

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LivenessResource.java
```diff
@@ -1,37 +1,40 @@
...
   public LivenessResource() {
     super();
   }
 
   @GET
-  public Response isAlive(@Auth Optional<PrincipalImpl> user) {
+  @Operation(summary = "Liveness Check", description = "Returns 204 No Content to indicate the service is live.")
+  public Response isAlive() {
+    log.debug("Liveness check endpoint accessed.");
+    // return Response.noContent().build();
```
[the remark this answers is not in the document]
supertick: fixed

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/ReadinessResource.java
supertick: fixed

lucille-core/src/main/java/com/kmwllc/lucille/core/ConfigException.java
supertick: ok changing ConfigException to RunnerManagerException

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: wasn't sure if it should return with error or throw - the original implementation had no similar use case
kinder-gentler log error was a guess..
removed comment

lucille-plugins/lucille-api/src/test/java/APIIntegrationTest.java
supertick: removed

lucille-plugins/lucille-api/src/test/java/endpoints/LucilleResourceTest.java
supertick: removed

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
supertick: fixed

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -256,12 +260,32 @@ public static RunResult runWithResultLog(Config config, RunType runType) throws
...
-   * Generates a run ID and performs an end-to-end run of the designated type.
+   * Non managed Run with internal generated runId 
+   * @param config
+   * @param type
+   * @return
+   * @throws Exception
    */
   public static RunResult run(Config config, RunType type) throws Exception {
-    String runId = UUID.randomUUID().toString();
+	  return run(config, type, null);
```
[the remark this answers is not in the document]
supertick: fixed

lucille-plugins/lucille-api/src/test/java/endpoints/LivenessResourceTest.java
supertick: removed - was trying to figure out what the intent/purpose of the two copied kubernetes like pre-existing calls were suppose to be doing.
removed

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LucilleResource.java
rseitz: This comment is about the java method `public Response createRun(` not the `POST config` endpoint name. Yes, we agreed to POST config in the meeting and that's all good, but it looks like the java method name here is left over from before. Also, the Operation summary "Create a new Lucille config to be run later" is accurate. But the Operation description "Creates a new Lucille run with the specified configuration" looks like it should be updated.

supertick: :facepalm: Thanks, of course, that makes sense.
fixed

lucille-plugins/lucille-api/src/test/java/auth/BasicAuthenticatorTest.java
supertick: test fixed

lucille-plugins/lucille-api/src/test/java/auth/BasicAuthenticatorTest.java
supertick: test fixed

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LucilleResource.java
```diff
@@ -1,61 +1,211 @@
...
+  @GET
+  @Tag(name = "Config", description = "Retrieve Lucille configuration details.")
+  @Path("/config/{configId}")
+  @Operation(summary = "Get Lucille config",
+      description = "Retrieves a specific Lucille config by ID.",
+      security = @SecurityRequirement(name = "basicAuth"))
+  public Response getConfig(@Parameter(hidden = true) @Auth Optional<PrincipalImpl> user,
+      @Parameter(description = "The UUID of the configuration to retrieve.", required = false,
+          example = "fca83cb6-c2c2-4cbf-93ef-41c08d5d4b58") @PathParam("configId") String configId) {
+    Response authResponse = authHandler.authenticate(user);
```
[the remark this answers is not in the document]
supertick: Yes, it can be normalized to the class level, and normalized further I believe by creating a ContainerRequestFilter with jersey and registering it.  But, it probably would be worthwhile to delve deeper and have a discussion about what authentication & authorization Lucille should support.  Perhaps not this pr ?

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: I can't do this because runId is required to be final in the async body.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,176 @@
...
+  private RunType runType;
+
+  @Schema(description = "Exception of run")
+  private Throwable exception;
+
+
+  public RunDetails(String runId, String configId) {
+    this.runId = runId;
+    this.configId = configId;
+    this.future = new CompletableFuture<>();
```
rseitz, line 53: It seems like "future" is being set to an empty CompletableFuture that never gets populated
This "future" is different from the one RunnerManager creates in runWithConfig where it does `CompletableFuture.runAsync(() -> {`. RunnerManager used to save that future in an instance variable but it's not doing that anymore. So it seems like the actual future that's being used to contain the Lucille run is never stored anywhere but instead we're storing an empty future. How are tests passing then? Please do let me know if I'm overlooking something.

supertick: Important catch, thanks, looking into it.

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -1,56 +1,128 @@
...
-    runnerManager.waitForRunCompletion();
-    assertFalse(runnerManager.isRunning());
+    assertTrue(runnerManager.isRunning(runId));
+    runnerManager.waitForRunCompletion(runId);
+    assertFalse(runnerManager.isRunning(runId));
   }
+
+
+  @Test
+  public void testSimultaneousRuns() throws Exception {
```
rseitz, line 71: As an experiment, try using a SleepConnector (which is in the test tree) to make each lucille run take, say, 20 seconds. You can do that like this:

```
    String config = "connectors: [\n" +
        "    {class: \"com.kmwllc.lucille.connector.SleepConnector\", name: \"connector1\", duration: 20000, pipeline: \"pipeline1\"}\n" +
        "]\n" +
        "pipelines: [{name: \"pipeline1\", stages: []}]\n" +
        "indexer {\n" +
        "  sendEnabled: false\n" +
        "}\n";
```

Now if you do `Runner.runInTestMode(c);` it indeed takes 20 seconds.
However, if you run this unit test as-is, all the 5 Lucille runs finish quickly and you see output like this:

```
25/01/14 10:28:46 INFO Runner: Connector connector1 feeding to pipeline pipeline1 complete. Time: 0.53 secs.
```

So there's a significant difference in behavior here, depending on whether you run Lucille in the traditional way or using the new code in this PR -- we need to understand that.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
rseitz: OK. The variable name "inRunId" tripped me up when I first saw it. How about "providedRunId" ?

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -1,56 +1,128 @@
...
+    String config = "connectors: [\n" +
+        "    {class: \"com.kmwllc.lucille.connector.CSVConnector\", path: \"conf/dummy.csv\", name: \"connector1\", pipeline: \"pipeline1\"}\n" +
+        "]\n" +
+        "pipelines: [{name: \"pipeline1\", stages: []}]\n" +
+        "indexer {\n" +
+        "  type: \"CSV\"\n" +
+        "}\n" +
+        "csv {\n" +
+        "  columns: [\"Name\", \"Age\", \"City\"]\n" +
+        "  path: \"conf/dummy.csv\"\n" +
```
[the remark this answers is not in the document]
rseitz: also note that we are asking 5 lucille runs to all write to the same dummy.csv here
that means if we wanted to enhance this test to check that the lucille runs actually generate the proper output, we wouldn't be able to because they're all overwriting each other's output

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
rseitz: If you are only trying to store the status of the run here in RunDetails, please do not use a dummy CompletableFuture, that is not what it's meant for. It's meant for executing asynchronous tasks but you are not using it to execute anything here. If you want to be able to block in a thread-safe way you would need to have the "real" CompleteableFuture that contains the lucille run and you'd call get() or join() on it. If you do not need to do that here but only want to set/and get status information about the run, consider a simple boolean isDone or simply check whether endTime has been set.

supertick: updated

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,179 @@
...
+    return future;
+  }
+  
+  public void setFuture(CompletableFuture<Void> future) {
+    this.future = future;
+  }
+
+  public void complete() {
+    this.endTime = Instant.now();
+    this.future.complete(null);
```
rseitz, line 106: When you create a CompletableFuture via runAsync, per the docs: "Returns a new CompletableFuture that is asynchronously completed by a task running in the [ForkJoinPool.commonPool()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#commonPool--) after it runs the given action."

What is the point of attempting to complete the future a second time here?

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  public void completeExceptionally(Throwable ex) {
+    this.endTime = Instant.now();
+    this.errorCount++;
+    this.exception = ex;
+    this.future.completeExceptionally(ex);
```
rseitz, line 113: Again, the future that is created via runAsync is also completed as part of that process. I do not understand the rationale for completing it again here.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
supertick: none now that its using the appropriate future

supertick: removed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
supertick: removed this.future.completeExceptionally(ex); , keeping 
RunDetail.completeExceptionally(ex) & RunDetail.complete() to capture details

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
supertick: Significant difference is there was no "wait" on the test, which did not make it a very effective test.
Fixed and added random interval between starting runs for more entropy.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
rseitz: I'm inclined towards throwing an exception
giving the appearance that nothing went wrong (if the runId isn't found) probably won't help anyone in the end...

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
rseitz: Let's require that the runId argument be non-null here. Throw an exception if it's null.

In `runWithConfig(Config config)` where we're currently delegating to runWithConfig(null, configId), we can generate the runId there instead of passing null.

To standardize the way runIds are generated, let's add this to Runner:

```
  /**
   * Utility method to generate a Run ID.
   */
  public static String generateRunId() {
    return UUID.randomUUID().toString();
  }
```
and use it in any place where we need to generate a runId

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LucilleResource.java
rseitz: You wrote "I didn't feel the RunManager needed to know about RunDetails" but looking at RunnerManager I see many references to RunDetails so those classes look to me like they're intended to work together. 
RunnerManager.runWithConfig should be updated to return RunDetails instead of a boolean. The boolean made sense when RunDetails didn't exist, but now the RunDetails does exist, it the more natural return type for a method that launches a run. After this change, we'll be able to get the generated runId from the returned RunDetails and we don't have to generate the runId here and pass it in.

supertick: Ok, that makes sense to me.  But this will generate a lot of signature changes, and pre-appologies .. I pressed the auto-format on this, so you can see what google standard format definition does in eclipse

supertick: updated

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -1,56 +1,162 @@
...
+
+    log.info("after waiting for run to complete {} ========", runnerManager.isRunning(runId));
+    assertFalse(runnerManager.isRunning(runId));
+
+    log.info("Test execution time: {} ms", durationMillis);
   }
+
+
+  @Test
+  public void testSimultaneousRuns() throws Exception {
```
rseitz, line 83: Here is a proposal for some enhancements to this test. Summary: loading config from file instead of inline string (suggesting we change the sleep duration in that file to 1000). Removing verbose output. Removing countdown latch. Allowing exceptions to propagate up and fail the test. Separating the starting of the 5 threads from the waiting. 

```
  @Test
  public void testSimultaneousRuns2() throws Exception {
    Config config = ConfigFactory.load("RunnerManagerTest/sleep.conf");
    RunnerManager runnerManager = RunnerManager.getInstance();
    String configId = runnerManager.createConfig(config);
    List<String> runIds = new ArrayList();

    for (int i = 0; i < 5; i++) {
      String runId = "test-run-" + i;
      // runnerManager.runWithConfig is non-blocking so we don't need to invoke it here via CompleteableFuture.runAsync()
      // but this approach simulates a scenerio where multiple threads are calling it concurrently
      CompletableFuture.runAsync(()-> {
        try {
          runnerManager.runWithConfig(runId, configId);
          assertFalse(runnerManager.getRunDetails(runId).isDone());
        } catch (RunnerManagerException e) {
          throw new RuntimeException(e);
        }
      });
      runIds.add(runId);
    }

    assertEquals(5, runIds.size());

    StopWatch stopWatch = new StopWatch();
    stopWatch.start();
    while (runIds.stream().anyMatch(x -> !runnerManager.getRunDetails(x).isDone())) {
      if (stopWatch.getTime(TimeUnit.SECONDS)>10) {
        fail("5 concurrent Lucille Runs are taking longer than 10 seconds to complete.");
      }
      Thread.sleep(100);
    }

    for (String runId : runIds) {
      RunDetails details = runnerManager.getRunDetails(runId);
      assertEquals(runId, details.getRunId());
      assertEquals(runId, details.getRunResult().getRunId());
      assertTrue(details.getRunResult().getStatus());
      assertTrue(Duration.between(details.getStartTime(), details.getEndTime()).getSeconds() >= 1,
          "Each run expected to take 1 second or more.");
    }
  }
```

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,177 @@
...
+  @Schema(
+      description = "End time of the run in ISO-8601 format. Null if the run has not completed.",
+      example = "2023-12-20T15:48:00.000Z")
+  private Instant endTime;
+
+  @Schema(description = "Number of documents processed during the run.", example = "5000")
+  private long docsProcessed;
+
+  @Schema(description = "Number of errors encountered during the run.", example = "5")
+  private long errorCount;
```
rseitz, line 32: The RunResult contains a list of ConnectorResults, each of which contains a count of successes and errors for that connector. If we need an errorCount for the API right now, we should add a method to RunResult that returns the total error count and use it. If we don't need an errorCount at present, let's remove errorCount, getErrorCount, and setErrorCount and address this issue later. The important think is to avoid storing these counts in two different places and possibly computing them in two different ways -- we already have a longstanding place where these counts exist, we just lack a method on RunResult for getting them.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,177 @@
...
+
+  public CompletableFuture<Void> getFuture() {
+    return future;
+  }
+  
+  public void setFuture(CompletableFuture<Void> future) {
+    this.future = future;
+  }
+
+  public void complete() {
```
rseitz, line 104: I posted a comment about removing or renaming this method (along with completeExceptionally) but can't see it anymore. I am confused about having two methods complete() and setEndTime() that do the same thing -- they have different names, one takes an arg and one doesn't.

```
  public void setEndTime(Instant endTime) {
    this.endTime = endTime;
  }
```

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,177 @@
...
+    this.endTime = Instant.now();
+  }
+
+  public void completeExceptionally(Throwable ex) {
+    this.endTime = Instant.now();
+    this.errorCount++;
```
rseitz, line 110: I have suggested removing the errorCount entirely and getting any needed counts from RunResult. In the meantime, I notice that this method is incrementing the count by 1 each time is called but how is that supposed to work? CompleteExceptionally is meant to be called at most once, right? Because a given run can only fail once.

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
supertick: updated

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,177 @@
...
+
+  public Instant getEndTime() {
+    return endTime;
+  }
+
+  public void setEndTime(Instant endTime) {
+    this.endTime = endTime;
+  }
+
+  public long getDocsProcessed() {
```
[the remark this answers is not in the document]
supertick: removed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
supertick: An error/exception can occur before a RunResult is created.

```java
    final Config config = configMap.get(configId);
    if (config == null) {
      log.error("Config with id {} not found", configId);
      RunnerManagerException exception = new RunnerManagerException("Config with id " + configId + " not found");
      runDetails.completeExceptionally(exception);
      throw exception;
    }
```
It looks like runWithResultLog can throw, so I was thinking
```java
      } catch (Exception e) {
        log.error("Failed to run lucille with ID '{}' via the Runner Manager.", runId, e);
        runDetails.completeExceptionally(e);
      } finally {
```
These seem like errors preventing the run, so I don't know if they should be tallied the same, as errors during the ingest.
Was trying to create a simple consistent data interface with RunDetails, and if a Run doesn't start it would be nice to have that information in the object.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
supertick: Do you want errors preventing the run to be included in those stats ?

rseitz: No, let's leave those stats as they are and not introduce any new forms of error counting or any new places where we're counting errors in this PR.

supertick: ok, thanks for the clarification.  RunDetails still needs a concept of failure, but it makes sense not to conflate this with the current stats of a run.

I have removed errorCount, complete, and setEndTime from RunDetails
https://github.com/kmwtechnology/lucille/pull/234/commits/d4a109b8ba99d55c6cfbd9b144e9bd8139541dcb#diff-7b5758a0d6d59b2f0fca746a826bed9df8ac05a4d49f35f9e984203cdbb01a3eL81-L92

The `exception` field in RunDetails was getting set, and its useful when serialized when a fun fails completely.
Then I added a hasError() in RunDetails which is different than the RunResults stats of a run .. as you said we can deal with rollup if necessary or desired later.

https://github.com/kmwtechnology/lucille/pull/234/commits/d4a109b8ba99d55c6cfbd9b144e9bd8139541dcb#diff-7b5758a0d6d59b2f0fca746a826bed9df8ac05a4d49f35f9e984203cdbb01a3eR120-R122

endTime gets set when RunResult is set (or when an exception occurred preventing the run)

Changed `completeExceptionally` to `setError` so there will be no confusion with futures.

LMK if this satisfies the remaining issues.

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -1,56 +1,143 @@
...
+      runIds.add(runId);
+    }
+
+    assertEquals(5, runIds.size());
+
+    StopWatch stopWatch = new StopWatch();
+    stopWatch.start();
+    while (runIds.stream().anyMatch(x -> {
+      RunDetails details = runnerManager.getRunDetails(x);
+      if (details == null) {
```
rseitz, line 122: why add this null check?
if we look up one of our known runIds and get null, that's an error and we should let a nullpointerexception bubble up and fail the test, not log a message at info level
were you seeing nulls here when running this test locally?

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -20,72 +20,77 @@
 import java.util.concurrent.TimeUnit;
```
rseitz, line 20: can we please avoid reformatting this whole file
this is too much change for this PR and I cannot review every single diff here

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
supertick: errorCount has been removed
Now when an error occurs preventing the run, being thrown up from lower levels and exception is set in RunDetails.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,155 @@
...
+      example = "2023-12-20T14:48:00.000Z")
+  private final Instant startTime;
+
+  @Schema(
+      description = "End time of the run in ISO-8601 format. Null if the run has not completed.",
+      example = "2023-12-20T15:48:00.000Z")
+  private Instant endTime;
+
+  @Schema(description = "Number of documents processed during the run.", example = "5000")
+  private long docsProcessed;
```
rseitz, line 34: as commented elsewhere, this duplicates information already stored in RunResult -- please remove

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -232,36 +237,61 @@ static RunType getRunType(boolean useKafka, boolean local) {
...
+   *
+   * @param config
+   * @param type
+   * @param runId
+   * @return
+   * @throws Exception
+   */
+  public static RunResult run(Config config, RunType type, String runId) throws Exception {
+    if (runId == null) {
+      runId = UUID.randomUUID().toString();
```
rseitz, line 292: why isn't this using the new utility method for generating run Ids?

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,213 @@
...
+   *
+   * @param runId the ID of the run to wait for
+   * @throws Exception if the run throws an exception or is interrupted
+   */
+  public void waitForRunCompletion(String runId) throws Exception {
+    RunDetails details = runDetailsMap.get(runId);
+    if (details != null) {
+      details.getFuture().get();
+    } else {
+      log.error("No such run with ID: " + runId);
```
rseitz, line 61: in general, we avoid logging an error when we are also throwing an exception with the same message. this results in duplicate output in the log. we should rely on whatever code catches or handles this exception to log it

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
supertick: I did see a null, it had to do with the id's not aligning.  That has been fixed, the debugging statement is now removed.

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
supertick: :) fixed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,213 @@
...
-   * run was skipped due to the previous run still existing.
+   * @param runId the unique ID for the lucille run
+   * @param config the configuration to use for the run
+   * @return true if the run was started successfully; false if a run with the given ID is already
+   *         in progress
    */
-  synchronized public boolean run() {
-    return runWithConfig(ConfigFactory.load());
+  public synchronized String createConfig(Config config) {
+    String configId = UUID.randomUUID().toString();
```
rseitz, line 101: why isn't this using the new utility method for generating run IDs??

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
supertick: I know there was interest in seeing the current running status of documents processed, I don't know where that is or how it can be retrieved.  Removing now.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,213 @@
...
+  public List<RunDetails> getRunDetails() {
+    return new ArrayList<>(runDetailsMap.values());
+  }
+
+  /**
+   * Clears all run details
+   * 
+   * @return
+   */
+  public void clearAllRunDetails() {
```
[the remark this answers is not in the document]
rseitz: if you need this for testing, make it into one method that does both actions at the same time and make it a package-private method, not a public method, and add a comment saying it is for testing only

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
supertick: Either kruft from backing out the format or I missed one.
fixed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: Thanks, nice to know, its not always the case in some projects :+1: 
removed.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
rseitz: sorry, missed that, ignore the comment

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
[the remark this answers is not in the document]
supertick: History was discussed very briefly at a lucille meeting, the take away was not to deal with it, however there was a need to clear during tests.
For example this test will fail in LucilleResourceTest
```java
  public void testGetAllRuns_Empty() {
  ...
  ```
  I can get rid of the test and then get rid of the function, is that what you want ?

File: lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java (hunk 33 above)
Comment (supertick) [reply lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java]: setEndTime nor complete exist anymore, endTime is set by finalizing with an exception or setting the RunResults

File: lucille-core/.gitignore [hunk 43]
```diff
@@ -0,0 +1 @@
+/bin/
```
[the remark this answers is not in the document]
supertick: I believe they were the result of running a full command line clean and package with tests.
I don't believe the contents have anything to do with this pr.
image

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,151 @@
...
+  private Instant endTime;
+
+  @Schema(hidden = true)
+  private transient CompletableFuture<Void> future;
+
+  @Schema(description = "Run result")
+  private RunResult runResult;
+
+  @Schema(description = "Current status")
+  private volatile String status = "new";
```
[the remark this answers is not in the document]
supertick: will do.
fixed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,212 @@
...
+    }
+    
+    if (runDetailsMap.containsKey(runId)) {
+      throw new RunnerManagerException("Run with runId " + runId + " already defined - cannot use this runId again");
+    }
+
+    final RunDetails runDetails = new RunDetails(runId, configId);
+    runDetailsMap.put(runId, runDetails);
+
+    if (isRunning(runId)) {
```
rseitz, line 127: you are throwing an exception above if `runDetailsMap.containsKey(runId)`
assuming that the runId is NOT in the map and you haven't thrown an exception, how could isRunning(runId) ever be true here, a few lines later?

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,212 @@
...
+      log.warn("Skipping new run with ID '{}'; previous lucille run is still in progress.", runId);
+      return runDetails;
+    }
+
+    final Config config = configMap.get(configId);
```
rseitz, line 132: this test should go at the beginning
if the config is not found, please don't bother creating the runID
throw the exception early
do not store a runId that can't be started with a RunnerManagerException inside it

lucille-core/.gitignore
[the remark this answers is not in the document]
supertick: hmm .. maybe it was the result of running an example :thinking: ... I'll test now again and see

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,212 @@
...
+            runId, configId);
+
+        // For now, we will always use local mode without Kafka
+        runDetails.setRunType(Runner.getRunType(false, true));
 
-        // For now we will always use local mode without kafka
-        RunType runType = Runner.getRunType(false, true);
 
-        Runner.runWithResultLog(config, runType);
+        log.info(config.entrySet().toString());
```
rseitz, line 149: hmm... why are we logging the entire config here at info level?
this is going to translate into a lot of redundant log output if the same config is used multiple times
change to log.debug or remove

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: I thought it was a good idea to consider the starting of a run as being immutable.
So, if that's the case it should be recorded.  And runId's can never be re-used.  It cannot be stored if it doesn't have a runId

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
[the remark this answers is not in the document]
rseitz: you can remove testGetAllRuns_empty() and testGetAllConfigs_empty() because we have explicitly decided not to deal with clearing state in this PR

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: updated to debug

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
rseitz: if we cannot start a run because we don't have a config, we let's throw an error ASAP
we should not generate a runID for a run that doesn't have enough info to be started
we should not create runDetails

lucille-core/.gitignore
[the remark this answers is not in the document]
supertick: nope, not a output of running example 
directory came back when running
`mvn clean package`

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: If you do that, then from the perspective of a client using the API, a failure to start will not be visible to them, it will only be visible in the lucille logs which a client will not have access to.

lucille-core/.gitignore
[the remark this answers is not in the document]
rseitz: what's in it? how is it being created? 
it is not an expected part of the lucille build process so if you've made a change that's somehow causing a bin directory to be generated we need to track it down. I don't think you have made such a change though. Please keep this .gitignore change out of the current PR, we can create a separate ticket to follow up

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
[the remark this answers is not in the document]
supertick: removed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
rseitz: the API should respond with an error. the user/client should be told about problem immediately when they try to start the run

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: agreed,
isRunning(runId) was left over from previous lucille logic
removed

lucille-core/.gitignore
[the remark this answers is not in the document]
supertick: will do..
removed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
supertick: on the api POST /run the return is a RunDetails
you would prefer a no content in the body and just an error code for this situation?

supertick: A lot has changed during the creation of the pr.
As it stands: 
This code services POST /run
```java
    try {
      // Extract configId from the request body
      String configId = requestBody.get("configId");
      if (configId == null || configId.isBlank()) {
        return StandardErrorResponse.buildResponse(Response.Status.BAD_REQUEST,
            "configId is required in the request body.");
      }

      String runId = Runner.generateRunId();
      RunDetails details = runnerManager.runWithConfig(runId, configId);
      log.info("Lucille run has been triggered. Run ID: " + runId);
      log.info("details: {}", details);

      if (!details.hasError()) {
        return Response.ok(details).build();
      } else {
        return Response.status(424)
            .entity("This Lucille run has been skipped. An instance of Lucille is already running.")
            .build();
      }
    } catch (Exception e) {
      return StandardErrorResponse.buildResponse(Response.Status.BAD_REQUEST,
          "Invalid configuration provided: " + e.getMessage());
    }
 ```
the       if (!details.hasError())  is probably not needed, as an exception will be caught and turned into a "StandardErrorResponse" which I created to try to standardize the json error responses.

I get a clear "standard" error returned when config is not available.
image

So, should I remove the if .. and let the catch handle all exceptions as is?

Comment (supertick) [reply lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java:124]: test now comes before creating RunDetails and removed redundant log.error statement
```java
    final Config config = configMap.get(configId);
    if (config == null) {
      RunnerManagerException exception = new RunnerManagerException("Config with id " + configId + " not found");
      throw exception;
    }
    
    final RunDetails runDetails = new RunDetails(runId, configId);
    runDetailsMap.put(runId, runDetails);
```

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,190 @@
...
+      throw new RunnerManagerException("runId cannot be null");
     }
+    
+    if (runDetailsMap.containsKey(runId)) {
+      throw new RunnerManagerException("Run with runId " + runId + " already defined - cannot use this runId again");
+    }
+
+    final Config config = configMap.get(configId);
+    if (config == null) {
+      RunnerManagerException exception = new RunnerManagerException("Config with id " + configId + " not found");
```
rseitz, line 126: can do this on one line:

```
throw new RunnerManagerException("Config with id " + configId + " not found");
```

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,190 @@
...
-        Runner.runWithResultLog(config, runType);
+        runDetails.setRunResult(Runner.runWithResultLog(config, runDetails.getRunType(), runId));
       } catch (Exception e) {
-        log.error("Failed to run lucille via the Runner Manager.", e);
+        log.error("Failed to run lucille with ID '{}' via the Runner Manager.", runId, e);
+        runDetails.setError(e);
+      } finally {
+        // optionally remove from memory
+        // runDetailsMap.remove(runId);
+        log.info("finished run with id {}", runId);
```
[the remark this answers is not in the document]
supertick: sure,
I'll take it all out... this was remnants I think from some of our discussions on history..
I think the conclusion was to "leave it" ... For a long running lucille, this might be a memory issue at some point, but the details on how to deal with that use case is out of scope.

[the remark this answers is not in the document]
supertick: Do you want me to take the finally block out ?

lucille-plugins/lucille-api/src/test/java/auth/BasicAuthenticatorTest.java
```diff
@@ -1,16 +1,22 @@
...
+
 import io.dropwizard.auth.PrincipalImpl;
 import jakarta.ws.rs.core.Response;
-import java.util.Optional;
-import javax.swing.text.html.Option;
-import org.junit.Test;
 
 public class BasicAuthenticatorTest {
+	
+	  private final AuthHandler authHandler = new AuthHandler(true);
```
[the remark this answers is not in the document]
supertick: fixed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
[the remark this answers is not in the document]
supertick: less is the new more .. :)
removed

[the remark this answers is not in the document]
rseitz: if you want to log success in _starting_ the run you can do that immediately after you call Runner.runWithResultLog 
you already have a place where you are logging an exception
the finally block is not needed currently -- please remove it

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -1,56 +1,133 @@
...
+      runIds.add(runId);
+    }
+
+    assertEquals(5, runIds.size());
+
+    StopWatch stopWatch = new StopWatch();
+    stopWatch.start();
+    while (runIds.stream().anyMatch(x -> {
+      RunDetails details = runnerManager.getRunDetails(x);
+      return details == null || !details.isDone();
```
rseitz, line 115: again, please do not do a null check here unless you are doing that in an assertion that would fail the test
otherwise we want a nullpointer exception to bubble up and fail the test
we should never see a null here and if we do it's an error that needs to be surfaced

supertick: probably kruft from debugging ... removed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,139 @@
...
+
+  public CompletableFuture<Void> getFuture() {
+    return future;
+  }
+
+  public void setFuture(CompletableFuture<Void> future) {
+    this.future = future;
+  }
+
+  public void setError(Throwable ex) {
```
rseitz, line 80: we are using three different similar terms here 
1. "error" as in "setError"
2. Throwable as in the argument type
3. and "exception" as in the name of the instance variable

can we simplify/standardize on throwable?

```
  public void setThrowable(Throwable throwable) {
    this.endTime = Instant.now();
    this.throwable = throwable;
  }

  public boolean hasThrowable() {
    return throwable != null;
  }
```

supertick: updated

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/StandardErrorResponse.java
```diff
@@ -0,0 +1,33 @@
...
```
rseitz, line 7: I don't see any places where we create a new StandardErrorResponse, so there also aren't any places where we call getStatus() or getMessage()...
Also, this class doesn't implement Response so the name StandardErrorResponse is somewhat confusing

I see that the static buildReponse method is used in many places

Please rename this class to ResponseUtils
Please remove the instance variables and accessors and make the constructor private with no args
Can rename buildResponse to buildErrorResponse so a call would look like ResponseUtils.buildErrorResponse()

supertick: :+1: done

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LucilleResource.java
```diff
@@ -1,61 +1,204 @@
...
-      return Response.ok(runStatus).build();
-    } else {
-      return Response.status(401).build();
+  @GET
+  @Tag(name = "Config", description = "Retrieve Lucille configuration details.")
+  @Path("/config/{configId}")
+  @Operation(summary = "Get Lucille config",
+      description = "Retrieves a specific Lucille config by ID.",
+      security = @SecurityRequirement(name = "basicAuth"))
+  public Response getConfig(@Parameter(hidden = true) @Auth Optional<PrincipalImpl> user,
```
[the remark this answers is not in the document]
rseitz: let's not add editing capability in this PR. For the first pass of the new API functionality, it would be what you said, re-upload and get a new configID

lucille-core/src/main/java/com/kmwllc/lucille/core/RunDetails.java
```diff
@@ -0,0 +1,139 @@
...
+    if (runResult != null) {
+      sb.append(", runResult='").append(runResult).append('\'');
+    }
+
+    if (future.isDone()) {
+      try {
+        future.get(); // Check if completed successfully
+        sb.append(", completedSuccessfully=true");
+      } catch (InterruptedException | ExecutionException e) {
+        sb.append(", completedExceptionally='").append(e.getCause().getMessage()).append('\'');
```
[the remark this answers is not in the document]
rseitz: what's being reported as "completedSuccessfully" just means the CompleteableFuture was "completed" i.e. the task ran and stopped and an exception wasn't thrown. However, inside the RunResult, we could still find that all the docs we tried to index errored out. So we might report "completedSuccessfully" even when the run failed.

We need to think about the most useful way to report status back to a user. In the meantime, we should just try to be as narrow and specific as possible about what we're actually reporting. In this case it's that we started a run and it stopped without an exception bubbling up... [stopping typing because I'm out of time...]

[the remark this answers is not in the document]
rseitz: took a closer look and am thinking that the startTime, endTime, and throwable, which are all included in the toString output, should be sufficient for exposing the status of the run

let's remove all interaction with "future" from inside this toString implementation. No future.isDone() or future.get() and no completedSuccessfully or completedExceptionally. We can add that back in later in a future PR if it proves necessary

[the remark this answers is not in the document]
supertick: removed

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -1,80 +1,185 @@
...
-    return !future.isDone();
+  /**
+   * Check whether a specific run is still in progress.
+   *
+   * @param runId the ID of the run to check
+   * @return true if the run exists and is not completed; false otherwise
+   */
+  public synchronized boolean isRunning(String runId) {
+    RunDetails details = runDetailsMap.get(runId);
+    return details != null && details.getFuture() != null && !details.isDone();
```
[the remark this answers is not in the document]
supertick: removed

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -1,56 +1,132 @@
...
+        } catch (RunnerManagerException e) {
+          throw new RuntimeException(e);
+        }
+      });
+      futures.add(future);
+      runIds.add(runId);
+    }
+    
+    // Ensure all futures have started
+    CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).join();
```
rseitz, line 108: this is a way of blocking until all completeable futures are complete

it is not a good way to simply check that they have started, because it also waits for them to finish

it makes the while loop below completely redundant and unnecessary, but what it does not do is make sure we we stay below the 10 second threshold. You could change the join to get(10, TimeUnit.SECONDS) and let a TimeoutException bubble up and fail the test. However, then you would not be testing the behavior of runnermanager.getRunDetails() and details.isDone() which is an important part of this test

why was this line added?

supertick: Because your previous test had a race condition... 
the call to start -> runnerManager.runWithConfig(runId, configId);  is asynchronous,
and you immediately spin through runnerManager.getRunDetails(x); to check details when they aren't even available yet.  This caused a tricky null pointer which was confusing to diagnose, because it worked locally in debug where it was slower but failed with an npe getting RunDetails in ci/cd. A small 10ms wait for the test's requests to launch would fix it .. but i thought the wait seemed crude.

rseitz: OK, in my comment i was mixing up the futures started in the test code with the futures started by RunnerManager. resolving...

PR #236, opened 2025-01-09
LC-605 FileConnector needs to continue if it hits an error on a file.

kwatters:
file connector doesn't currently work with a windows file path specified that includes the drive letter

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
@@ -10,29 +10,34 @@ import com.kmwllc.lucille.core.ConnectorException;
 import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.Publisher;
 import com.typesafe.config.Config;
+
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
+import java.nio.file.FileVisitResult;
+import java.nio.file.FileVisitor;
 import java.nio.file.Files;
-import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
-import java.util.stream.Stream;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class LocalStorageClient extends BaseStorageClient {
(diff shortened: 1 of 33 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/StorageClient.java
```diff
@@ -48,28 +48,28 @@ public interface StorageClient {
    * Gets the appropriate client based on the URI scheme and validate with authentication/settings from cloudOptions
    */
   static StorageClient create(URI pathToStorage, String docIdPrefix, List<Pattern> excludes, List<Pattern> includes,
-      Map<String, Object> cloudOptions, Config fileOptions) {
+      Map<String, Object> cloudOptions, Config fileOptions, Publisher publisher) {
```
rseitz, line 51: Is this change necessary for resolving the performance issue that this PR is addressing?
There was an explicit decision during the design & review of the FileConnector framework that StorageClient should not bound to any one Publisher. A StorageClient knows about a storage backend and lets you access files from that backend. Once you have a StorageClient you can tell it to publish files with a given publisher but you shouldn't need a publisher in order to create a StorageClient. If there's a concern here, can we address this in a separate ticket/PR?

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -74,19 +82,10 @@ public void shutdown() throws IOException {
...
-            tryProcessAndPublishFile(publisher, fullPathStr, fileExtension, new FileReference(fullPath));
-          });
-    } catch (InvalidPathException e) {
-      throw new ConnectorException("Path string provided cannot be converted to a Path.", e);
-    } catch (SecurityException | IOException e) {
-      throw new ConnectorException("Error while traversing file system.", e);
-    }
+    
+    // LocalStorage class implements the FileVisitor so we can just walk it.
+    Files.walkFileTree(startingDirectoryPath, this);
```
kwatters, line 87: the interface here needs to pass a FileVisitor, so when a file is visited, it's getting the LocalStorageClient which implements the visitFile method.  The result is that LocalStorageClient needs a handle to the publisher.  Otherwise we would need to make the publisher implement FileVisitor.

rseitz: 1) The change here removes `paths.filter(this::isValidPath)`. That test is how we support includes and excludes. If that is removed without being replaced elsewhere, tests should fail, otherwise we have a gap in test coverage.

2) We could have a wrapper around Publisher that implements FileVisitor. I'm not convinced we need to store the Publisher or modify Publisher itself.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/StorageClient.java
rseitz: Here is a way to implement the essential changes in this PR without adding Publisher everywhere
https://github.com/kmwtechnology/lucille/pull/244

PR #239, opened 2025-01-13
LC-483: Support a replacement field in ReplacePatterns

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/stage/ReplacePatterns.java
```diff
@@ -52,14 +56,15 @@ public class ReplacePatterns extends Stage {
 
   public ReplacePatterns(Config config) {
     super(config, new StageSpec().withRequiredProperties("source", "dest", "regex", "replacement")
```
rseitz, line 58: should the user be required to specify "replacement" if they're using "replacement_field"?

how about this behavior -- if the user has specified "replacement_field" in the config but a particular document doesn't have that field, then we check if "replacement" is specified. If "replacement" is specified, we use it, but if "replacement" isn't specified we skip this document and don't do any replacement.  

So the user must specify at least one of "replacement" and "replacement_field" but when they're providing "replacement_field" they have the option of omitting "replacement"

let me know what you think

mrwrench24: Wondering what you think about the case where...
- A `replacement_field` is specified
- It is present in a document, but it is mapped to `null`.

Should we skip matching on the document or try to fall back on the default `replacement` String?

rseitz: I'm thinking if replacement_field is null, we try to fall back on the replacement string as you said, but if replacement string itself isn't provided, then we skip

mrwrench24: New commits allow us to fall back on `replacement` in either case. + new tests.

lucille-core/src/main/java/com/kmwllc/lucille/stage/ReplacePatterns.java
```diff
@@ -22,7 +22,12 @@
  * <br>
  * - regex (List&lt;String&gt;) : A list regex expression to find matches for. Matches will be extracted and placed in the destination fields.
  * <br>
- * - replacement (String) : The String to replace regex matches with.
+ * - replacement (String, Optional) : The String to replace regex matches with. If null, pattern replacement will only take place if
+ * a replacement_field is specified and found in a document.
+ * <br>
+ * - replacement_field (String, Optional): Specify a field in the document that maps to a String. If non-null, replacements of a
+ * pattern within a document will use the string mapped to the replacement_field, if present. Otherwise, the default
+ * replacement String is used (if it is not null).
```
rseitz, line 30: minor: please be explicit about what happens if we try to use the default replacement string but it IS null

lucille-core/src/main/java/com/kmwllc/lucille/stage/ReplacePatterns.java
```diff
@@ -51,15 +57,16 @@ public class ReplacePatterns extends Stage {
...
+    super(config, new StageSpec().withRequiredProperties("source", "dest", "regex")
+        .withOptionalProperties("replacement", "replacement_field", "update_mode", "ignore_case", "multiline", "dotall", "literal"));
 
     this.patterns = new ArrayList<>();
 
     this.sourceFields = config.getStringList("source");
     this.destFields = config.getStringList("dest");
     this.regexExprs = config.getStringList("regex");
-    this.replacement = config.getString("replacement");
+    this.replacement = ConfigUtils.getOrDefault(config, "replacement", null);
```
rseitz, line 68: let's require that the user provide at least one of replacement or replacement_field

rseitz (approved):

PR #240, opened 2025-01-16
LC-608: Run ID printed in logs

mrwrench24:
At first I wasn't sure if this was enough to keep the run_id specific to each run / not overwrite it for others. But then I started two runs with the Admin API (as I was set up to test it) and the ids logged were correct.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
@@ -187,11 +187,12 @@ public abstract class Indexer implements Runnable {
 
     if (doc == null) {
       sendToIndexWithAccounting(batch.flushIfExpired());
-    } else {
-      sendToIndexWithAccounting(batch.add(doc));
+      MDC.remove(Document.RUNID_FIELD);
+      return;
     }
 
-    MDC.clear();
+    sendToIndexWithAccounting(batch.add(doc));
+    MDC.remove(Document.RUNID_FIELD);
   }
 
   private void sendToIndexWithAccounting(List<Document> batchedDocs) {
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
@@ -53,12 +53,15 @@ public abstract class Indexer implements Runnable {
 
   private Instant lastLog = Instant.now();
 
+  // A runID for a local (local / test) run. Null if not in one of those modes / started independently.
+  private final String localRunId;
+
   public void terminate() {
     running = false;
     log.debug("terminate");
   }
 
-  public Indexer(Config config, IndexerMessenger messenger, String metricsPrefix) {
+  public Indexer(Config config, IndexerMessenger messenger, String metricsPrefix, String localRunId) {
     this.messenger = messenger;
     this.idOverrideField =
         config.hasPath("indexer.idOverrideField")
@@ -117,6 +120,8 @@ public abstract class Indexer implements Runnable {
     this.stopWatch = new StopWatch();
     this.meter = metrics.meter(metricsPrefix + ".indexer.docsIndexed");
     this.histogram = metrics.histogram(metricsPrefix + ".indexer.batchTimeOverSize");
+
+    this.localRunId = localRunId;
   }
 
   /**
@@ -148,12 +153,14 @@ public abstract class Indexer implements Runnable {
 
   @Override
   public void run() {
+    // might be null b/c in a non-local mode, and that's okay!
+    MDC.put(RUNID_FIELD, localRunId);
+
     try {
       while (running) {
         checkForDoc();
       }
       sendToIndexWithAccounting(batch.flush()); // handle final batch
-      MDC.remove(RUNID_FIELD);
     } finally {
       close();
     }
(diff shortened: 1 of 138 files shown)


mrwrench24:
Not ready yet

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -173,6 +175,10 @@ private void checkForDoc() {
       // blocking poll with a timeout which we assume to be in the range of
       // several milliseconds to several seconds
       doc = messenger.pollDocToIndex();
+
+      if (doc != null) {
+        MDC.put(Document.RUNID_FIELD, doc.getRunId());
+      }
```
mrwrench24, line 181: I wasn't sure if I should set the RunID to NULL here if a doc is null. It seemed like we were missing out on some logged RunIDs when I did so.

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -467,7 +470,13 @@ private static ConnectorResult runConnectorWithComponents(Config config,
       String metricsPrefix = runId + "." + connector.getName() + "." + connector.getPipelineName();
 
       if (startWorkerAndIndexer && connector.getPipelineName() != null) {
-        workerPool = new WorkerPool(config, pipelineName, workerMessengerFactory, metricsPrefix);
+        // If local/test we want to give WorkerPool the run_id directly.
+        // (Otherwise let Kafka messengers pass it thru in the documents.)
+        if (runType.equals(RunType.LOCAL) || runType.equals(RunType.TEST)) {
```
mrwrench24, line 475: Not sure if Kafka local should be included here.

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
rseitz: Even if `doc` is null here, this method may still send documents to the index. That happens if we've built up a non-empty batch of documents that "expires" (i.e. times out) so we flush it. See       `sendToIndexWithAccounting(batch.flushIfExpired());` down below.

The complication here though is that in distributed mode, someone could use the Runner multiple times to launch several ingests each with their own run ID. Documents from multiple runs could get picked up by the same Indexer and be included in the same batch. So we don't currently have a guarantee that the documents inside any given Batch will all have the same RunID. We'll need to discuss in the larger group whether that's something we should be enforcing. But for now it means we need to be extra careful about how we're setting the run ID in the MDC and how long we're leaving it set... because we ever want a doc from one run to be logged with a runID from a different concurrent run...

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
rseitz: Good question -- ideally we wouldn't include it here. KafkaLocal could then be a good/easy way for us to test whether the run ID is properly handled when we're using Kafka.

lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
```diff
@@ -93,6 +94,8 @@ public PublisherImpl(Config config, PublisherMessenger messenger, String runId,
 
   @Override
   public void publish(Document document) throws Exception {
+    MDC.put(Document.RUNID_FIELD, runId);
```
mrwrench24, line 97: This could be causing issues.

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -412,7 +416,7 @@ private static ConnectorResult runConnectorInternal(Config config, String runId,
       }
     } else {
       try {
-        ConnectorThread connectorThread = new ConnectorThread(connector, publisher, ThreadNameUtils.createName("Connector"));
+        ConnectorThread connectorThread = new ConnectorThread(connector, publisher, runId, ThreadNameUtils.createName("Connector"));
```
rseitz, line 419: do we want to include runID in the thread name? can address in a separate PR

lucille-core/src/main/java/com/kmwllc/lucille/message/KafkaWorkerMessenger.java
```diff
@@ -48,6 +51,7 @@ public Document pollDocToProcess() throws Exception {
     if (consumerRecords.count() > 0) {
       ConsumerRecord<String, KafkaDocument> record = consumerRecords.iterator().next();
       KafkaDocument doc = record.value();
+      MDC.put(RUNID_FIELD, doc.getRunId());
```
rseitz, line 54: can remove

lucille-core/src/main/java/com/kmwllc/lucille/message/HybridWorkerMessenger.java
```diff
@@ -68,6 +71,8 @@ public KafkaDocument pollDocToProcess() throws Exception {
     if (consumerRecords.count() > 0) {
       ConsumerRecord<String, KafkaDocument> record = consumerRecords.iterator().next();
       KafkaDocument doc = record.value();
+
+      MDC.put(RUNID_FIELD, doc.getRunId());
```
rseitz, line 75: can remove

PR #241, opened 2025-01-17
LC-613: Independent run metrics

mrwrench24:


lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
```diff
@@ -755,32 +755,30 @@ public void testIndexerDeleteMarkerConfig() throws Exception {
   public void testMetrics() throws Exception {
     SharedMetricRegistries.clear();
     assertEquals(0, SharedMetricRegistries.names().size());
-    Runner.runInTestMode("RunnerTest/twoConnectors.conf");
+    Map<String, TestMessenger> runnerMessengerMap = Runner.runInTestMode("RunnerTest/twoConnectors.conf");
+    String connector1ID = runnerMessengerMap.get("connector1").getRunId();
```
rseitz, line 759: please rename connector1ID to run1ID. What you are getting here is the runID for the first run we're executing in this test. We don't really have a notion of IDs for connectors (though they can have names specified in the config).

you can add a junit assertion that run1ID is the same as what you are currently calling connector2ID below, and you don't need to store that value separately

after you do the second run (i.e. second call to Runner.runInTestMode) down below, please repeat this logic -- capture the new runId -- call it run2ID -- and make sure that we can access the timers beginning with that second runId -- in the code that looks like  metrics.getTimers(MetricFilter.contains...

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
```diff
@@ -755,32 +755,30 @@ public void testIndexerDeleteMarkerConfig() throws Exception {
...
-    // same JVM;
-    // if we re-execute the run from above the counts will not be reset but will increase from their earlier values;
-    // we may wish to prevent this in the future by including the runId in the metrics naming scheme
-    // or by updating Runner to call SharedMetricRegistries.clear() before each run, assuming runs are sequential
+    // metrics for the same connector/pipeline pairs are independent across multiple runs in the same JVM.
     Runner.runInTestMode("RunnerTest/twoConnectors.conf");
-    assertEquals(2, connector1Pipeline1Stage1Timers.get(connector1Pipeline1Stage1Timers.firstKey()).getCount());
-    assertEquals(2, connector2Pipeline2Stage1Timers.get(connector2Pipeline2Stage1Timers.firstKey()).getCount());
+    assertEquals(1, connector1Pipeline1Stage1Timers.get(connector1Pipeline1Stage1Timers.firstKey()).getCount());
+    assertEquals(1, connector2Pipeline2Stage1Timers.get(connector2Pipeline2Stage1Timers.firstKey()).getCount());
```
rseitz, line 781: the variables connector1Pipeline1Stage1Timers and connector2Pipeline2Stage1Timers should be renamed to run1connector1Pipeline1Stage1Timers and run1connector2Pipeline2Stage1Timers

the current assertions should be kept in place -- this is showing that the metrics for run1 are unchanged by the execution of run2

but we also need to capture run2connector1Pipeline1Stage1Timers and run2connector2Pipeline2Stage1Timers and assert that those are set correctly

rseitz (approved):

PR #242, opened 2025-01-17
LC-481: Small improvements for Stage unit tests

mrwrench24:
The tests primarily ensure that exceptions are thrown when Stages are built with an invalid configuration.


lucille-core/src/test/resources/MatchQueryTest/blank_matched_queries.conf
```diff
@@ -0,0 +1,21 @@
+# This is a configuration file managed by https://github.com/lightbend/config
```
rseitz, line 1: please omit these two boilerplate comment lines here and in the other conf files for MatchQueryTest. (We have lots of .conf files in the codebase and it doesn't make sense include this comment everywhere.)

rseitz (approved):

PR #243, opened 2025-01-21
LC-426: Improved Unit Tests for CSV/Open/Elasticsearch Indexers

mrwrench24:
ElasticsearchIndexer: 66% / 57% → 99% / 87%
OpensearchIndexer: 74% / 64% → 92% / 81%
CSVIndexer: 72% / 54% → 92% / 95%

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/indexer/OpenSearchIndexerTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/indexer/OpenSearchIndexerTest.java
@@ -801,6 +801,7 @@ public class OpenSearchIndexerTest {
 
   @Test
   public void testInvalidConfig() {
+    // cannot use the indexOverrideField
     TestMessenger messenger = new TestMessenger();
     Config config = ConfigFactory.load("OpenSearchIndexerTest/invalidConfig.conf");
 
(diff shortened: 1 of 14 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/indexer/OpenSearchIndexerTest.java
```diff
@@ -73,6 +80,13 @@ private void setupOpenSearchClient() throws IOException {
     DeleteByQueryResponse mockDeleteByQueryResponse = Mockito.mock(DeleteByQueryResponse.class);
     when(mockClient.deleteByQuery(any(DeleteByQueryRequest.class))).thenReturn(mockDeleteByQueryResponse);
     when(mockDeleteByQueryResponse.failures()).thenReturn(new ArrayList<>());
+
+    mockFailPingClient = Mockito.mock(OpenSearchClient.class);
```
rseitz, line 84: Since we're only using this in one test -- testOpenSearchIndexerException() -- we can set up the mock in that test method. It doesn't need to be instance variable of OpenSearchIndexerTest that's initialized in setup() before each test method -- that would make more sense if this resource was used in most or all of the tests.

PR #244, opened 2025-01-21
LC-605 file connector changes v2

rseitz:
Reworked proposal for LC-605 PR.


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -148,10 +124,13 @@ private Document pathToDoc(Path path) throws ConnectorException {
 
       // setting fields on document
       doc.setField(FILE_PATH, fullPath);
-      if (attrs.lastModifiedTime() != null) doc.setField(MODIFIED, attrs.lastModifiedTime().toInstant());
-      if (attrs.creationTime() != null) doc.setField(CREATED, attrs.creationTime().toInstant());
+      if (attrs.lastModifiedTime() != null)
+        doc.setField(MODIFIED, attrs.lastModifiedTime().toInstant());
+      if (attrs.creationTime() != null)
+        doc.setField(CREATED, attrs.creationTime().toInstant());
       doc.setField(SIZE, attrs.size());
```
mrwrench24, line 131: Def nitpicky but might want to move this call to `setField` above / space it out so it's clear it's not part of the conditionals above (or intended to be)

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -93,9 +93,12 @@ public BaseStorageClient(URI pathToStorageURI, String docIdPrefix, List<Pattern>
    */
   protected void tryProcessAndPublishFile(Publisher publisher, String fullPathStr, String fileExtension, FileReference fileReference) {
```
mrwrench24, line 94: Also nitpicks but looks like we can remove the import of `File` now :)

mrwrench24 (commented):
Besides two small style notes - LGTM

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
rseitz: Thx. Looks like this code started out as some one-line conditionals without curly braces. Our convention is to always use curly braces. This is the reason here -- a later refactor put each conditional on multiple lines, and without the braces they're now hard to distinguish. Will fix.

mrwrench24 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -41,52 +43,34 @@ public LocalStorageClient(URI pathToStorageURI, String docIdPrefix, List<Pattern
   @Override
   public void init() throws ConnectorException {
     try {
-      fs = FileSystems.getDefault();
-      // get current working directory path
-      startingDirectoryPath = fs.getPath(startingDirectory);
+      File f = new File(startingDirectory);
```
rseitz, line 46: @kwatters This change is from your original PR. Could you help me understand what it is doing?

startingDirectory is set to the return value of getStartingDirectory() which is going to return (new File(pathToStorageURI.getPath())).getPath();

There's a conditional added here where we check if (new File(startingDirectory)).exists()
What is reasoning behind these two cases?

rseitz: What I'm seeing is that `BaseStorageClient`'s constructor takes `URI pathToStorageURI` and stores it in an instance variable. Then startingDirectory gets set:  `this.startingDirectory = getStartingDirectory();`

Looking at the base implementation of getStartingDirectory() we see it strips a leading slash:

```
    /**
     * default implementation to obtain starting directory
     */
    protected String getStartingDirectory() {
      String startingDirectory = Objects.equals(pathToStorageURI.getPath(), "/") ? "" : pathToStorageURI.getPath();
      if (startingDirectory.startsWith("/")) return startingDirectory.substring(1);
      return startingDirectory;
    }
```

So I guess my questions are:

1) why do we need to override getStartingDirectory() here in LocalStorageClient, why can't we just use the base implementation?
2) startingDirectoryPath is only used once in the traverse() method. Looks like we need an actual Path object but we have a URI (pathToStorageURI) and a String (startingDirectory). Why not just do Paths.get(startingDirectory) when we need that path, right inside traverse()?

mrwrench24 (approved):
I did a little bit of testing and things look good to me here.

PR #247, opened 2025-01-23
LC-424: Created tests for OpenSearchUtils

mrwrench24:
This required a refactor to `OpenSearchUtils` (namely `getOpenSearchRestClient`). A lot of code was in a callback function that didn't need to be in there, so now we can test it and get the line coverage :).

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
@@ -92,14 +92,14 @@ public class OpenSearchUtils {
         throw new RuntimeException(e);
       }
 
-      final var connectionManager = PoolingAsyncClientConnectionManagerBuilder.create()
-          .setTlsStrategy(tlsStrategy)
-          .build();
-
       final var transport = ApacheHttpClient5TransportBuilder
           .builder(hosts)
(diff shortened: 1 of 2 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/util/OpenSearchUtils.java
```diff
@@ -37,63 +37,61 @@ public class OpenSearchUtils {
...
 
-
-      // code for Apache Client is taken from following link:
+      // code for building an Apache Client is inspired by the following link:
       // https://github.com/opensearch-project/opensearch-java/blob/main/samples/src/main/java/org/opensearch/client/samples/SampleClient.java
       // When comparing to example code, here are differences:
       //  - We gather data from our config rather than providing directly
       //  - We disable TLS/SSL verification only if acceptInvalidCerts is true (from config)
+
+      final BasicCredentialsProvider credentialsProvider = new BasicCredentialsProvider();
```
rseitz, line 54: Looks like a good (and straightforward) refactor to improve testability of this code. Since this is an error prone area of the codebase though -- even though we have unit tests that use mocking -- it would be ideal to also do a manual test to make sure we can still connect to a live opensearch instance... let's discuss.

rseitz (approved):

PR #248, opened 2025-01-24
LC-618: Special path separator for archive files

mrwrench24:
Now we have a different character for when we are handling zip / tar files. This just was a small change in `BaseStorageClient` as well as a minor update for `CSVFileHandler` because of how it was assigning IDs to documents.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
@@ -64,6 +64,7 @@ public class FileConnector extends AbstractConnector {
   public static final String SIZE = "file_size_bytes";
   public static final String CONTENT = "file_content";
   public static final String FILE_SEPARATOR = ":";
+  public static final String ARCHIVE_FILE_SEPARATOR = "!";
 
   // cloudOption Keys
   public static final String AZURE_CONNECTION_STRING = "connectionString";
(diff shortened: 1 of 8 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -64,6 +64,7 @@ public class FileConnector extends AbstractConnector {
   public static final String SIZE = "file_size_bytes";
   public static final String CONTENT = "file_content";
   public static final String FILE_SEPARATOR = ":";
```
rseitz, line 66: Do we still need FILE_SEPARATOR at all? I see only one reference -- in BaseStorageClient line 113. Wondering if that usage should stay as is or be changed?

mrwrench24: I had been wondering about this - the question is whether we want a different separator for non-archive compressed files. (A single `.gz` but non-`tar` file.)

mrwrench24: This could also be a symptom of us having not been nice to Windows - since the file(path) separator is `;`, so maybe we should replace anyways.

mrwrench24: I might be in a complete rabbit hole right now, but reading about https://www.komprise.com/glossary_terms/alternate-data-streams-ads/#:~:text=ADS%20is%20a%20feature%20that,txt”. - alternate data streams in NTFS...

rseitz (approved):

PR #249, opened 2025-01-28
LC-426: Fix Opensearch Indexer Test

mrwrench24:


rseitz:
When doing a quick followup to fix a previous PR, we can use the same ticket number.

rseitz (approved):

PR #250, opened 2025-01-30
LC-611: Unit Testing for Concurrent, Non-Trivial Runs

mrwrench24:
Each run takes ~1.25 seconds (on my machine). I can make them take longer if needed (primarily by adding more data, I guess…). They read from a 16 line CSV file and do some operations. There are 3 runs in total, each uses similar stages but with different configurations, and different fields actually get indexed into the output CSV. Then I run some validations on each output file to make sure everything went smoothly. I tried to design the configs so that, although they use similar stages, the output is very specific to that run - for example, some of the fields that get indexed for the first configuration are not in the second and third configuration. The ReplacePatterns of the second configuration is reliant on its previous RenameFields output, which none of the other configs do the same.


rseitz (approved):
approving with minor comment

PR #251, opened 2025-02-04
LC-609: Doc IDs in Logs + new DocLogger

mrwrench24:
Adds DocIDs to the MDC for logging. Also adds some new log statements about document processing that are sent to the `DocLogger` - a separate logger which, when in use, will be very verbose. (Level is set to "OFF" to disable.)


lucille-core/src/main/resources/log4j2.xml
```diff
@@ -2,17 +2,17 @@
 <Configuration>
     <Appenders>
         <Console name="console" target="SYSTEM_OUT">
-            <PatternLayout pattern="%d{yy/MM/dd HH:mm:ss} %p %c{1}: %m%n" />
+            <PatternLayout pattern="%d{yy/MM/dd HH:mm:ss} %X{id} %p %c{1}: %m%n" />
```
rseitz, line 5: looks like we're putting the doc ID at the same place in the patternlayout where we put the Run ID in the LC-608 PR?  Suggest combining LC-608 and LC-609 into one unified PR so that we can be sure both sets of changes work together...

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -213,12 +218,13 @@ private void sendToIndexWithAccounting(List<Document> batchedDocs) {
       log.error("Error sending documents to index: " + e.getMessage(), e);
 
       for (Document d : batchedDocs) {
-        try {
+        try (MDCCloseable docIdMDC = MDC.putCloseable(ID_FIELD, d.getId())) {
```
rseitz, line 221: can we also set the RunID here?

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -232,12 +238,13 @@ private void sendToIndexWithAccounting(List<Document> batchedDocs) {
     }
 
     for (Document d : batchedDocs) {
-      try {
+      try (MDCCloseable docIdMDC = MDC.putCloseable(ID_FIELD, d.getId())) {
```
rseitz, line 241: can we also set the RunID here?

PR #254, opened 2025-02-07
LC-502: Add retry status conditions for FetchUri stage

kiratraynor:
Adds a statusCodeRetryList to the FetchUri config to allow retries for requests that return specific status codes (including wild cards).
- does not allow "200", "20x", or "2xx" (not case specific)
- logs warning if any codes in 200-299 are included, but continues
- logs a warning if you've included an invalid number or wild card and removes it from the list
- logs warning if you have added codes to the retry list, but have the max number of retries set to 0


lucille-core/src/main/java/com/kmwllc/lucille/core/StatusCodeResponseInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @Override
+  public void process(HttpResponse httpResponse, HttpContext httpContext) throws IOException {
+    HttpClientContext clientContext = HttpClientContext.adapt(httpContext);
+    int statusCode = httpResponse.getStatusLine().getStatusCode();
+    numRetry++;
+    if (listContainsValue(statusCode)) {
+      log.error("Got status {} after {} retries", statusCode, numRetry - 1);
+      throw new IOException("Retry request to '" + clientContext.getTargetHost() + clientContext.getRequest().getRequestLine().getUri() + "'");
```
rseitz, line 34: in the exception message string, should there be any space or separator between the values of getTargetHost() and getUri() or is it deliberate to not separate them?

kiratraynor: This was deliberate so that the full url is printed.

lucille-core/src/main/java/com/kmwllc/lucille/core/StatusCodeResponseInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+
+  // check if the status code retry lists contains the code itself or a wildcard that contains includes that code.
+  // ex. if code=500 and statusCodeRetryList=["429","5xx"], then this would return true.
+  protected boolean listContainsValue(int code) {
```
rseitz, line 52: if we need to make this method visible for testing, I suggest making it package-private instead of protected (i.e. just leave off the access modifier). Protected suggests we might want to someday subclass this class and override this method, but I don't think that's the case here.

lucille-core/src/main/java/com/kmwllc/lucille/core/StatusCodeResponseInterceptor.java
```diff
@@ -0,0 +1,60 @@
...
+  private int numRetry = 0;
+  private final List<String> statusCodeList;
+  private final List<String> statusCodeWildcardList;
+
+  public StatusCodeResponseInterceptor() {
+    this.statusCodeList = new ArrayList<>();
+    this.statusCodeWildcardList = new ArrayList<>();
+  }
+
+  public StatusCodeResponseInterceptor(List<String> statusCodeRetryList) {
```
rseitz, line 25: wondering if we should have a constructor that takes two lists -- statusCodeList and statusCodeWilcardList
the FetchUriStage would create those two lists once and only once, and pass them in to the constructor every time it creates a new StatusCodeResponseInterceptor, which is every time fetch a new URI. Would save repeated/redudnant preprocessing of the lists for every URI

kiratraynor: Yes, I can add this logic to the start method in FetchUri.

rseitz (approved):

PR #255, opened 2025-02-11
LC-608 + LC-609: Doc + Run ID in Logs

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -232,12 +252,13 @@ private void sendToIndexWithAccounting(List<Document> batchedDocs) {
     }
 
     for (Document d : batchedDocs) {
-      try {
+      try (MDCCloseable docIdMDC = MDC.putCloseable(ID_FIELD, d.getId())) {
```
rseitz, line 255: set runId here along with docId?

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -213,12 +232,13 @@ private void sendToIndexWithAccounting(List<Document> batchedDocs) {
       log.error("Error sending documents to index: " + e.getMessage(), e);
 
       for (Document d : batchedDocs) {
-        try {
+        try (MDCCloseable docIdMDC = MDC.putCloseable(ID_FIELD, d.getId())) {
```
rseitz, line 235: set runId here along with docId?

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -173,6 +186,13 @@ private void checkForDoc() {
       // blocking poll with a timeout which we assume to be in the range of
       // several milliseconds to several seconds
       doc = messenger.pollDocToIndex();
+
+      // continuously update the MDC if we haven't been given a localRunID (in Kafka modes).
+      // batch logging IDs won't be perfect because we don't enforce that docs in a batch come from the same
+      // run. We don't clear the MDC so that the final batch flush can have an associated run ID.
+      if (localRunId == null && doc != null) {
+        MDC.put(RUNID_FIELD, doc.getRunId());
```
rseitz, line 194: this is where we discussed setting a placeholder / UNKNOWN runId since the codepath in sendToIndexWithAccounting (where this MDC value will still be set) could involve docs from different runs

mrwrench24: Makes sense. I am trying to find an elegant way to include both this placeholder as well as a run ID in the try-with-resources blocks where we send events for documents as well. MDC has a deque where you can push/pop keys, which seemed to be an ideal solution, but it doesn't appear to be used very commonly and was preventing the JVM from closing.

rseitz (approved):

PR #256, opened 2025-02-12
LC-628: FileConnector streams Cloud Files. FileHandler only uses / passes through InputStreams.

mrwrench24:
When processing cloud files (into a file handler - **not** in stages), the FileConnector loaded the cloud file's content into a `byte[]` in memory. Now, the contents are streamed, so all of the content does not have to be loaded into memory at once in order to use FileConnector / FileHandler(s).

`FileHandler` no longer has a `processFile`/`processFileAndPublish` method that takes in `byte[]` or a `Path`, instead, it only accepts `InputStreams` to their respective file types.


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -138,10 +138,10 @@ protected void tryProcessAndPublishFile(Publisher publisher, String fullPathStr,
       if (!fileOptions.isEmpty() && FileHandler.supportAndContainFileType(fileExtension, fileOptions)) {
 
         if (fileReference.isCloudFileReference()) {
-          // get the file content
-          byte[] content = getFileReferenceContent(fileReference);
+          // Get a stream for the file content, so we don't have to load it all at once.
+          InputStream contentStream = getFileReferenceContentStream(fileReference);
           // get the right FileHandler and publish based on content
-          publishUsingFileHandler(publisher, fileExtension, content, fullPathStr);
+          publishUsingFileHandler(publisher, fileExtension, contentStream, fullPathStr);
```
mrwrench24, line 144: I didn't want to overdo it in this PR alone, but part of me was wondering if we can just remove the `Path` function call from the interface altogether, and just use `InputStream` from now on (including for local files).

lucille-core/src/main/java/com/kmwllc/lucille/connector/JSONConnector.java
```diff
@@ -32,8 +34,9 @@ public void execute(Publisher publisher) throws ConnectorException {
     }
 
     try {
+      InputStream stream = FileUtils.getLocalInputStream(pathStr);
       log.debug("Processing file: {}", path);
-      jsonFileHandler.processFileAndPublish(publisher, path);
+      jsonFileHandler.processFileAndPublish(publisher, stream, path.toString());
```
rseitz, line 39: do we need path here at all, given that we have pathStr?

lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/CSVFileHandler.java
```diff
@@ -242,20 +228,9 @@ private String[] getHeaderFromCSV(CSVReader csvReader) throws FileHandlerExcepti
...
-    } catch (IOException e) {
-      throw new FileHandlerException("Error creating CSVReader for file " + FilenameUtils.getName(pathStr), e);
-    }
-  }
-
-  private CSVReader getCsvReader(byte[] fileContent) throws FileHandlerException {
+  private CSVReader getCsvReader(InputStream inputStream) throws FileHandlerException {
     try {
-      return new CSVReaderBuilder(new InputStreamReader(new ByteArrayInputStream(fileContent), StandardCharsets.UTF_8))
+      return new CSVReaderBuilder(new InputStreamReader(inputStream, StandardCharsets.UTF_8))
```
rseitz, line 233: confirm that CSVReaderBuilder add it's own buffering (if necessary) and we shouldn't wrap InputStreamReader in a BufferedReader

rseitz (approved):
approving with some comments to follow up on
please wait to merge until LC-534 is approved and can be merged immediately after

lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/CSVFileHandler.java
mrwrench24: I added BufferedReader back. I wasn't able to get a direct answer about whether the CSVReader is / isn't buffered internally. But we were using a buffered reader for paths before, and many examples online are passing a BufferedReader through, so it seems appropriate.

PR #258, opened 2025-02-18
LC-534: Removed VFSInputStream + framework for fetching individual InputStreams for files.

mrwrench24:
1. Removed `VFSInputStream` from `FileUtils`. This was used to handle URIs passed to the `getReader` method.
2. Created `FileContentFetcher`. This is an object that replaced a lot of the old FileUtils method. Objects that need to repeatedly (or, perhaps, just once) get the contents (via an `InputStream`) of a path String will use this object (think `TextExtractor` as a good example). The `Config` should be propagated through to the `FileContentFetcher` to create the appropriate `StorageClients`. It will use entries for `azure`, `s3`, and `gcp`, if present, to create those `StorageClients`.
3. Updated some previous code that used `FileUtils` to use the `FileContentFetcher`. This includes `TextExtractor` and `ExtractEntities`, which make repeated calls, as well as `DictionaryLookup` and `KafkaUtils`, which will not. When users want to store/get contents of cloud files, they need to specify options (`azure`, `s3`, `gcp`) in their Config, similar to `FileConnector`.
4. `StorageClients` are now more reusable / lightweight to initialize. They previously took in many arguments at construction to use later on during a single traversal. Now, they only take in a `Config` (to check for cloud storage provider options), or in the case of `LocalStorageClient`, nothing. These arguments are now handled in `TraversalParams` which are part of a call to `traverse`. 
5. Now, `AzureStorageClient` internally uses a different client - one that can be started up / persist rather than being dependent on a specific bucket or container name. So, multiple calls to `getFileContentStream` can be made on the same `StorageClient`.
6. `FileConnector` has been updated to use the individual entries for `azure`, `s3`, and `gcp` instead of specifying one `cloudOptions` map.
7. Now that `cloudOptions` are no longer used, `Config` is propagated through `FileConnector`/`StorageClient` instead of a raw `Map`, like before.
8. `Config` for `StorageClients` are now validated in their constructors. 
9. Some code for `StorageClients` has been moved to `BaseStorageClient`, which do some pre/post operations and call a `protected` method on the specific implementation to do the actual traversal / file fetching.


lucille-core/src/main/java/com/kmwllc/lucille/message/KafkaUtils.java
```diff
@@ -39,20 +41,31 @@ public class KafkaUtils {
...
 
-  private static Properties loadExternalProps(String filename) {
-    try (Reader propertiesReader = FileUtils.getReader(filename, StandardCharsets.UTF_8.name())) {
-      Properties consumerProps = new Properties();
-      consumerProps.load(propertiesReader);
-      return consumerProps;
-    } catch (IOException e) {
-      throw new IllegalStateException(String.format("Cannot load kafka property file %s.", filename));
+  private static Properties loadExternalProps(String filename, Map<String, Object> cloudOptions) {
+    FileContentFetcher fetcher = new FileContentFetcher(cloudOptions);
```
rseitz, line 45: for use cases like this, we should investigate making a FileContentFetcher.getReader method that take cloudOptions as an arg and does the startup and shutdown logic internally...

lucille-core/src/main/java/com/kmwllc/lucille/util/FileContentFetcher.java
```diff
@@ -0,0 +1,141 @@
...
+   * if an error occurs while initializing any of the StorageClients.
+   */
+  public void startup() throws StageException {
+    for (StorageClient client : availableClients.values()) {
+      try {
+        client.init();
+      } catch (ConnectorException e) {
+        shutdown();
+
+        throw new StageException("Unable to initialize StorageClient.", e);
```
rseitz, line 54: not sure this should be a StageException as this method could be called outside a stage...

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -46,38 +36,15 @@ public abstract class BaseStorageClient implements StorageClient {
...
-    this.fileOptions = fileOptions;
     this.fileHandlers = new HashMap<>();
-    this.getFileContent = !fileOptions.hasPath(GET_FILE_CONTENT) || fileOptions.getBoolean(GET_FILE_CONTENT);
-    this.handleArchivedFiles = fileOptions.hasPath(HANDLE_ARCHIVED_FILES) && fileOptions.getBoolean(HANDLE_ARCHIVED_FILES);
-    this.handleCompressedFiles = fileOptions.hasPath(HANDLE_COMPRESSED_FILES) && fileOptions.getBoolean(HANDLE_COMPRESSED_FILES);
-    this.moveToAfterProcessing = fileOptions.hasPath(MOVE_TO_AFTER_PROCESSING) ? fileOptions.getString(MOVE_TO_AFTER_PROCESSING) : null;
-    this.moveToErrorFolder = fileOptions.hasPath(MOVE_TO_ERROR_FOLDER) ? fileOptions.getString(MOVE_TO_ERROR_FOLDER) : null;
+
+    // only matters for traversals
     this.maxNumOfPages = cloudOptions.containsKey(MAX_NUM_OF_PAGES) ? (Integer) cloudOptions.get(MAX_NUM_OF_PAGES) : 100;
```
mrwrench24, line 48: This could be moved to the params, but I also see it making sense that you set this for **a** `StorageClient` that you might reuse and aren't going to change on a case-by-case basis... your machine can only handle so much and that doesn't change.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/StorageClient.java
```diff
@@ -42,36 +36,80 @@ public interface StorageClient {
...
+   * <br> "accountName" : azure account name
+   * <br> "accountKey" : azure account key
+   * <br> Mapped to: <b>https</b>
+   * <br> If using S3StorageClient:
+   * <br> "accessKeyId" : s3 key id
+   * <br> "secretAccessKey" : secret access key
+   * <br> "region" : s3 storage region
+   * <br> Mapped to: <b>s3</b>
+   */
+  static Map<String, StorageClient> clientsFromCloudOptions(Map<String, Object> cloudOptions) {
```
rseitz, line 96: minor comment about method naming:
the signature of this method shows that it takes cloudOptions so it's not super-helpful to include "FromCloudOptions" in the method name also
suggest: createClients()

bigger question to think about -- we have methods like GoogleStorageClient.validOptions() -- I wonder if instead we should have GoogleStorageClient.getOptions() which would return a Map (or a Config) of only those options that pertain to Google... and then when we create individual clients we would only pass in the smaller set of options that actual pertain to the client we're creating...

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/TraversalParams.java
```diff
@@ -0,0 +1,95 @@
...
+    this.includes = includes;
+    this.excludes = excludes;
+    this.fileOptions = fileOptions;
+    this.getFileContent = !fileOptions.hasPath(GET_FILE_CONTENT) || fileOptions.getBoolean(GET_FILE_CONTENT);
+    this.handleArchivedFiles = fileOptions.hasPath(HANDLE_ARCHIVED_FILES) && fileOptions.getBoolean(HANDLE_ARCHIVED_FILES);
+    this.handleCompressedFiles = fileOptions.hasPath(HANDLE_COMPRESSED_FILES) && fileOptions.getBoolean(HANDLE_COMPRESSED_FILES);
+    this.moveToAfterProcessing = fileOptions.hasPath(MOVE_TO_AFTER_PROCESSING) ? fileOptions.getString(MOVE_TO_AFTER_PROCESSING) : null;
+    this.moveToErrorFolder = fileOptions.hasPath(MOVE_TO_ERROR_FOLDER) ? fileOptions.getString(MOVE_TO_ERROR_FOLDER) : null;
+
+    if (forAzure) {
```
rseitz, line 49: is there any way to move the special purpose Azure logic out of TraversalParams and into AzureStorageClient itself? In other words, could AzureStorageClient take the same bucketOrContainerName and startingDirectory as other clients, and do the necessary transformation of those values after accessing them from TraversalParams?

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -14,126 +15,121 @@
...
-    return pathToStorageURI.getPath().split("/")[1];
-  }
-
-  @Override
-  protected String getStartingDirectory() {
-    String path = pathToStorageURI.getPath();
-    // path is in the format /containerName/folder1/folder2/... so need to return folder1/folder2/...
-    String[] subPaths = path.split("/", 3);
-    return subPaths.length > 2 ? subPaths[2] : "";
+  public AzureStorageClient(Map<String, Object> cloudOptions) {
```
rseitz, line 34: look at replacing this Map with a Config

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/TraversalParams.java
```diff
@@ -0,0 +1,95 @@
...
+
+  // provided arguments
+  public final URI pathToStorageURI;
+  public final String docIdPrefix;
+  private final List<Pattern> excludes;
+  private final List<Pattern> includes;
+  public final Config fileOptions;
+
+  // derived arguments
+  public final boolean getFileContent;
```
rseitz, line 29: pls. create getters for all instance variables we need to access and make the variables themselves private

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/GoogleStorageClient.java
```diff
@@ -31,9 +31,8 @@ public class GoogleStorageClient extends BaseStorageClient {
   private static final Logger log = LoggerFactory.getLogger(GoogleStorageClient.class);
   private Storage storage;
 
-  public GoogleStorageClient(URI pathToStorage, String docIdPrefix, List<Pattern> excludes, List<Pattern> includes,
-      Map<String, Object> cloudOptions, Config fileOptions) {
-    super(pathToStorage, docIdPrefix, excludes, includes, cloudOptions, fileOptions);
+  public GoogleStorageClient(Map<String, Object> cloudOptions) {
```
rseitz, line 34: another place where we want to take a Config instead of a Map

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/StorageClient.java
mrwrench24: first part is addressed... second part will be dependent on looking into `Config` changes, probably

lucille-core/src/main/java/com/kmwllc/lucille/stage/util/DictionaryManager.java
```diff
@@ -97,15 +103,17 @@ public static synchronized Map<String, String[]> getDictionary(String path, bool
    * @param dictPath  the path of the dictionary file
    * @return the populated HashMap
    */
-  private static HashMap<String, String[]> buildHashMap(String dictPath, boolean ignoreCase, boolean setOnly) throws StageException {
+  private static HashMap<String, String[]> buildHashMap(String dictPath, boolean ignoreCase, boolean setOnly, Config cloudOptions) throws IOException {
```
rseitz, line 106: pls. add a comment mentioning that cloudOptions is not included in the key

lucille-core/src/main/java/com/kmwllc/lucille/util/FileContentFetcher.java
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * Initialize the storage clients associated with this FileContentFetcher. Calls shutdown on all clients and throws an Exception
+   * if an error occurs while initializing any of the StorageClients.
+   */
+  public void startup() throws IOException {
+    for (StorageClient client : availableClients.values()) {
+      try {
+        client.init();
+      } catch (ConnectorException e) {
```
rseitz, line 43: client.init() shouldn't throw ConnectorException -- let's look for another option

lucille-core/src/main/java/com/kmwllc/lucille/util/FileContentFetcher.java
```diff
@@ -0,0 +1,187 @@
...
+      }
+    }
+  }
+
+  /**
+   * Attempts to return an InputStream for the file at the given path. If the path begins with "classpath:" the prefix will be removed
+   * and the file will be read from the classpath. Otherwise, if the path is a URI, a corresponding storage client will be used;
+   * if not, the local file system will be used. Will not return a null InputStream - instead, an exception will be thrown.
+   */
+  public InputStream getInputStream(String path) throws IOException {
```
rseitz, line 70: need to validate that startup() has been called and shutdown() hasn't been called

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -65,108 +58,127 @@ public void shutdown() throws IOException {
...
+            });
           });
-        });
+    } finally {
+      clearFileHandlers();
+    }
+  }
+
+  @Override
+  public InputStream getFileContentStream(URI uri) throws IOException {
```
rseitz, line 89: should confirm that the storage client has been initialized and not closed/shutdown

lucille-core/src/main/java/com/kmwllc/lucille/util/FileContentFetcher.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  // -- Convenience / Static Methods --
+  /**
+   * Returns an InputStream for the file at the given path, which could be in the classpath, a local file path, or a URI
+   * to a supported Storage service. An exception may occur - in this case, the StorageClient(s) created will be shutdown.
+   * If an InputStream is returned successfully, its close() method will shutdown the StorageClient(s) this method temporarily
+   * creates. Be sure to close the returned stream!
+   */
+  public static InputStream getSingleInputStream(String path, Config cloudOptions) throws IOException {
```
rseitz, line 140: look for a more explicit naming option

see if we can avoid initializing storage clients that aren't needed for accessing the given path

lucille-core/src/main/java/com/kmwllc/lucille/util/FileContentFetcher.java
```diff
@@ -0,0 +1,187 @@
...
+    // chars, rather than throwing an Exception.
+    // https://stackoverflow.com/questions/26268132/all-inclusive-charset-to-avoid-java-nio-charset-malformedinputexception-input
+    return new BufferedReader(new InputStreamReader(stream, encoding));
+  }
+
+  /**
+   * Counts the number of lines in the file at the given path.
+   */
+  public int countLines(String path) throws IOException {
+    try (BufferedReader reader = new BufferedReader(getReader(path))) {
```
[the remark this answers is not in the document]
mrwrench24: Turns out this was here in the existing code (from `FileUtils`) for good reason... it gives access to the `readLine` method. We could just cast it / check its an instanceof a `BufferedReader` since it only calls FileFetcher's method...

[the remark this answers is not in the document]
rseitz: i think it's reasonable to change the signature of FileContentFetcher.getReader to return BufferedReader. That way callers know the reader is already buffered & they don't have to wrap it; also we wouldn't have to cast to BufferedReader here in countLines

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -46,41 +37,43 @@ public abstract class BaseStorageClient implements StorageClient {
...
-    this.maxNumOfPages = cloudOptions.containsKey(MAX_NUM_OF_PAGES) ? (Integer) cloudOptions.get(MAX_NUM_OF_PAGES) : 100;
+
+    // only matters for traversals
+    this.maxNumOfPages = cloudOptions.hasPath(MAX_NUM_OF_PAGES) ? cloudOptions.getInt(MAX_NUM_OF_PAGES) : 100;
+  }
+
+  @Override
+  public final void init() throws IOException {
+    initializeStorageClient();
+    this.initialized = true;
```
rseitz, line 59: can remove "this."

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -46,41 +37,43 @@ public abstract class BaseStorageClient implements StorageClient {
...
+  }
+
+  @Override
+  public final void shutdown() throws IOException {
+    this.initialized = false;
```
rseitz, line 64: should this be conditional on initialized==true  ?

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
rseitz: should this be conditional on initialized==false ?

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -44,35 +45,75 @@ public abstract class BaseStorageClient implements StorageClient {
...
+   */
   public boolean isInitialized() {
     return this.initialized;
   }
 
-  protected abstract void initializeStorageClient() throws IOException;
+  @Override
+  public final void traverse(Publisher publisher, TraversalParams params) throws Exception {
+    if (!isInitialized()) {
+      throw new IOException("This StorageClient has not been initialized.");
```
rseitz, line 94: shouldn't be IOException because we haven't actually started an IO
I think IllegalStateException makes sense here

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -44,35 +45,75 @@ public abstract class BaseStorageClient implements StorageClient {
...
+      clearFileHandlers();
+    }
+  }
+
+  protected abstract void traverseStorageClient(Publisher publisher, TraversalParams params) throws Exception;
+
+  @Override
+  public final InputStream getFileContentStream(URI uri) throws IOException {
+    if (!isInitialized()) {
+      throw new IOException("This StorageClient has not been initialized.");
```
rseitz, line 110: same comment as above -- change to IllegalStateException

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -20,120 +20,131 @@
...
-
-  @Override
-  protected String getStartingDirectory() {
-    String path = pathToStorageURI.getPath();
-    // path is in the format /containerName/folder1/folder2/... so need to return folder1/folder2/...
-    String[] subPaths = path.split("/", 3);
-    return subPaths.length > 2 ? subPaths[2] : "";
+  protected void validateOptions(Config cloudOptions) {
+    if (!validOptions(cloudOptions)) {
+      throw new IllegalArgumentException("Either '" + AZURE_CONNECTION_STRING + "' or '" + AZURE_ACCOUNT_NAME + "' & '" + AZURE_ACCOUNT_KEY + "' has to be in cloudOptions for AzureStorageClient.");
```
rseitz, line 40: exception message still references cloudOptions but we took that out -- should update in accordance with the new config format

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -34,137 +31,147 @@ public class S3StorageClient extends BaseStorageClient {
...
-    super(pathToStorage, docIdPrefix, excludes, includes, cloudOptions, fileOptions);
+  public S3StorageClient(Config s3CloudOptions) {
+    super(s3CloudOptions);
   }
 
   @Override
-  public void init() throws ConnectorException {
+  protected void validateOptions(Config cloudOptions) {
+    if (!validOptions(cloudOptions)) {
+      throw new IllegalArgumentException("Missing '" + S3_ACCESS_KEY_ID + "' or '" + S3_SECRET_ACCESS_KEY + "' or '" + S3_REGION + "' in cloudOptions for S3StorageClient.");
```
rseitz, line 41: exception message still references cloudOptions but we took that out -- should update in accordance with the new config format

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -34,137 +31,147 @@ public class S3StorageClient extends BaseStorageClient {
...
-  public S3StorageClient(URI pathToStorage, String docIdPrefix, List<Pattern> excludes, List<Pattern> includes,
-      Map<String, Object> cloudOptions, Config fileOptions) {
-    super(pathToStorage, docIdPrefix, excludes, includes, cloudOptions, fileOptions);
+  public S3StorageClient(Config s3CloudOptions) {
+    super(s3CloudOptions);
   }
 
   @Override
-  public void init() throws ConnectorException {
+  protected void validateOptions(Config cloudOptions) {
```
rseitz, line 39: change from "Config cloudOptions" to "Config config" ?

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -172,28 +183,26 @@ private Document blobItemToDoc(BlobItem blob, InputStream is, String decompresse
...
-    return shouldIncludeFile(blob.getName(), includes, excludes);
+    return params.shouldIncludeFile(blob.getName());
   }
 
-  public static void validateOptions(Map<String, Object> cloudOptions) {
-    if (!cloudOptions.containsKey(AZURE_CONNECTION_STRING) &&
-        !(cloudOptions.containsKey(AZURE_ACCOUNT_NAME) && cloudOptions.containsKey(AZURE_ACCOUNT_KEY))) {
-      throw new IllegalArgumentException("Either '" + AZURE_CONNECTION_STRING + "' or '" + AZURE_ACCOUNT_NAME + "' & '" + AZURE_ACCOUNT_KEY + "' has to be in cloudOptions for AzureStorageClient.");
-    }
+  public static boolean validOptions(Config cloudOptions) {
```
rseitz, line 199: this method seems to only be called from within validateOptions()
please move this logic inside validateOptions and eliminate this static validOptions method

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/GoogleStorageClient.java
```diff
@@ -162,17 +172,15 @@ private Document blobToDoc(Blob blob, InputStream content, String decompressedFu
...
     }
 
     return doc;
   }
 
-  public static void validateOptions(Map<String, Object> cloudOptions) {
-    if (!cloudOptions.containsKey(GOOGLE_SERVICE_KEY)) {
-      throw new IllegalArgumentException("Missing " + GOOGLE_SERVICE_KEY + " in cloudOptions for GoogleStorageClient.");
-    }
+  public static boolean validOptions(Config cloudOptions) {
```
rseitz, line 182: same comment as elsewhere:
this method seems to only be called from within validateOptions()
please move this logic inside validateOptions and eliminate this static validOptions method

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -34,137 +31,147 @@ public class S3StorageClient extends BaseStorageClient {
...
+    return params.getPathToStorageURI().getScheme() + "://" + getBucketOrContainerName(params) + "/" + obj.key();
   }
 
-  public static void validateOptions(Map<String, Object> cloudOptions) {
-    if (!cloudOptions.containsKey(S3_ACCESS_KEY_ID) ||
-        !cloudOptions.containsKey(S3_SECRET_ACCESS_KEY) ||
-        !cloudOptions.containsKey(S3_REGION)) {
-      throw new IllegalArgumentException("Missing '" + S3_ACCESS_KEY_ID + "' or '" + S3_SECRET_ACCESS_KEY + "' or '" + S3_REGION + "' in cloudOptions for S3StorageClient.");
-    }
+  public static boolean validOptions(Config cloudOptions) {
```
rseitz, line 171: same comment as elsewhere:
this method seems to only be called from within validateOptions()
please move this logic inside validateOptions and eliminate this static validOptions method

rseitz (approved):
Approving with request to do a manual test against S3 before merging

mrwrench24:
Manual test succeeded. See the Jira ticket for more info / the `.conf` file I used.

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -102,7 +106,6 @@ public FileConnector(Config config) throws ConnectorException {
     List<String> excludeRegex = config.hasPath("excludes") ?
         config.getStringList("excludes") : Collections.emptyList();
     this.excludes = excludeRegex.stream().map(Pattern::compile).collect(Collectors.toList());
-    this.cloudOptions = config.hasPath("cloudOptions") ? config.getConfig("cloudOptions").root().unwrapped() : Map.of();
     this.fileOptions = config.hasPath("fileOptions") ? config.getConfig("fileOptions") : ConfigFactory.empty();
     try {
       this.storageURI = new URI(pathToStorage);
```
kiratraynor, line 111: This might not be necessary, but I'm wondering if there should be a check and log (info/warn?) if multiple storage clients are set here. For example, I think it would technically allowed to add a "gcp" param and an "s3" param at the same time. If the pathToStorage is a gcp url ("gs"), it will work fine, but logging a warning that there are two configurations for a storage client are included could be good.

lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/CSVFileHandler.java
```diff
@@ -3,13 +3,14 @@
...
 import com.kmwllc.lucille.core.Document;
-import com.kmwllc.lucille.util.FileUtils;
+import com.kmwllc.lucille.util.FileContentFetcher;
 import com.opencsv.CSVParser;
 import com.opencsv.CSVParserBuilder;
 import com.opencsv.CSVReader;
 import com.opencsv.CSVReaderBuilder;
 import com.opencsv.exceptions.CsvValidationException;
 import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
```
kiratraynor, line 13: [nit] looks like this import was added but is unused

kiratraynor (approved):

kiratraynor:
LGTM!

PR #263, opened 2025-02-27
LC-571: Generic File Handler Stage

mrwrench24:
Creates the `ApplyFileHandlers` stage, which tries to apply file handlers, as appropriate, using the path from a specified field in the documents. Or, you can provide a `byte[]` and a path String used to determine the appropriate FileHandler to apply on the `byte[]` rather than fetching the contents. Uses `FileContentFetcher` to resolve the paths and potentially handle cloud files, if the necessary configuration is supplied by the user.


lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyFileHandlers.java
```diff
@@ -0,0 +1,118 @@
...
+    String filePath = doc.getString(filePathField);
+    String fileExtension = FilenameUtils.getExtension(filePath);
+
+    if (!FileHandler.supportAndContainFileType(fileExtension, handlerOptions)) {
+      return null;
+    }
+
+    try {
+      InputStream fileContentStream = fileFetcher.getInputStream(filePath);
+      FileHandler handler = FileHandler.create(fileExtension, handlerOptions);
```
rseitz, line 109: where do we use the fileHandlers we populated in start()

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyFileHandlers.java
```diff
@@ -0,0 +1,139 @@
...
+    this.filePathField = ConfigUtils.getOrDefault(config, "filePathField", "file_path");
+    this.fileContentField = ConfigUtils.getOrDefault(config, "fileContentField", "file_content");
+
+    this.fileFetcher = new FileContentFetcher(config);
+    this.fileHandlers = new HashMap<>();
+  }
+
+  @Override
+  public void start() throws StageException {
+    for (String fileExtensionSupported : SUPPORTED_FILE_TYPES) {
```
rseitz, line 77: can this logic for populating a filehandlers array be extracted into a utility method?
we have to perform the same logic in BaseStorageClient where we have an initializeFileHandlers() method
I'm thinking, just like we have a static FileHandler.create() could we also have something like FileHandler.createFromConfig... returning a HashMap populated with all the handlers specified in the config?

lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/FileHandler.java
```diff
@@ -56,6 +58,19 @@ static FileHandler create(String fileExtension, Config fileOptions) {
     }
   }
 
+  /**
+   * Populates the given map, mapping file extension Strings to FileHandlers. File Handlers are created based on the given
+   * config.
+   */
+  static void populateFromConfig(Map<String, FileHandler> map, Config optionsWithHandlers) {
```
rseitz, line 65: is there a reason you prefer having this populate a given map instead of creating a new map and returning it?

returning a new map seems more predictable as we don't have to consider the state of the existing map -- i.e. we can't be overwriting anything that might already be there, and there can't be anything stale/extraneous in there

mrwrench24: I did this so `BaseStorageClient` and `ApplyFileHandlers` could still have the map as a final field that they later populate. This is more important for the `BaseStorageClient`, since we only use `FileHandlers` during a traversal (when we have `TraversalParams`). We could just overwrite the field, I wasn't sure which approach you'd rather go with.

rseitz: I think it's acceptable for BaseStorageClient's fileHandler's list to be non-final, so I'd prefer having the utility method return the HashMap. But it should return an unmodifiable version of the map, by passing it through Collections.unmodifiableMap()

rseitz (approved):

PR #266, opened 2025-03-03
LC-400: Parquet File Handler

mrwrench24:
1. Creates a `ParquetFileHandler` and a `ParquetFileIterator` which have more digestible logic and fit into the `FileConnector` framework. The `FileConnector` supports a `parquet` entry in `fileOptions` like a CSV or JSON file handler as well. (`parquet` is still a separate module, which is still a good idea - Hadoop, which is still used, has a lot of transitive dependencies.) 
2. Some minor logic was moved into `ParquetDocUtils`. `ParquetConnector` takes advantage of the two functions here as well - a small change to the connector itself. 
3. In order to support the new framework, **the `ParquetFileHandler` reads the entire `InputStream` into memory.** There is a lot of reasoning as to why this had to be done - the `InputStream` used under the hood needs to be `Seekable` - in other words, _remarkably_ flexible. While it would be nice to not have to do this, I don't see a way to take in a generic `InputStream`, which could be an S3 InputStream, for example, and quickly "bounce around" the contents as if they are linear, when in fact, we are communicating with a cloud provider! To comply with a `ParquetFileReader`, I created a custom `InputFile` the `ParquetFileReader` can open - the `ByteArrayInputFile`. This `InputFile` is just a "file" for the `byte[]`. I also created a custom `InputStream` for this - a `SeekableByteArrayInputStream`. 

There was also some generic simplification of the logic in reading Parquet files. You'll notice a lot of the code in `ParquetFileIterator` comes from the `ParquetConnector` with some slight variations. `start` has been renamed to `numToSkip`.

Unlike the `ParquetConnector`, the  `start` and `limit` parameters are applied per _each Parquet file_, unlike the `ParquetConnector`, where they are applied _per traversal_.


meghanboyd:
Leaving in draft state until @mrwrench24  can do final review -- please decline if appropriate or discuss with group if there is another approach still requiring investigation.

mrwrench24:
I'm closing this since I think we'd have to jump through a few hoops here.
1. Using the `pathStr` in `FileHandler.processFile()`, which is intended for debugging purposes only.
2. Not seeing straightforward configuration for Google Cloud and Hadoop (I can see it exists, but it won't be as simple as Azure + S3 would be)
3. As we had mentioned previously, it is a very unlikely use case, so it doesn't feel worthwhile to make a confusing use case that is unlikely + would need a lot of verbose configuration, probably

PR #269, opened 2025-03-04
LC-640: FetchFileContent stage

mrwrench24:
Creates a stage that allows pipelines to fetch file contents (`byte[]`) outside of the `FileConnector` itself by using a `FileContentFetcher`.


lucille-core/src/test/java/com/kmwllc/lucille/stage/FetchFileContentTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+import org.junit.Test;
+
+public class FetchFileContentTest {
+
+  private final StageFactory factory = StageFactory.of(FetchFileContent.class);
+
+  Path helloPath = Paths.get("src/test/resources/FetchFileContentTest/hello.txt");
```
rseitz, line 23: helloPath, goodbyePath, helloContents, and goodbyeContents should be private

the explicitly declared FetchFileContentTest() constructor isn't needed so I'd remove it

rseitz: pls. just move the initialization of helloContents and goodbyeContents into the constructor
that makes it clearer to someone scanning the code why the constructor is declared with that exception
also gives you the chance to catch & handle the exception (which you don't have to do here because this is only a test)
in general, when initializing instance variables, do it inside the constructor if the initialization is potentially expensive and/or error-prone (especially anything involving I/O)

rseitz (approved):

PR #270, opened 2025-03-05
LC-643: More rigid VerifyIngestResult assertions

mrwrench24:
In the event of a test failure, it'll be a bit clearer which field caused the assertion to fail - so we can see it directly from the nightly smoketest and won't have to replicate locally by running the example and then running docker compose up, to gain access to Solr.

### DIFF
--- a/lucille-examples/lucille-distributed-example/src/test/java/com/kmwllc/lucille/distributed/VerifyIngestResult.java
+++ b/lucille-examples/lucille-distributed-example/src/test/java/com/kmwllc/lucille/distributed/VerifyIngestResult.java
@@ -5,6 +5,7 @@ import static org.junit.Assert.assertEquals;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 import org.junit.Test;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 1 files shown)


lucille-examples/lucille-distributed-example/src/test/java/com/kmwllc/lucille/distributed/VerifyIngestResult.java
```diff
@@ -27,30 +27,46 @@ public void verifyIngestResult() throws Exception {
...
-    assertTrue(
-        docs.stream().anyMatch(doc -> doc.get("id").asText().equals("3") && doc.get("my_country").get(0).asText().equals("USA")
-            && doc.get("my_price").get(0).asInt() == 30 && doc.get("my_name").get(0).asText().equals("Burger")));
-    assertTrue(
-        docs.stream().anyMatch(doc -> doc.get("id").asText().equals("4") && doc.get("my_country").get(0).asText().equals("USA")
-            && doc.get("my_price").get(0).asInt() == 10 && doc.get("my_name").get(0).asText().equals("Salad")));
-    assertTrue(
-        docs.stream().anyMatch(doc -> doc.get("id").asText().equals("5") && doc.get("my_country").get(0).asText().equals("France")
-            && doc.get("my_price").get(0).asInt() == 12 && doc.get("my_name").get(0).asText().equals("Wine")));
+    JsonNode csvDoc1 = docs.get(0);
```
rseitz, line 30: can we be sure the docs will always appear in dest.json in this exact order?

dest.json is created via a solr query http://solr:8983/solr/quickstart/query?q=*:*'

with a "*:*" query without a sort param, the docs will be returned in the order indexed

if we have multiple worker threads its conceivable that the docs will be indexed in a different order from what's expected here

could add a sort to the query (on the id) to be explicit about the order we want, but that would change the order from what the test is currently expecting

rseitz: could also sort the docs here in this Junit test code before checking their values... probably easier to do it this way

PR #276, opened 2025-03-12
LC-647: Removed VFSConnector / VFS Dependencies

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/connector/README.md
mrwrench24, line 1: I deleted this instead of rewriting it since we have some `FileConnector` documentation in a separate ticket

rseitz (approved):

PR #278, opened 2025-03-17
LC-652: Query OpenSearch Stage

mrwrench24:
This stage allows users to run an Opensearch query, using information from Documents as parameters / variables, and then specify a portion of the response that they want to place on the Document.

The stage relies on Opensearch's search templates to operate. There are two ways they can be used:
1. Users can provide a `templateName` - the name of an existing search template they have saved to their Opensearch cluster. 
2. Users can provide the body of the template (`searchTemplate`). The template will not be saved to their Opensearch cluster by Lucille.

For each Document we process, we create an appropriate request to send to Opensearch.
1. When given a template name, the request looks like this:
```
{
  "id": "template_name",
  "params": {
    "park_name": "Neck Creek Preserve"
  }
}
```
2. When given a `searchTemplate`, the request looks like this:
```
{
  # Just the user's template
  "source": {
    "query": {
      "match_phrase": {
        "park_name": "{{park_name}}"
      }
    }
  },

  # Added by Lucille
  "params": {
    "park_name": "Neck Creek Preserve"
  }
}
```

Users need to specify which parameter names they **require** to be extracted from Documents and which they decide are **optional**. You can specify default values in the Opensearch template (see their documentation for more information), but if parameters without a default value are missing, Opensearch just returns a response with 0 hits.

I manually tested this, both with an existing template in Opensearch as well as a template defined in Lucille, and everything appeared to be working nicely.


lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -0,0 +1,148 @@
...
+      JsonNode currentNode = objectMapper.readTree(response.body());
+
+      JsonNode responseFieldNode = currentNode.at(opensearchResponsePath);
+      opensearchQueryResponse = responseFieldNode.toString();
+    } catch (Exception e) {
+      throw new StageException("Error occurred executing the Opensearch Query.", e);
+    }
+  }
+
+  // TODO: If we require Java 21 we could add a call to httpClient.close().
```
mrwrench24, line 105: `HttpClient` doesn't receive a `.close()` (or similar method) until Java 21 and later.

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -0,0 +1,177 @@
...
+
+    this.destinationField = ConfigUtils.getOrDefault(config, "destinationField", "response");
+
+    httpClient = HttpClient.newHttpClient();
+  }
+
+  @Override
+  public void start() throws StageException {
+    // Register the search template with Opensearch.
+    HttpRequest request = HttpRequest.newBuilder()
```
rseitz, line 98: for the first pass at this stage, I don't think we should provide the ability to save the template in OpenSearch; instead, we can send the entire user-provided query with the params on every request. If the user has _already_ saved the template in Opensearch, they can provide a simplified query in the stage config that references the saved template by its ID. If the user hasn't saved the template, they can provide the full body of the query...

rseitz: I just mean we can take the query the user provides and fill in the params section of it. They have the choice of providing a full template (which would then get sent to opensearch on every request, but never saved) or they can include json that references an already saved template. Can discuss further...

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -0,0 +1,205 @@
...
+        .build();
+
+    try {
+      HttpResponse<String> response = httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
+
+      ObjectMapper objectMapper = new ObjectMapper();
+      JsonNode currentNode = objectMapper.readTree(response.body());
+
+      JsonNode responseFieldNode = currentNode.at(opensearchResponsePath);
+      doc.setField(destinationField, responseFieldNode.toString());
```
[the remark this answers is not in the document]
rseitz: suggest we log a warning

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -0,0 +1,205 @@
...
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+    ObjectNode requestJson;
+
+    if (searchTemplateJson != null) {
+      // Use an existing search template. Need the json for the template, will populate it with "params" for this document.
+      requestJson = (ObjectNode) searchTemplateJson;
+    } else {
+      // 2. Lookup by an existing template. Populate the "id" field. Will populate it with "params" for this document.
+      ObjectMapper mapper = new ObjectMapper();
```
rseitz, line 136: can be private static final

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -0,0 +1,205 @@
...
+    return null;
+  }
+
+  /**
+   * Adds a "params" node to the root of the given JSON, populated with the required / optional param names present on the given Document.
+   * Throws a NoSuchFieldException if a "requiredParam" is not present on the given document.
+   * (package access so we can test, make sure we are getting the appropriate objects)
+   */
+  void populateJsonWithParams(Document doc, ObjectNode json) throws NoSuchFieldException {
+    ObjectMapper objectMapper = new ObjectMapper();
```
rseitz, line 177: can be private static final

spencersolomon6 (approved):

kiratraynor (approved):

PR #279, opened 2025-03-18
LC-563: Incremental File Connector, V1

mrwrench24:
Allows you to filter out files that were modified before/after a certain amount of time. Minor refactor in `FileConnector` to take in a new map, `filterOptions`, where `includes` and `excludes` are now found. Some minor refactoring in `StorageClient` to apply these "filtration checks" in one place - at the beginning of `processAndPublishFileIfValid`.


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -184,10 +185,10 @@ private Document blobItemToDoc(BlobItem blob, InputStream is, String decompresse
     return doc;
   }
 
-  private boolean isValid(BlobItem blob, TraversalParams params) {
-    if (blob.isPrefix()) return false;
-
-    return params.shouldIncludeFile(blob.getName());
+  @Override
+  protected boolean validFile(FileReference fileRef) {
```
mrwrench24, line 189: isValidFile

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -135,63 +104,89 @@ private String getFullPath(BlobItem blobItem, TraversalParams params) {
...
+  private class AzureFileReference extends BaseFileReference {
 
-    if (properties.getCreationTime() != null) {
-      doc.setField(FileConnector.CREATED, properties.getCreationTime().toInstant());
-    }
+    private final BlobItem blobItem;
 
-    doc.setField(FileConnector.SIZE, properties.getContentLength());
+    public AzureFileReference(BlobItem blobItem) {
+      super(blobItem.getProperties().getLastModified().toInstant());
```
rseitz, line 118: does the BlobItem hold this value or does it have to make an external call to Azure to retrieve it?
if there's an external call here, we should only do it once we know that the lastModified time is actually needed

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/GoogleStorageClient.java
```diff
@@ -128,69 +94,88 @@ protected String getStartingDirectory(TraversalParams params) {
...
-    String fullPath = getFullPath(blob, params);
-    String docId = DigestUtils.md5Hex(fullPath);
-    Document doc = Document.create(params.getDocIdPrefix() + docId);
+  private class GoogleFileReference extends BaseFileReference {
+
+    private final Blob blob;
 
-    doc.setField(FileConnector.FILE_PATH, fullPath);
+    public GoogleFileReference(Blob blob) {
+      super(blob.getUpdateTimeOffsetDateTime().toInstant());
```
rseitz, line 112: I see from the docs that this is an inexpensive call not involving RPC: "Blob object includes the BlobId instance, the set of properties inherited from the [BlobInfo](https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.BlobInfo) class and the Storage instance. The class provides methods to perform operations on the object. Reading a property value does not issue any RPC calls." https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.Blob.  A code comment mentioning this would be helpful.

mrwrench24: adding comment...

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
mrwrench24: It doesn't appear to... I dug through the `.class` files and there doesn't appear to be any remote calls, just plumbing of various classes / internal classes etc.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -129,58 +101,82 @@ protected String getStartingDirectory(TraversalParams params) {
...
+    this.s3 = s3;
+  }
+
+
+  private class S3FileReference extends BaseFileReference {
+
+    private final S3Object s3Obj;
+
+    public S3FileReference(S3Object s3Obj) {
+      super(s3Obj.lastModified());
```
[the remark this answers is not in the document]
mrwrench24: The `S3Object` class has a `lastModified` field which is returned by `lastModified()` (a simple getter.). Particularly since the `S3StorageClient` uses "V2" of listing objects.

lucille-core/src/test/resources/FileConnectorTest/success.conf
```diff
@@ -10,6 +10,8 @@
     }
     moveToAfterProcessing: "success"
   }
-  includes: []
-  excludes: []
+  filterOptions: {
```
rseitz, line 13: do we allow the user to omit filterOptions? (I'm thinking we should, and should have a test for it)

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
rseitz: very very minor but there are several calls going on here including the call to super
could you update this comment and the others to say something like "getting the last modified time is inexpensive -- it is stored inside the BlobItem"

lucille-core/src/test/resources/FileConnectorTest/success.conf
mrwrench24: It's not a required property. I'm not sure why I copy pasted the map into each file - maybe includes / excludes was in each file and I was just refactoring it...

I deleted the map from the example files that aren't using it and the tests are passing as before... thx!

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
mrwrench24: thanks... I had thought I had gone through and had these just use getLastModified (the stored / cached `Instant`) but I guess I hadn't.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -197,4 +96,86 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+      return path.toString();
+    }
+
+    @Override
+    public boolean isValidFile() {
+      return true;
+    }
+
+    @Override
+    public InputStream getContentStream(TraversalParams params) {
```
rseitz, line 132: should add javadoc saying that the returned InputStream is non-buffered
depending on how the caller is going to use this stream, the caller might want to wrap it in a BufferedInputStream

PR #280, opened 2025-03-20
LC-656: FileReference refactor / enhancements.

mrwrench24:
**Note** that the base of this PR is `LC-563` as it contains those changes as well.

This ticket is primarily a refactor to `FileReference` and includes some subsequent and unrelated cleanup in `BaseStorageClient` / the implementations. As a result, `FileReference` is now a bit "safer" to use, and we no longer need to create `StorageClient`-specific methods that simply extract information from a FileReference merely by knowing which storage object (`Blob`, `Path`, `S3Object`) is appropriate to retrieve. 

`FileReference` is now an interface, with each `StorageClient` having an inner class implementation. A small amount of common functionality is stored in `BaseFileReference`. Many of the "file reference getter" methods have been removed and replaced with direct methods on the `FileReference` - methods like `isValidFile`, `getFileName`, `convertFileReferenceToDoc`, and `getFileReferenceContentStream`.

In terms of `BaseStorageClient` changes not necessarily related to `FileReference`:
1. I made some `BaseStorageClient` methods (like `handleArchive` / `handleStream` files) private since they never get referenced in an individual `StorageClient` implementation.
2. I removed some abstract methods from `BaseStorageClient` - `getStartingDirectory` and `getBucketOrContainerName` - since they never get referenced in `BaseStorageClient`. Each `StorageClient` implementation still has a `private` implementation, but future `StorageClients` may not necessarily need to implement this method!
3. We now get the `fullPathStr` and `fileExtension` at the very beginning of `processAndPublishFileIfValid` - this was a few lines of code that each `StorageClient` was just repeating and can help prevent errors in the future, maybe.


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/FileReference.java
```diff
@@ -1,61 +1,59 @@
...
+
+  /**
+   * @return The extension associated with this file.
+   */
+  String getFileExtension();
+
+  /**
+   * @return Whether this FileReference is for a cloud file.
+   */
+  boolean isCloudFileReference();
```
mrwrench24, line 30: This method doesn't get used. not sure if we want to leave it in just in case it is needed in the future.

rseitz: let's remove -- can add later if needed

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/FileReference.java
```diff
@@ -1,61 +1,59 @@
...
+  /**
+   * @return An InputStream for the file's contents, using the given TraversalParams as needed.
+   */
+  InputStream getContentStream(TraversalParams params);
+
+  /**
+   * @return A Lucille Document from this file reference. Will get the file's contents if params.shouldGetFileContent()
+   * is true.
+   */
+  Document toDoc(TraversalParams params);
```
rseitz, line 52: wondering about asDoc instead of toDoc

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -20,12 +20,102 @@
 import java.net.URI;
 import java.time.Duration;
 import org.apache.commons.codec.digest.DigestUtils;
-import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class AzureStorageClient extends BaseStorageClient {
 
+  private class AzureFileReference extends BaseFileReference {
```
rseitz, line 28: suggest moving this to end of the enclosing class

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/GoogleStorageClient.java
```diff
@@ -21,12 +21,97 @@
 import java.nio.channels.Channels;
 import java.util.Objects;
 import org.apache.commons.codec.digest.DigestUtils;
-import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class GoogleStorageClient extends BaseStorageClient {
 
+  public class GoogleFileReference extends BaseFileReference {
```
rseitz, line 29: review whether this should be public or private

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -23,130 +22,85 @@
...
-      if (params.shouldGetFileContent()) {
-        doc.setField(CONTENT, Files.readAllBytes(path));
+    @Override
+    public Document toDoc(TraversalParams params) {
+      String fullPath = path.toAbsolutePath().normalize().toString();
+      String docId = DigestUtils.md5Hex(fullPath);
+      Document doc = Document.create(createDocId(docId, params));
+
+      try {
+        BasicFileAttributes attrs = Files.readAttributes(path, BasicFileAttributes.class);
```
rseitz, line 77: let's confirm that this isn't adding a file read where we weren't always doing it before

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -184,10 +167,8 @@ public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) th
     @Override
     public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
       // Visit the file and actually process it!
-      Path fullPath = file.toAbsolutePath().normalize();
-      String fullPathStr = fullPath.toString();
-      String fileExtension = FilenameUtils.getExtension(fullPathStr);
-      processAndPublishFileIfValid(publisher, fullPathStr, fileExtension, new FileReference(fullPath, attrs.lastModifiedTime().toInstant()), params);
+      FileReference fileRef = new LocalFileReference(file, attrs.lastModifiedTime().toInstant());
```
rseitz, line 170: given that we have the attributes already here, can we pass them into the LocalFileReference instead of acquiring them a second time inside toDoc?

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -26,6 +25,81 @@
...
+    public InputStream getContentStream(TraversalParams params) {
+      String objKey = s3Obj.key();
+      GetObjectRequest objectRequest = GetObjectRequest.builder().bucket(getBucketOrContainerName(params)).key(objKey).build();
+      return s3.getObject(objectRequest);
+    }
+
+    @Override
+    public Document toDoc(TraversalParams params) {
+      String fullPath = getFullPath(params);
+      String docId = DigestUtils.md5Hex(fullPath);
```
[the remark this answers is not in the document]
mrwrench24: I looked into this, although there is a lot of repeated code, I think this might do more harm than good.

- The two "asDoc" methods in the `FileReference` interface will become four methods in `BaseFileReference` - the two "asDoc" methods, plus two `FileReference` specific implementations (abstract method) for actually adding the fields (creation, size, file contents.)
- Two methods (one for regular, one for archive / compressed) for populating the document are still needed because of the different ways to get the file content - whether we have the input stream (archive / compressed) or not (non-archive/compressed)
- `createDocId`, a protected method in `BaseStorageClient`, has to become a static method in `StorageClient`. This isn't ideal because we would rather it work with `AbstractConnector.createDocId` - this gets us farther from our goal here.

If we refactor how we handle archive / compressed files that don't get passed to a file handler in the future, this task might become more doable. @rseitz let me know what you think - the tradeoffs here certainly could be worth it, but I was worried I was jumping through a lot of hoops to get this done.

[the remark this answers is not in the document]
mrwrench24: I pushed the "createEmptyDocument" changes we had discussed earlier. 

I'm thinking we may want to revisit our handling of archive files / FileReferences interacting in the future, but we don't want to hold back on the state improvements. I'll make a followup ticket and we can discuss if needed!

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -204,4 +96,87 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
...
+          throw new IllegalArgumentException("Unable to get file contents from '" + path + "' for document", e);
+        }
+      }
+
+      return doc;
+    }
+
+    @Override
+    public Document asDoc(InputStream in, String decompressedFullPathStr, TraversalParams params) throws IOException {
+      String docId = DigestUtils.md5Hex(decompressedFullPathStr);
```
rseitz, line 165: can we use createEmptyDocument here?

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/GoogleStorageClient.java
```diff
@@ -126,72 +94,89 @@ protected String getStartingDirectory(TraversalParams params) {
...
-    if (params.shouldGetFileContent()) {
-      doc.setField(FileConnector.CONTENT, content.readAllBytes());
+      return doc;
     }
 
-    return doc;
-  }
+    @Override
+    public Document asDoc(InputStream in, String decompressedFullPathStr, TraversalParams params) throws IOException {
+      final String docId = DigestUtils.md5Hex(decompressedFullPathStr);
```
rseitz, line 161: can we use createEmptyDocument here?

mrwrench24: Not in the existing changes... But I pushed a small change that allows us to do this. I'm a bit worried it could create more confusion / opens the door to potential errors, but let me know what you think...

rseitz (approved):

mrwrench24:
Just noting again this was merged into LC-563.

PR #282, opened 2025-03-25
LC-649: Parquet Connector Refactor + Unit Tests

mrwrench24:
1. Refactors the `ParquetConnector` to use `ParquetFileIterator` to get documents after it acquires / opens a `ParquetFileReader`. This code was created + tested in a prior ticket but hasn't yet been merged.
2. Created tests for the `ParquetConnector`.

**Notes:**
- There _was_ a one-off bug with `skipAndUpdateStart` in the previous connector, where setting start to 0 or 1 was the same, setting it to 2 would skip only one document, 3 would only skip two, ... etc. This was naturally resolved by just introducing the `ParquetFileIterator`. If we don't want to merge the Iterator change, we should fix the bug (and we'd see it in the unit tests before - I was.)
- The behavior of the `start` parameter is different now. Previously, the Connector would just skip the first `n` documents it encounters. Now it skips the first `n` documents _from each file it encounters_. I'm not sure what the use case was here. It seemed like an open door to unpredictable data loss, since it'll depend on the order in which files are returned by Hadoop! It didn't really work with the refactor, but the more I thought about it, the more I felt this just opened the door to an unsatisfactory ingestion.
- `path` was renamed to `pathToStorage`. I know we have a ticket open to do the opposite of this for the `FileConnector`, but I keep getting errors trying to test connectors that still use path. It seems to be conflicting with some other key / environment variable somewhere - maybe the actual terminal `path` somehow, I don't know. We can discuss :)


meghanboyd:
@mrwrench24 we should add code commend that if you're trying to paginate you should only use one connector at a time

rseitz (approved):

mrwrench24:
added javadoc (for config parameters) and also added comment:

`If you are paginating (using start / limit), it is recommended you use individual Connectors for each Parquet file.`
@meghanboyd

PR #283, opened 2025-03-26
LC-318: Enhanced DocLogging

mrwrench24:
Created after our discussion on opening a PR from the logs work. Has additional / updated DocLogger statements as well as nanoseconds in the EcsLayout. And moves the `EcsLayout` into `main/resources`.


lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
spencersolomon6, line 1: Should we log if a document is dropped in the Stage class?

mrwrench24: Not sure if the checks I added are sufficient... lmk

lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
```diff
@@ -160,17 +162,25 @@ public Iterator<Document> processConditional(Document doc) throws StageException
         context = timer.time();
       }
       try {
+        docLogger.info("Stage {} to process {}.", name, doc.getId());
         return processDocument(doc);
       } finally {
+        if (doc.isDropped()) {
+          docLogger.info("Document {} was dropped in Stage {}", doc.getId(), name);
+        }
```
spencersolomon6, line 170: Sorry, looking at this again, that `doc.isDropped()` flag is true for every stage after the document is dropped, meaning we would log this message for all of those Stages. I think we would have to log in the DropDocument Stage (or any other stage which uses .setDropped), but that's probably not necessary unless there's some use case we would want to support with that (I am not aware of one).

I think this can probably be removed, sorry about that.

mrwrench24: Ahhhh okay. We could also maybe add a statement in `setDropped`, now that I think about it, which might be the easiest way to get a statement in anyways. I'll just remove this log statement now...

spencersolomon6 (approved):

PR #284, opened 2025-03-27
LC-653: PromptOllama Stage

mrwrench24:
Introduces the `PromptOllama` stage. This allows you to specify a system prompt you can use to get information from an LLM about Lucille Documents. We are particularly focused on supporting JSON responses from the LLMs. When processing a JSON response, the fields will be automatically integrated into the corresponding Lucille Document. For example, you might be passing through emails from the Enron dataset, and you can ask an LLM to determine whether it is potentially evidence of fraud and provide a brief summary of the email in a JSON object. These fields will then be added to the Document.

There is also support for non-JSON responses. In this case, the entire response goes into the `ollamaResponse` field.

I manually disabled the INFO level logs from Ollama - it was logging the entire request each time it is made, which includes the entire LLM chat history, so this would get _really_ verbose, really quickly.


lucille-core/src/main/java/com/kmwllc/lucille/stage/PromptOllama.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    // Always try to get JSON from the response, even if requireJSON is false.
+    try {
+      JsonNode node = mapper.readTree(chatResult.getResponseModel().getMessage().getContent());
+      // put all the fields from the JSON onto the Lucille Document.
+      node.fields().forEachRemaining(entry -> doc.setField(entry.getKey(), entry.getValue()));
+    } catch (JsonProcessingException e) {
+      if (requireJSON) {
+        throw new StageException("Error getting JSON from response", e);
```
spencersolomon6, line 115: I think we should specify in this log that `requireJSON` was set to true and we did not get a valid JSON response

lucille-core/src/main/java/com/kmwllc/lucille/stage/PromptOllama.java
```diff
@@ -0,0 +1,140 @@
...
+      chatResult = ollamaAPI.chat(request);
+    } catch (Exception e) {
+      throw new StageException("Error communicating with Ollama server.", e);
+    }
+
+    // Always try to get JSON from the response, even if requireJSON is false.
+    try {
+      JsonNode node = mapper.readTree(chatResult.getResponseModel().getMessage().getContent());
+      // put all the fields from the JSON onto the Lucille Document.
+      node.fields().forEachRemaining(entry -> doc.setField(entry.getKey(), entry.getValue()));
```
spencersolomon6, line 112: What happens if a field which already exists on the document is returned in the JSON response? Currently, the existing fields would be overwritten. If this is the behavior we want, we should make sure to document it.

I think we should at least use the `setOrAdd` method so that the original data isn't lost, or we should have an `update_mode` param, which can be append or overwrite.

spencersolomon6 (changes requested):
Couple comments in the Stage class, overall looks good though. Nice mocking in the tests!

lucille-core/src/main/java/com/kmwllc/lucille/stage/PromptOllama.java
mrwrench24: Small thing - to avoid a verbose extraction of the update mode I just used `UpdateMode.fromConfig()` which uses `update_mode` instead of `updateMode`. LMK if you think it's worth adding support for the camelCase name in `UpdateMode`?

spencersolomon6: Nope, I agree that sticking with convention and using `UpdateMode.fromConfig()` is best, so whatever param that expects. Looks good to me

spencersolomon6 (approved):

agileresearchservices (approved):
Looks clear and concise to me. Nice work!

PR #286, opened 2025-03-31
LC-237: Validate Connectors (& Framework for Validation)

mrwrench24:
In order to declare legal / required properties for a `Connector`, you call a `Spec.connector()` and pass it through as part of the call to `super`, like a `Stage`. 

- New class - `Spec`. Use for parts of Lucille that use a `Config` and have certain properties that are / aren't allowed or needed.
- Use a specific implementation (`Spec.stage()`, `Spec.connector()`), allowing you to leverage default values for all specs that are used for objects of a certain type.
- Configs get automatically validated in `Stage` or `AbstractConnector` (abstract / superclasses).
- Connectors get validated in `Runner` validation modes, similar to how Pipelines are. Some minor auxiliary changes in Runner were made to support this.

**Important Notes:**
- When running tests and using Config files in the repo, it's important to call `ConfigFactory.parseResourcesAnySyntax` instead of `ConfigFactory.load()`. This will avoid loading a lot of additional properties that would cause validation to fail.
- Naturally, `StageSpec` has been moved to its own class instead of an inner class for `Stage`, and some various logic / properties has been moved out of `Stage` itself. There's a convenience method to validate Conditions.
- Some of the `ConfigValidationTest` changes (the `Stage` one - not from a Lucille example) may look a little confusing at first. The biggest change are that the returned map has additional entries for connectors now, so they've usually been increased by 2 (the tested config has two connectors).
- The Lucille example configurations have been updated and now have their connectors validated as part of a build. Many of the examples needed default arguments provided so that the tests could pass.
- Maybe, in the future, in addition to validating `Indexers` as well as other Config-related elements, we could also add more options to the core interface, like "requiring one of", for example


lucille-core/src/main/java/com/kmwllc/lucille/core/configSpec/ConfigSpecUtils.java
```diff
@@ -0,0 +1,71 @@
+package com.kmwllc.lucille.core.configSpec;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Utility methods for a StageSpec (namely its validation of a Config)
+ */
+public class ConfigSpecUtils {
```
rseitz, line 10: these static methods could go directly on the ConfigSpec interface

lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
```diff
@@ -52,7 +50,7 @@ public abstract class Stage {
   private Counter childCounter;
 
   public Stage(Config config) {
-    this(config, new StageSpec());
+    this(config, ConfigSpec.forStage());
```
rseitz, line 53: Since this paradigm is repeated in so many places, I'm thinking of a name change to make it even more concise: ConfigSpec -> Spec and forStage() -> stage()

So it would look like Spec.stage().withRequiredProperties(...

what do you think?

lucille-core/src/main/java/com/kmwllc/lucille/core/Spec.java
```diff
@@ -0,0 +1,236 @@
...
+public class Spec {
+
+  private final Set<String> defaultLegalProperties;
+
+  private final Set<String> requiredProperties;
+  private final Set<String> optionalProperties;
+  private final Set<String> requiredParents;
+  private final Set<String> optionalParents;
+
+  private String displayName;
```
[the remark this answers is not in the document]
mrwrench24: I'm a little worried the only juice we are going to get out of this is passing in a name to `validate`. I think catching the exceptions in the abstract classes is a little much (since we are only wrapping the exception just to add a name, and could be conducive to future errors for objects using a `Spec`). But it doesn't have to be stored in the spec. 

But we can't pass a name in as part of a call to `stage()` or `connector()` for example. 

And catching the exceptions makes validation mode feel a little more brittle. I think that passing the name into the validate method might be the cleanest solution? Let me know what you think.

spencersolomon6 (approved):

mrwrench24:
I've been doing some manual testing and agree with a lot of what you've said... I've become convinced we need a somewhat larger overhaul/cleanup to get more into the details here at some point. And perhaps not all of them are such a simple fix.

1. The error message is perhaps an easy fix, I'll try to take a look at that now.
2. For multiple errors / bad properties in a Stage... this is also hopefully a moderate fix... we can just have `Spec.validate` collect the bad properties during a validation and include that in the error message. That would help us address the "number" being included for Connector. Currently, I think it _is_ possible to get more than 1 by having connectors of the same name - but you are correct in that multiple _validation_ errors will not be added here.
3. Child properties not being validated is something that I'm aware of and I'm not quite sure how to "cleanly" remedy, with a few exceptions (for example, we probably want to `Spec` the file handlers, namely CSV file handler, which takes in a lot of options...). Namely I'm worried about stages like `TextExtractor`, maybe, that use a `FileContentFetcher`. Today we define the allowed parents, then we'd have to define the values inside as well - like `S3.secretAccessKey`, which would add a lot of additional properties we'd have to "declare". Another good example might be `applyFileHandlers`...

TLDR I'll try to do 1 + 2 but I think 3 is an existing issue that is probably a larger effort / design to fix.

mrwrench24:
i just pushed changes for items 1 and 2. on number 2 (multiple errors) - we now include all the "invalid arguments" in the exception message. as such, i went with just removing the whole idea of "exception count" from stringify validation - we will still print out messages for duplicate connectors but i figured this notion isn't really going to apply here / in the future. the message is also now clearer about an unknown parent / property whereas before, the message always just said property. tests updated to verify + a new one was added.

rseitz:
Posting an example of what validate output looks like with the current changes (showing points 1 and 2 from discussions have been addressed). May want to refine output formatting in a separate ticket.

```
java -Dconfig.file=conf/s3-opensearch.conf -cp 'target/lib/*' com.kmwllc.lucille.core.Runner -validate
25/04/03 16:18:05   INFO Runner: Connector Configuration is invalid. Printing the list of exceptions for each element
	Connector: fileConnector
		Error(s) with fileConnector Config: [Config contains unknown parent fileOptions1, Config contains unknown property pipeline1]
25/04/03 16:18:05   INFO Runner: Pipeline Configuration is invalid. Printing the list of exceptions for each element
	Pipeline: pipeline1
		Error(s) with com.kmwllc.lucille.tika.stage.TextExtractor (TextExtractor) Config: [Config contains unknown property metadata_prefix1, Config contains unknown property byte_array_field1]
		Error(s) with com.kmwllc.lucille.stage.ChunkText (ChunkText) Config: [Config contains unknown property dest1, Config must contain property source, Config contains unknown property source1]
```

rseitz (approved):

PR #287, opened 2025-04-01
LC-660: Update S3StorageClient to use default credentials provider chain when access key is omitted

rseitz:


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -36,22 +37,32 @@ public S3StorageClient(Config s3CloudOptions) {
 
   @Override
   protected void validateOptions(Config config) {
-    if (!config.hasPath(S3_ACCESS_KEY_ID)
-        || !config.hasPath(S3_SECRET_ACCESS_KEY)
-        || !config.hasPath(S3_REGION)) {
-      throw new IllegalArgumentException("Missing '" + S3_ACCESS_KEY_ID + "' or '" + S3_SECRET_ACCESS_KEY + "' or '" + S3_REGION + "' in Config for S3StorageClient.");
+    if (!config.hasPath(S3_REGION)) {
```
rseitz, line 40: @kiratraynor do you think region should always be required or is there a scenario where a user would want to omit it too?

kiratraynor: I think there could be a use case. If is not included, it would work similarly to the default credentials provider ([DefaultAwsRegionProviderChain](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/regions/providers/DefaultAwsRegionProviderChain.html)) where it'll try a list of things looking for a region to use and if it doesn't find one it'll error. I think it would be valid to allow users to not input a region, in hopes that it will pull the region from system properties, environment variable, aws config files, or ec2 metadata.
I think we should allow a user to not give a region at all if that makes sense to you as well. And I think that providing (or not providing) a region should be independent of the credential provider type.

rseitz: thx. updated

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -36,22 +37,29 @@ public S3StorageClient(Config s3CloudOptions) {
 
   @Override
   protected void validateOptions(Config config) {
-    if (!config.hasPath(S3_ACCESS_KEY_ID)
-        || !config.hasPath(S3_SECRET_ACCESS_KEY)
-        || !config.hasPath(S3_REGION)) {
-      throw new IllegalArgumentException("Missing '" + S3_ACCESS_KEY_ID + "' or '" + S3_SECRET_ACCESS_KEY + "' or '" + S3_REGION + "' in Config for S3StorageClient.");
+    if (config.hasPath(S3_ACCESS_KEY_ID) ^ config.hasPath(S3_SECRET_ACCESS_KEY)) {
```
mrwrench24, line 40: Probably want to specify in `FileConnector` that these are no longer required if they are both omitted.

Fortunately, in the JavaDoc of `FileContentFetcher` users, we point them towards looking at `FileConnector` for the appropriate arguments - so we don't have to update those.

mrwrench24 (approved):
Just a small thing about java doc but otherwise looks good!

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
mrwrench24: Whoops! I can do this in LC-663.

rseitz: crossed path -- thx for the catch and doing it in LC-663 would be great

PR #288, opened 2025-04-02
LC-238: Validate Indexers

mrwrench24:
A few changes, besides adding `Spec.indexer()` to Indexers: 
1. Added `bypass` functionality to the `PineconeIndexer`
2. Moved `sendEnabled` from some `Elasticsearch`/`solr`/`OpenSearch`/ config maps to the `Indexer` config map in test files. `sendEnabled` is only a legal property for `Indexer` configs, not specific implementation configs.
3. Some other related property changes (usually in test files) that are no longer in use
4. Fixes a bug in `IndexerFactory` where you had to place an unknown type in `Indexer` config in order to supply and use a custom Indexer class.


lucille-core/src/main/java/com/kmwllc/lucille/core/Spec.java
```diff
@@ -49,6 +54,21 @@ public static Spec connector() {
...
 
+  /**
+   * Creates a Spec for a specific Indexer implementation (elasticsearch, solr, csv, etc.) There are no common, legal properties
+   * associated with specific Indexers.
+   * <p> <b>Note:</b> This spec should <b>not</b> be used to validate the general "indexer" block of a Lucille configuration.
+   * <p> <b>Note:</b> You should define your required/optional properties/parents <i>without</i> including your indexer-specific
+   * parent name / key. For example, do <i>not</i> write "elasticsearch.index"; instead, write "index".
+   * @return a Spec with default legal properties suitable for a specific Indexer implementation.
+   */
+  public static Spec indexer() { return new Spec(); }
```
mrwrench24, line 65: TODO: Should "sendEnabled" be used here?

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -217,6 +244,16 @@ public static Map<String, List<Exception>> validatePipelines(Config config) thro
     return exceptionMap;
   }
 
+  private static Map<String, List<Exception>> validateIndexer(Config config) {
+    try {
+      IndexerFactory.fromConfig(config, new LocalMessenger(), true, "");
```
mrwrench24, line 249: something odd about IndexerFactory we might want to look at, potentially in this ticket - you have to add a `type`, otherwise, it'll always use Solr - even if you specify a class.

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
```diff
@@ -43,7 +41,10 @@ public class ElasticsearchIndexer extends Indexer {
 
   public ElasticsearchIndexer(Config config, IndexerMessenger messenger, ElasticsearchClient client,
       String metricsPrefix, String localRunId) {
-    super(config, messenger, metricsPrefix, localRunId);
+    super(config, messenger, metricsPrefix, localRunId, Spec.indexer()
+        .withRequiredProperties("index", "url")
+        .withOptionalProperties("update", "parentName", "sendEnabled", "type")
```
mrwrench24, line 46: I don't see "type" ever being used but is referenced in numerous tests

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -46,7 +47,10 @@ public class OpenSearchIndexer extends Indexer {
   private final boolean update;
 
   public OpenSearchIndexer(Config config, IndexerMessenger messenger, OpenSearchClient client, String metricsPrefix, String localRunId) {
-    super(config, messenger, metricsPrefix, localRunId);
+    super(config, messenger, metricsPrefix, localRunId, Spec.indexer()
+        .withRequiredProperties("index", "url")
+        .withOptionalProperties("update", "sendEnabled", "type", "acceptInvalidCert"));
```
mrwrench24, line 52: Again, not seeing "type" being used...

lucille-examples/lucille-vector-ingest-example/conf/10M-parquet-pinecone.conf
```diff
@@ -30,17 +39,15 @@ indexer {
...
-  apiKey: ${PINECONE_API_KEY}
-  environment: ${PINECONE_ENVIRONMENT}
-  projectName: ${PINECONE_PROJECT}
-  timeout: 10
-  namespaces: [
-    "kmw_msmarco",
-  ]
-  embeddingFields: [
-    "values",
-  ]
```
mrwrench24, line 43: some changes here. Was not seeing embeddingFields, timeout, environment, or projectName referenced/used anywhere within the code. 
Namespaces was of the wrong type in this example.

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
mrwrench24: removed by elasticsearch

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
mrwrench24: Opensearch never really used types. Not to get too technical.

mrwrench24:
Should be good for review here :)

lucille-core/src/main/java/com/kmwllc/lucille/core/Connector.java
```diff
@@ -116,7 +116,7 @@ static List<Exception> getConnectorConfigExceptions(Config connectorConfig) {
         }
       }
     } catch (ReflectiveOperationException e) {
-      exceptionList.add(new ConnectorException("Connector class not found: " + e.getMessage()));
+      exceptionList.add(new ConnectorException("Reflective error with Connector: " + e.getMessage()));
```
rseitz, line 119: I would say something like "connector class or suitable constructor"... i.e. spell out the likely reflective errors that might occur

lucille-core/src/main/java/com/kmwllc/lucille/stage/Concatenate.java
```diff
@@ -18,10 +18,9 @@
  *
  * Config Parameters:
  *
- *   - source (List&lt;String&gt;) : list of source field names
  *   - dest (String) : Destination field. This Stage only supports supplying a single destination field.
  *   - format_string (String) : The format String, which will have field values substituted into its placeholders
- *   - defualt_inputs (Map&lt;String, String&gt;, Optional) : Mapping of input fields to a default value. You do not have to
+ *   - default_inputs (Map&lt;String, String&gt;, Optional) : Mapping of input fields to a default value. You do not have to
```
spencersolomon6, line 23: Nice catch!

lucille-core/src/main/java/com/kmwllc/lucille/core/Spec.java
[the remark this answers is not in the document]
mrwrench24: This is where the configuration of indexers is causing a little bit of confusion / forcing me to jump through some hoops here. There's two different Indexer "configs" to be validated here - the first being `indexer`, and the second being an "indexer-specific" implementation - `solr`, `elastic`, etc. These various indexers don't share any properties.

This method is what is called inside specific indexers (solr, elastic search) to define the legal properties. I figured it would be clearest to call `Spec.indexer()` inside these constructors, and then handle `OPTIONAL_INDEXER_CONFIG_PROPERTIES` in `Indexer` itself. Let me know if you think that makes sense...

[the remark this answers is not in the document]
spencersolomon6: Decision here is to not let this issue block the merging of this PR and wait for Rudi to return for further discussion. We will likely make broader changes to the Indexer and how properties are inherited from the Abstract Indexer. The Stage Specs will then be updated as part of these larger design changes.

spencersolomon6 (approved):

PR #291, opened 2025-04-08
LC-658: Replace JSONata library

mrwrench24:
Replaces `JSONata4Java` w/ `jsonata-java`. It is seen to run 2 - 4x faster and doesn't have the same (minor) shortcomings of `JSONata4Java` that are pointed out in its documentation.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
@@ -1,6 +1,6 @@
 package com.kmwllc.lucille.core;
 
-import com.api.jsonata4java.expressions.Expressions;
+import com.dashjoin.jsonata.Jsonata;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
@@ -490,7 +490,7 @@ public interface Document {
    * @throws DocumentException if any reserved fields are mutated or if the result is not an object (primitive or array)
    * @see <a href="https://github.com/IBM/JSONata4Java">JSONNata implementation</a>
    */
-  void transform(Expressions expr) throws DocumentException;
+  void transform(Jsonata expr) throws DocumentException;
 
   /**
    * Returns an Iterator that contains only this document.
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJSONNata.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJSONNata.java
@@ -1,14 +1,11 @@
 package com.kmwllc.lucille.stage;
 
+import com.dashjoin.jsonata.JException;
+import com.dashjoin.jsonata.Jsonata;
 import com.kmwllc.lucille.core.Spec;
-import java.io.IOException;
 import java.util.Iterator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import com.api.jsonata4java.expressions.EvaluateException;
-import com.api.jsonata4java.expressions.Expressions;
-import com.api.jsonata4java.expressions.ParseException;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.kmwllc.lucille.core.ConfigUtils;
 import com.kmwllc.lucille.core.Document;
 import com.kmwllc.lucille.core.DocumentException;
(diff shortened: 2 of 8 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJSONNata.java
```diff
@@ -1,89 +1,75 @@
...
     }
 
-    JsonNode output = null;
-    try {
-      output = parsed.evaluate(doc.getJson(source));
-    } catch (EvaluateException e) {
-      log.warn("Evaluation exception occurred when applying transformation: ", e);
-      return null;
-    }
+    Object output = parsedExpression.evaluate(doc.getJson(source).toString());
```
kiratraynor, line 72: For a use case of a jsonata expression with functions that point to field names, mapping this toString() will not have the correct functionality.
When it gets to `Jsonata.evaluate`, it will eventually run [`Jsonata.evaluateName`](https://github.com/dashjoin/jsonata-java/blob/main/src/main/java/com/dashjoin/jsonata/Jsonata.java#L684) and run `Functions.lookup` where it will check if the input (the value given to the evaluate method here) is a type `List` or `Map`. If it's neither one of those, it will return null. And since here we're giving it a string, it will always return null for this use case.

Here we can have a try/catch that will convert the source value to a Map and if it cannot, then convert it to a string and evaluate the input whether it be a string object or map object.

lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyJSONNataTest.java
kiratraynor, line 1: I think it would be good to add a test for the use case of an expression that uses jsonata functions. Maybe something like
```
{
  class: "com.kmwllc.lucille.stage.ApplyJSONNata"
  source: "source"
  destination: "dest"
  expression: """
            $exists(field.value) ?
            $number(field.value) : null
          """
}
```
Where the doc contains a nested field "source.field.value" that is something like `"8.2"` and then the resulting "dest" field will be equal to `8.2`.

kiratraynor: Also maybe adding a test similar to this (using some jsonata function), but without a source field would be good to have.

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -490,7 +490,7 @@ default void validateFieldNames(String... names) throws IllegalArgumentException
    * @throws DocumentException if any reserved fields are mutated or if the result is not an object (primitive or array)
    * @see <a href="https://github.com/IBM/JSONata4Java">JSONNata implementation</a>
    */
```
kiratraynor, line 492: Small, but I think we'll want to update this comment as well to link to dashjoin jsonata-java instead like how you did for the `ApplyJSONNata` class.

lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyJSONNataTest.java
mrwrench24: let me know if the test is what you were looking for. I think the first (real) test that was already there might also be what you're looking for (specifically for one that mutates the whole document, it uses the "$keys()" function)

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJSONNata.java
mrwrench24: I was able to resolve this... hopefully it works

lucille-core/src/test/resources/ApplyJSONNataTest/fieldAccessValue.conf
kiratraynor, line 1: The new tests look good! Just this conf file and the next have two `source` fields.

mrwrench24:
One small thing is that the stage class name has a typo - it's JSONata instead of JSONNata. I can change it but don't want to ruin any conf files in use. @kiratraynor let me know what you think

kiratraynor:
Yes I was noticing that as well... I think it could be worth changing the name since it is a typo. It will be a breaking change if anyone is using it but I think it would make sense. Maybe we can bring this to the larger group and see what everyone thinks @mrwrench24.

kiratraynor:
> Yes I was noticing that as well... I think it could be worth changing the name since it is a typo. It will be a breaking change if anyone is using it but I think it would make sense. Maybe we can bring this to the larger group and see what everyone thinks @mrwrench24.

The decision is to rename this state to match the package name, so `ApplyJSONata`.

kiratraynor (approved):

PR #292, opened 2025-04-09
LC-666: Specific Indexer Failures for Solr, Open, and Elastic

mrwrench24:
using some existing POC-style changes, `Indexer.sendToIndex` now returns a `Set<Document>` representing the Documents that were _not_ successfully indexed. Exceptions are only thrown in more exceptional circumstances now. 
Internally, the Indexers maintain a `Map<String, Document>` to easily retrieve the specific Documents that failed based on the IDs returned by the errors (which will be a DocID / ID Override value).

One note w.r.t. the ID Override is that there isn't support for multiple documents with the same ID Override value. I'm not sure if this is an allowed / supported behavior. We could use a List to mitigate this, it would just cause a bit of overhead.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
@@ -275,6 +275,10 @@ public abstract class Indexer implements Runnable {
       failedDocs = Set.of();
     }
 
+    if (!failedDocs.isEmpty()) {
+      log.warn("{} Documents were not indexed successfully.", failedDocs.size());
+    }
+
     // Mark all the documents in failedDoc as failed
     for (Document d : failedDocs) {
       try {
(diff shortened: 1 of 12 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -249,17 +258,42 @@ private void sendToIndexWithAccounting(List<Document> batchedDocs) {
           }
         }
       }
+
+      // We've sent a message for each Document. Don't want to run the code below.
```
rseitz, line 262: the code below is a return which will transfer control to the finally...

spencersolomon6: Small note but it took me a while to understand the intention here because I though "the code below" in this comment was referring to the finally block, which didn't make sense functionally or semantically. I understand the intention now, but it would probably be easier to understand by just adding "We don't want to run the code below the finally block".

kiratraynor (approved):

PR #293, opened 2025-04-09
LC-664 - add parent for lucille-distributed runner and fix other warnings for pom files

kiratraynor:


mrwrench24 (approved):
I'm not really an expert in Maven... but I see the module being listed at the end of the build, so it LGTM :)

spencersolomon6 (approved):

PR #297, opened 2025-04-16
LC-655: Incremental File Connector, V2

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -67,16 +68,29 @@ protected void shutdownStorageClient() throws IOException {
   }
 
   @Override
-  protected void traverseStorageClient(Publisher publisher, TraversalParams params) throws Exception {
+  protected void traverseStorageClient(Publisher publisher, TraversalParams params, StorageClientState state) throws Exception {
     serviceClient.getBlobContainerClient(getBucketOrContainerName(params))
-        .listBlobs(new ListBlobsOptions().setPrefix(getStartingDirectory(params)).setMaxResultsPerPage(maxNumOfPages),
+        .listBlobsByHierarchy("/",
+            new ListBlobsOptions().setPrefix(getStartingDirectory(params)).setMaxResultsPerPage(maxNumOfPages),
```
mrwrench24, line 74: https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-list-java
This allows us to get directories returned as part of list blobs... which is desired now

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/LocalStorageClient.java
```diff
@@ -177,5 +187,14 @@ public Document asDoc(InputStream in, String decompressedFullPathStr, TraversalP
 
       return doc;
     }
+
+    // Directories should include a final '/' to indicate they are directories, similar to the other storage clients
+    private static String getFullPathHelper(Path path, BasicFileAttributes attrs) {
+      if (attrs.isDirectory()) {
+        return path.toAbsolutePath().normalize() + "/";
```
mrwrench24, line 194: normalizing a path to a directory --> it won't end with '/', so we add it here.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/StorageClientStateManager.java
```diff
@@ -0,0 +1,295 @@
...
+    if (!pathToStorage.toString().endsWith("/")) {
+      pathsToProcess.add(pathToStorage + "/");
+    }
+
+    while (!pathsToProcess.isEmpty()) {
+      String currentPath = pathsToProcess.poll();
+
+      // Get a potential SQL entry for this path.
+      String entryQuery = "SELECT * FROM \"" + traversalTableName + "\" WHERE name = '" + currentPath + "'";
+      try (Statement statement = jdbcConnection.createStatement();
```
mrwrench24, line 166: TODO: I am aware of / curious about whether we can avoid "building" state. In other words, we could avoid a (albeit, a minor) hit to scalability (the `Map<String, Instant>`), and instead execute these queries in the `FileConnectorState`. But, I'm thinking this loop here (a "traversal" of the database? ;)) is necessary to help us keep track of deletions - and prevent infinite growth of the database. At some point, we have to see everything we have state for, in order to know what we have to delete.

From my research, a HashMap<String, Instant> with ~500k entries (roughly the size of Enron dataset's documents) will take up ~100 MB of memory. I've found that the map isn't very time consuming to build. In a traversal that usually takes ~6 min, it appears to add ~10-15 seconds of runtime - both for building state and updating the database.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -85,12 +85,17 @@ protected void traverseStorageClient(Publisher publisher, TraversalParams params
...
           resp.contents().forEach(obj -> {
-            S3FileReference fileRef = new S3FileReference(obj);
-            processAndPublishFileIfValid(publisher, fileRef, params);
+            S3FileReference fileRef = new S3FileReference(obj, params);
+            processAndPublishFileIfValid(publisher, fileRef, params, state);
           });
         });
   }
 
+  @Override
```
kwatters, line 94: we should just reuse the same table name for all storage clients.  the file connector should specify where it stores its state so it knows where to find the state database.

mrwrench24:
Sorry for some of the confusion at the meeting yesterday. been a while since I looked at this.

Here's what I remember testing - Lucille users can use H2 (a dependency of Lucille). The connectionString would be something like: jdbc:h2:`{LOCAL_FILE_PATH}` to have the database stored on the file system. (Obviously storing it just in memory isn't enough ;] ). The `FileConnectorStateManager` _will_ create a table (under the configured or default name) if it doesn't exist already.

What this PR _doesn't_ include is some way for you to provide no `state` configuration, and Lucille will then do something like create a `state` directory on the file system, provide a file path (likely something involving the connector name) to store the database at, etc. 

This could be accomplished by adding defaults to all of the remaining required `state` config parameters. Can move forward if needed!

rseitz:
I think it makes sense to go ahead and add defaults to the state config parameters so someone can use this stage without having to specify the database details

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -35,23 +36,24 @@
...
  *
- * <br>
- *
  * <code>filterOptions</code>:
  * <ul>
  *   <li>includes (List&lt;String&gt;, Optional): list of regex patterns to include files.</li>
  *   <li>excludes (List&lt;String&gt;, Optional): list of regex patterns to exclude files.</li>
- *   <li>modificationCutoff (Duration, Optional): Filter files that haven't been modified since a certain amount of time. For example, specify "1h", and only files that were modified more than an hour ago will be published.</li>
+ *   <li>lastModifiedCutoff (Duration, Optional): Filter files that haven't been modified since a certain amount of time. For example, specify "1h", and only files that were modified <b>within</b> the last hour will be published.</li>
+ *   <li>lastPublishedCutoff (Duration, Optional): Filter files that haven't been published by Lucille since a certain amount of time. Relies on your state configuration to determine when files were last published. If you do not include configuration for state, this will have no effect. Specify "1h" to only include / publish files that were last published <b>more</b> than an hour ago.</li>
```
rseitz, line 51: ... files that _were never published before, or_ last published more than an hour ago

doc/site/content/en/docs/Architecture/Components/Connectors/file_connector.md
```diff
@@ -18,22 +18,43 @@ specify the maximum number of files (`maxNumOfPages`) that Lucille will load int
...
+
+### Filter Options
+Filter options determine **which** files will/won't be processed & published in your traversal. All filter options are optional. 
+If you specify multiple filter options, files must comply with all of them to be processed & published.
+* `includes`: A list of patterns for the only file names that you want to include in your traversal.
+* `excludes`: A list of patterns for file names that you want to exclude from your traversal.
+* `lastModifiedCutoff`: Filter out files that haven't been modified recently. For example, specify `"1h"`, and only
+files modified within the last hour will be processed & published.
+* `lastPublishedCutoff`: Filter out files that were recently published by Lucille. For example, specify `"1h"`, and only
+files published by Lucille more than an hour ago will processed & published. Requires you to provide **state** configuration, otherwise,
```
rseitz, line 47: ...only files published by Lucille more than an hour ago _(or never published)_ will be processed...

rseitz (approved):
approving as is. may want a followup PR to add the ability to turn off the marking of files as "encountered" -- with the goal of minimizing the number of SQL statements we have to perform and thereby increasing performance... in the case where the user knows there won't be many deletions in the source system

PR #298, opened 2025-04-22
LC-671: Create RSS Connector

mrwrench24:
Connects to an RSS feed at a given URL, and publishes Documents for each `<item>` found in the feed. The XML elements for the item (author, title, description, guid...) will be added as fields onto the Document, which will then be published. Any elements that aren't found (not all are required for an RSS `<item>`) won't be placed on the Document. 

You can specify `pubDateCutoff`, so only new RSS items will be transformed into Documents and published. For example, `30s` means only RSS items from the last 30 seconds become Documents.

You can specify `refreshIncrement`, allowing the Connector to check the same URL every 5 minutes (or something like that). The Connector will keep track of RSS items it encountered, and will not publish a Document for the same item multiple times. SIGINT is handled gracefully by the Connector (Publisher will still get flushed).

By default, the Connector will use the `guid` of items as the Document ID. If not found, a UUID will be used. You can set `useGuidForDocID` to `false` to use a UUID for all docs. 

It's pretty cool to use this in conjunction with the `ApplyJSoup` Stage and scrape the links you collect from the RSS feed!

There's two tests that mean the unit tests take a bit longer, about 5 seconds.


spencersolomon6 (approved):

mrwrench24:
The recent commit was from something I noticed while working on creating an RSS example. I got one bad "fetch", and it caused everything to be published again, even though there was no content. But there wasn't an exception thrown trying to access the feed.

lucille-core/src/main/java/com/kmwllc/lucille/connector/RSSConnector.java
```diff
@@ -0,0 +1,220 @@
...
+      this.pubDateCutoffDuration = null;
+    }
+
+    this.refreshIncrement = config.hasPath("refreshIncrement") ? config.getDuration("refreshIncrement") : null;
+  }
+
+  @Override
+  public void execute(Publisher publisher) throws ConnectorException {
+    if (refreshIncrement != null) {
+      Signal.handle(
```
rseitz, line 101: Still not sure this makes sense here. This takes effect JVM-wide, I believe.

mrwrench24: Hmmm. Okay. To be honest, I'm not quite sure what the conventional way would be to handle something like this.

Part of the reason I added this -  if I'm remembering correctly - is I was seeing some Documents not get flushed? Maybe I was misunderstanding the console output, but it seemed like keeping the Connector up and running meant there would be one Document that wouldn't make it through the whole run.

mrwrench24: Yes, I think the point was to allow it to run incrementally until you'd want to terminate it. Maybe never. And watch the RSS feed evolve over time. But we could add a natural stopping point in the config - number of iterations, run duration, etc.

lucille-core/src/main/java/com/kmwllc/lucille/connector/RSSConnector.java
```diff
@@ -0,0 +1,222 @@
...
+
+    this.useGuidForDocID = ConfigUtils.getOrDefault(config, "useGuidForDocID", true);
+
+    if (config.hasPath("pubDateCutoff")) {
+      this.pubDateCutoffDuration = config.getDuration("pubDateCutoff");
+    } else {
+      this.pubDateCutoffDuration = null;
+    }
+
+    if (config.hasPath("runDuration") != config.hasPath("refreshIncrement")) {
```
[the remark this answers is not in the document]
mrwrench24: I guess I could see a use case for only running it once. But requiring this to be specified does make sense to me either way (and, would clean up the `execute` method quite a bit). If you only want it to run once, you could specify a short `runDuration`, etc.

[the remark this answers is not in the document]
mrwrench24: Following up from meeting yesterday.

Confirmed that:
1. We want to allow users to run the RSSConnector just once.
2. We do not (yet) want to allow the RSSConnector to run indefinitely. 

The PR supports both. `runDuration` and `refreshIncrement` should remain optional. they have to be specified in tandem in order to run incrementally.

PR #300, opened 2025-04-24
LC-661: Small FileReference Refactor

mrwrench24:
Hold more properties in the `BaseFileReference`, expose a direct way to get file contents, allowing `asDoc` methods to be reused in `BaseFileReference`. Renamed `asDoc` to `decompressedFileAsDoc` to be a bit clearer on the use of the method.

Use a special method for `getFileContent`, returning a `byte[]`, instead of `getContentStream` to avoid 
1. Having to throw an IOException 
2. The overhead associated with opening an InputStream just to immediately exhaust it
instead of just reading the `InputStream` from `getContentStream` which was already available.


agileresearchservices (approved):
It looks clean to me.

PR #302, opened 2025-04-25
LC-674 - Upgrade parquet-avro for CVE

kiratraynor:
Upgrade org.apache.parquet:parquet-avro from 1.15.0 to 1.15.1.

[NVD - CVE-2025-30065](https://nvd.nist.gov/vuln/detail/CVE-2025-30065)

### DIFF
--- a/lucille-plugins/lucille-parquet/pom.xml
+++ b/lucille-plugins/lucille-parquet/pom.xml
@@ -41,6 +41,11 @@
       <version>3.4.1</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.hadoop</groupId>
+      <artifactId>hadoop-mapreduce-client-core</artifactId>
+      <version>3.4.1</version>
+    </dependency>
+    <dependency>
       <groupId>com.amazonaws</groupId>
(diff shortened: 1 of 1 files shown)


lucille-plugins/lucille-parquet/pom.xml
```diff
@@ -33,7 +33,7 @@
     <dependency>
       <groupId>org.apache.parquet</groupId>
       <artifactId>parquet-avro</artifactId>
-      <version>1.15.0</version>
+      <version>1.15.1</version>
```
mrwrench24, line 36: unfortunately, this didn't resolve the Parquet issue I had noticed. This doesn't seem to work with Hadoop 3.4.1, but that is the most recent (Oct. 2024) version of Hadoop-common available. 
However, this upgrade **specifically** is not the cause of the error. Not sure what the best approach forward here would be.

mrwrench24: Calls out to `ParquetFileReader.open` cause:
`java.lang.NoClassDefFoundError: org/apache/hadoop/mapreduce/lib/input/FileInputFormat`

mrwrench24 (approved):

PR #304, opened 2025-04-25
LC-590: FileConnector supports multiple paths

mrwrench24:
`pathToStorage` becomes `pathsToStorage`. The `FileConnector` contains a map of `StorageClient`s that it will use, as appropriate, for each path. You cannot specify multiple `pathsToStorage` and either `moveToAfterProcessing` or `moveToErrorFolder`.

I think the biggest open question about this PR is whether we want to throw an Exception if one traversal fails.


lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -127,42 +129,58 @@ public FileConnector(Config config) throws ConnectorException {
...
+          if (storageClient == null) {
+            log.warn("No StorageClient was available for {}, the path will be skipped. Did you include the necessary configuration?", pathToTraverse);
+            continue;
+          }
+
+          TraversalParams params = new TraversalParams(pathToTraverse, getDocIdPrefix(), fileOptions, filterOptions);
+          storageClient.traverse(publisher, params);
+        } catch (Exception e) {
+          throw new ConnectorException("Error occurred while initializing client or publishing files.", e);
+        }
```
mrwrench24, line 175: It's a small detail, but wondering if we want to throw an exception if one path fails. I think the reason I'm leaning towards this is that, if an exception is thrown, the `StorageClient` itself could be in an unusable state, so if we are working with all Google Cloud paths, or something, this might not be very helpful.

agileresearchservices (approved):
Looks ok to me. But, this one is heavier. I'd rather Spencer or Kira take an additional look just in case.

meghanboyd:
Decision: if you specify a moveTo directory there can be only one pathToStorage. Additionally, the storage provider must be consistent for moveTo & pathToStorage.

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -158,42 +160,59 @@ public FileConnector(Config config) throws ConnectorException {
...
-      storageClient.traverse(publisher, params);
-    } catch (Exception e) {
-      throw new ConnectorException("Error occurred while initializing client or publishing files.", e);
+      for (URI pathToTraverse : storageURIs) {
+        String clientKey = pathToTraverse.getScheme() != null ? pathToTraverse.getScheme() : "file";
+        StorageClient storageClient = storageClientMap.get(clientKey);
+
+        if (storageClient == null) {
+          log.warn("No StorageClient was available for {}, the path will be skipped. Did you include the necessary configuration?", pathToTraverse);
+          continue;
```
mrwrench24, line 198: Kind of stems from the open question about if we want one exception to throw everything off. But this is a different and certainly could warrant an actual Exception being thrown here.

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
rseitz: I think it's ok to throw this exception
each traversal should be optimistic
so, if any traversal does throw an exception, that indicates a "significant" problem and we should bubble it up

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -158,42 +158,63 @@ public FileConnector(Config config) throws ConnectorException {
...
     }
   }
 
   @Override
   public void execute(Publisher publisher) throws ConnectorException {
     try {
-      storageClient = StorageClient.create(storageURI, config);
-    } catch (Exception e) {
-      throw new ConnectorException("Error occurred while creating storage client.", e);
+      for (StorageClient client : storageClientMap.values()) {
```
rseitz, line 187: should this be included in the try/finally below?
as it stands now, we might init some clients, then encounter an exception on a subsequent client and never shutdown any of them

rseitz (changes requested):
posted comment about shutting clients down after a failed init; otherwise looks good

PR #305, opened 2025-04-28
LC-672: Specific Indexer Failures in Weaviate.

mrwrench24:
Unfortunately, I also looked into adding this to Pinecone, but the API doesn't expose enough information to do so.


agileresearchservices (approved):
Looks fine to me.

PR #306, opened 2025-04-29
LC-411: "Catch All" + Other Validation Improvements

mrwrench24:


lucille-core/src/test/resources/ConfigValidationTest/badConnector.conf
```diff
@@ -19,19 +19,19 @@ pipelines: [
     stages: [
```
mrwrench24, line 19: updating this file to only include issues with the connector, nothing else.

agileresearchservices (commented):
Since this one is a little heavy, I'd like to defer to Kira/Spencer/KW on this one.

lucille-core/src/test/resources/ConfigValidationTest/badRunnerConfig.conf
```diff
@@ -0,0 +1,66 @@
...
+  {
+    name: "connector2",
+    class: "com.kmwllc.lucille.connector.CSVConnector",
+    pipeline: "pipeline2",
+    idField: "id",
+    path: "classpath:RunnerTest/singleDoc.csv",
+  }
+]
+
+pipelines: [
```
rseitz, line 18: If this file is only to test a bad runner config, can we make the rest of the config simpler, like one connector with one pipeline that has one stage? Same comment for some of the other tests files that target one specific issue...

rseitz (approved):
approving with comment about test configs

PR #308, opened 2025-05-01
LC-597: Test for acceptInvalidCert in SolrIndexer + Patched Resource Leak

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
```diff
@@ -68,25 +63,40 @@ private static CloudHttp2SolrClient getCloudClient(Config config) {
...
   /**
    * Generates a HttpClient with preemptive authentication if required.
+   * This method has SIDE EFFECTS. It will set SSL system properties if corresponding properties
+   * are found in the config.
    *
    * @param config The configuration file to generate the HttpClient from.
    * @return the HttpClient
    */
-  public static Http2SolrClient getHttpClient(Config config) {
+  static Http2SolrClient getHttpClientAndSetCheckPeerName(Config config) {
```
spencersolomon6, line 99: Any specific reason for making these implicitly public?

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
mrwrench24: We didn't have any usages outside of `SolrUtils` itself. But we did have a unit test calling the method directly that I didn't want to delete, so it's just package access

PR #309, opened 2025-05-05
LC-562: Support File Moves + Deletes on Cloud Storage

mrwrench24:
`fileOptions.moveToAfterProcessing` and `fileOptions.moveToErrorFolder` now work for cloud storage. You can specify a URI to a directory you want to move files to. As added in LC-590, you can only use these file options with one path in `pathsToStorage`.

As part of this ticket, `FileReference.getFullPath()` now returns a `URI` to the file. Internally, all `BaseFileReference`s must be constructed with a URI to the file being referenced. Some miscellaneous changes in `BaseStorageClient` were made relating to this change.

There is slightly different behavior when moving cloud files compared to local files to make it a bit easier / more feasible to use. Since directories aren't "real" in cloud storage, we have to prevent duplicate keys. As such, when a cloud file moves, its key moves along with it.

**Example:**
Let's say we are traversing `s3://bucket/files/`. 
And in our Config, `fileOptions.moveToAfterProcessing` = `s3://bucket/after/`.
When we encounter & publish `s3://bucket/files/file1.txt`, the "moved file" will be `s3://bucket/after/files/file1.txt` - since the "key" of the original file is `files/file1.txt`.
If we didn't do it this way, working on something like the Enron dataset wouldn't support `moveToAfterProcessing`, since many of the emails are in the format "1.txt", "2.txt", ... etc. We'd very quickly get a duplicate key here.

I know this is a very debatable design decision and probably something to discuss further.


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/StorageClient.java
```diff
@@ -42,6 +42,14 @@ public interface StorageClient {
    */
   InputStream getFileContentStream(URI uri) throws IOException;
 
+  /**
+   * Moves the file at the given String to the folder at the given URI.
+   * @param filePath The full path to the file that you want to move.
+   * @param folder A URI to the folder that you want to move the file to.
+   * @throws IOException If an error occurs moving the file.
+   */
+  void moveFile(URI filePath, URI folder) throws IOException;
```
mrwrench24, line 51: Doesn't have to be a fully exposed method - could just be an abstract method in `BaseStorageClient`.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/AzureStorageClient.java
```diff
@@ -95,6 +92,28 @@ protected InputStream getFileContentStreamFromStorage(URI uri) throws IOExceptio
...
+      destBlob = folder.getPath().split("/")[2] + "/" + sourceBlob;
+    } else {
+      destBlob = sourceBlob;
+    }
+
+    BlobClient sourceBlobClient = serviceClient.getBlobContainerClient(sourceContainer).getBlobClient(sourceBlob);
+    BlobClient destBlobClient = serviceClient.getBlobContainerClient(destContainer).getBlobClient(destBlob);
+
+    // Starts the copy in the background
+    sourceBlobClient.beginCopy(destBlobClient.getBlobUrl(), null);
```
mrwrench24, line 114: just pointing out that the cloud clients will run these in the background. so they'll get completed, but not necessarily before the next connector begins running etc.

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/StorageClient.java
rseitz: I think it's ok to leave it here on StorageClient

lucille-core/src/test/java/com/kmwllc/lucille/connector/storageclient/GoogleStorageClientTest.java
```diff
@@ -352,43 +353,123 @@ public void testPublishOnCompressedAndArchived() throws Exception {
...
+    storage.create(blobInfo4, "foo".getBytes());
+
+    googleStorageClient.setStorageForTesting(storage);
+    googleStorageClient.initializeForTesting();
+
+    googleStorageClient.traverse(publisher, params);
+
+    // the moveFile() method creates a thread for handling the copy and delete... have to give just a bit of time for those
+    // threads to run before running the assertions below.
+    Thread.sleep(250);
```
rseitz, line 392: can remove this once we remove the thread creation

lucille-core/src/test/java/com/kmwllc/lucille/connector/storageclient/GoogleStorageClientTest.java
```diff
@@ -352,43 +353,123 @@ public void testPublishOnCompressedAndArchived() throws Exception {
...
     googleStorageClient.setStorageForTesting(storage);
     googleStorageClient.initializeForTesting();
-    assertThrows(UnsupportedOperationException.class, () -> googleStorageClient.traverse(publisher, params));
 
-    // closes storage too
+    googleStorageClient.traverse(publisher, params);
+
+    // the moveFile() method creates a thread for handling the copy and delete... have to give just a bit of time for those
+    // threads to run before running the assertions below.
+    Thread.sleep(250);
```
rseitz, line 452: can remove this once we remove thread creation

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -104,6 +104,28 @@ protected InputStream getFileContentStreamFromStorage(URI uri) throws IOExceptio
...
+    CopyObjectRequest copyRequest = CopyObjectRequest.builder()
+        .sourceBucket(sourceBucket).sourceKey(sourceKey)
+        .destinationBucket(destBucket).destinationKey(destKey)
+        .build();
+
+    DeleteObjectRequest deleteRequest = DeleteObjectRequest.builder()
+        .bucket(sourceBucket).key(sourceKey)
+        .build();
+
+    // using the async client to prevent a slowdown. allow the object to copy, and then delete it, all async.
```
rseitz, line 124: stale comment -- references "async"

PR #312, opened 2025-05-07
LC-659: POC for OpenSearch Log4j Appender

mrwrench24:


agileresearchservices (commented):
So this goes from local opensearch, through Lucille, back to local opensearch?

mrwrench24:
@agileresearchservices sorry for never responding.

This just automatically ingests the JSON-ified Lucille logs into an Opensearch Instance without any manual intervention.

meghanboyd:
@mrwrench24 can we please add code comment explaining how to enable this feature? We want it off by default.

lucille-core/pom.xml
```diff
@@ -380,6 +380,12 @@
       <artifactId>rssreader</artifactId>
       <version>3.9.3</version>
     </dependency>
+
+    <dependency>
```
rseitz, line 384: Should this dependency be commented out with a note to uncomment if OpenSearchAppender is also uncommented in log4j.xml?

rseitz (approved):

PR #313, opened 2025-05-09
LC-680 Create Lucille APIs to list connectors, stages and indexers

supertick:
## Enhance Pipeline Component Discoverability and Configuration Typing

This PR introduces new APIs for discovering pipeline components (Connectors, Stages, Indexers) and refactors component configuration to use static typed inner classes, improving type safety and maintainability. These changes are foundational for building a user-friendly UI or wizard to simplify pipeline creation.

**Key Changes:**

1.  **Reflective Component Listing APIs:**
    *   Added three new API endpoints that leverage reflection to dynamically list available Connectors, Stages, and Indexers from the classpath.

2.  **`JsonDoclet` for Enhanced Metadata:**
    *   Implemented a `JsonDoclet` to generate structured Javadoc information (including class and configuration parameter descriptions) as a JSON file during CI/CD. This JSON can be queried at runtime to provide rich, contextual information for each component.

3.  **Static Typed Inner Configuration Classes:**
    *   Initiated a pattern of using static typed inner classes for component configuration. For example, the `AddRandomBoolean` stage now contains an `AddRandomBooleanConfig` inner class.
    *   **Benefits of this approach:**
        *   **Type Safety:** Configuration values are parsed into strongly-typed fields (e.g., `String`, `int`), catching type mismatches at compile time.
        *   **Readability & Maintainability:** Clearly defines accepted parameters, their types, and default values directly within the component's structure.
        *   **IDE Support:** Enables autocompletion and safer refactoring of configuration keys.
        *   **Centralized Validation:** Allows for validation logic (e.g., range checks for `percentTrue` in `AddRandomBooleanConfig`) to be encapsulated within the config class itself.
        *   **Testability:** Configuration parsing and validation can be unit-tested in isolation.
        *   **Self-Documentation:** The config class structure inherently documents the component's settings.

**Impact:**

*   Enables dynamic population of UI elements with available pipeline components and their configurations.
*   Provides detailed descriptions and strongly-typed configuration options for each component, facilitating easier, more informed, and less error-prone pipeline design.
*   Improves the overall robustness and developer experience when working with component configurations.


lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomBoolean.java
```diff
@@ -20,24 +21,45 @@
...
+    super(configIn, Spec.stage()
         .withOptionalProperties("field_name", "percent_true"));
 
-    this.fieldName = ConfigUtils.getOrDefault(config, "field_name", "data");
-    this.percentTrue = ConfigUtils.getOrDefault(config, "percent_true", 50);
+    config = new AddRandomBooleanConfig();
+    config.apply(configIn);
 
-    if (percentTrue > 100 || percentTrue < 0) {
+    if (config.percentTrue > 100 || config.percentTrue < 0) {
```
kwatters, line 55: this would be handled by conifg.validate()   , this logic moved into the stage config.

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomBoolean.java
```diff
@@ -20,24 +21,45 @@
...
 
-  private final String fieldName;
-  private final int percentTrue;
+  public static class AddRandomBooleanConfig extends BaseStageConfig {
+
+    private String fieldName = "data";
+    private int percentTrue = 50;
+
+    public void apply(Config config) {
+      fieldName = ConfigUtils.getOrDefault(config, "field_name", fieldName);
```
kwatters, line 30: should this call super.apply(config) first?

gperryatnasuni: Yes, was what I was thinking too - I'll put it in

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomBoolean.java
gperryatnasuni: Yeah, I was trying to determine how validation would work for an "updated" class - initially, I was on the safe side and having validation in both - but yeah, it should be updated to only validate in the Inner Class

lucille-core/src/main/java/com/kmwllc/lucille/doclet/JsonDoclet.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        return true;
+    }
+
+    /**
+     * Main method to run JsonDoclet as a standalone application.
+     * Uses sensible defaults for source path, package, and output file.
+     */
+    public static void main(String[] args) throws Exception {
```
kwatters, line 128: we should probably move this to a unit test instead of having a main method here.

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomBoolean.java
```diff
@@ -20,24 +21,45 @@
...
+    public int getPercentTrue() { return percentTrue; }
+
+  }
+
+  AddRandomBooleanConfig config;
 
-  public AddRandomBoolean(Config config) throws StageException {
-    super(config, Spec.stage()
+  public AddRandomBoolean(Config configIn) throws StageException {
+    super(configIn, Spec.stage()
```
kwatters, line 49: it'd be nice if we could create the stage spec from the stageconfig class so it could become boiler plate instead of having to define the required/optional fields.

meghanboyd:
To do:

- make static config for sequenceConnector
- annotations for required parameters
- keep both APIs working

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomBoolean.java
supertick: it does now

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomBoolean.java
supertick: yeah, i temporarily left it in to show that validation using Spec wouldn't interfere with validation() on the static class .. removed now

lucille-core/src/main/java/com/kmwllc/lucille/doclet/JsonDoclet.java
supertick: Not sure if deriving your own doclet requires a main - the javadoc tool requires it .. I'll dig into it

lucille-core/src/main/java/com/kmwllc/lucille/core/BaseConfig.java
```diff
@@ -1,5 +1,40 @@
...
     
+    private String name;
+    private String className;
+    private String[] conditions;
+    private String conditionPolicy;
+
+    public void apply(Config config) {
+        name = ConfigUtils.getOrDefault(config, "name", null);
+        className = ConfigUtils.getOrDefault(config, "class", null);
+        conditionPolicy = ConfigUtils.getOrDefault(config, "conditionPolicy", null);
```
mrwrench24, line 15: These are specific to stages... I think they might want to be in `BaseStageConfig` again. Unless I'm misunderstanding some of the changes here.

lucille-core/src/main/java/com/kmwllc/lucille/core/BaseConfig.java
```diff
@@ -1,5 +1,40 @@
...
+    }
+
+    // TODO - add validation - classname and name are not required?
+    public void validate() throws Exception {
+//        if (name == null || className == null) {
+//            throw new StageException("name and className must be specified");
+//        }
```
mrwrench24, line 22: I think part of the reason they are optional (specifically, `class`) is so we don't have to include them in our unit test Configs. Maybe a sign of tech debt. But, when you run validation, it makes sense that (for `class`, at least), you'll see only "missing class", and not other errors, since it's not clear what the properties would be without a Stage class specified.

lucille-core/src/main/java/com/kmwllc/lucille/core/ConnectorException.java
```diff
@@ -1,6 +1,6 @@
 package com.kmwllc.lucille.core;
 
-public class ConnectorException extends Exception {
+public class ConnectorException extends BaseConfigException {
```
mrwrench24, line 3: `ConnectorException`s can be thrown for non-Config related causes. (e.g. something bad happens during a `FileConnector` traversal)

Same for Indexer and Stage

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -66,76 +67,182 @@ public class QueryOpensearch extends Stage {
 
   private static final Logger log = LoggerFactory.getLogger(QueryOpensearch.class);
   private static final ObjectMapper mapper = new ObjectMapper();
-  private static final HttpClient httpClient = HttpClient.newHttpClient();
```
mrwrench24, line 69: Not sure if the history got a bit messed up here, but this was from a branch I merged in yesterday. So it probably showed up after merging in main.

PR #315, opened 2025-05-09
LC-545: Safe Singletons 

mrwrench24:
1. `DictionaryManager` / "dictionaries" in general use a `List<String>` instead of `String[]` to prevent the contents from being modified.
2. `QueryOpensearch` has a shared `HttpClient` to improve its scalability. (This is in line with the recommended usage of `HttpClient`, evidently. It is thread safe.) The issue here would be running a Pipeline that uses `QueryOpensearch` - potentially multiple times - either on multiple threads or in multiple concurrent runs. Each instance of the Stage in each Thread/Run will open an HTTPClient, and since this is done _at the same time_, it'll cause issues.


agileresearchservices (approved):
Looks fine to me

PR #316, opened 2025-05-14
LC-250: Update Elasticsearch Client

mrwrench24:
- Removes the old dependency on `elasticsearch-rest-high-level-client`.
- Added missing `acceptInvalidCert` optional property 
- upgrades `Elasticsearch-clients` to `9.0.1`.
- `ElasticSearchUtils` uses the new `Rest5Client` instead of the legacy `RestClient`.
- HttpCore/Client have updated versions


meghanboyd:
We want to make sure index/bulk api is backwards compatible, make sure we can index back to elastic 8.17 (most recent 8.x series) with updated client

PR #320, opened 2025-05-20
LC-620: Improved Test Coverage in Lucille-Core/util

mrwrench24:
Old Coverage Levels: 91% / 84%
New Coverage Levels: 95% / 95%

Some notes about this PR:
- It includes the changes from LC-597 since that had added tests. 
- In addition to improved test coverage, there are some very small code changes to point out.
1. `FileUtils.isValidURI` - doesn't check whether the scheme is empty. To my understanding, this isn't possible.
2. Resulting from the previous change, `FileContentFetcher` doesn't check if a valid URI has a non-null scheme.

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/util/FileUtilsTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/util/FileUtilsTest.java
@@ -1,11 +1,41 @@
 package com.kmwllc.lucille.util;
 
+import java.nio.file.Paths;
 import org.junit.Test;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 public class FileUtilsTest {
+
+  @Test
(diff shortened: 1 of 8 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/util/FileUtilsTest.java
```diff
@@ -1,12 +1,41 @@
...
+    assertTrue(Paths.get(lucilleHome).isAbsolute());
+    assertTrue(Paths.get(lucilleHome).endsWith("lucille-core"));
+
+    // 2. Getting it when it is not set, falls back to user.dir
+    System.clearProperty("LUCILLE_HOME");
+    System.setProperty("user.dir", "/test/folder");
+    String homeDir = FileUtils.getLucilleHomeDirectory();
+    assertEquals("/test/folder", homeDir);
+
+    if (oldLucilleHome != null) {
```
rseitz, line 30: suggest putting this in a finally block

PR #321, opened 2025-05-22
LC-681 New stage for executing a python script

supertick:


lucille-core/src/main/java/com/kmwllc/lucille/stage/PythonStage.java
```diff
@@ -0,0 +1,379 @@
...
+        }
+
+        for (Map.Entry<String, Object> entry : updatedFields.entrySet()) {
+          if (!reserved.contains(entry.getKey())) {
+            doc.setField(entry.getKey(), entry.getValue());
+          }
+        }
+      }
+      doc.setField("python_stage_executed", true);
+      return null;
```
kwatters, line 299: normally , this return value is an iterator to any children documents returned by the stage.   this implies that this stage can not emit children documents.

lucille-core/src/main/java/com/kmwllc/lucille/stage/PythonStage.java
```diff
@@ -0,0 +1,565 @@
...
+  private final String pythonExecutable;
+  private final String functionName;
+  private final int port;
+  private GatewayServer gateway;
+  private final AtomicBoolean started = new AtomicBoolean(false);
+  private transient Process pythonProcess; // Transient handle to the launched Python process
+  private Py4jClient py4jClient = null;
+  private volatile boolean ready = false;
+  private ObjectMapper mapper = new ObjectMapper();
+  private transient Py4jExecutorInterface handler;
```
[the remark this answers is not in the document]
supertick: I thought it might be serialized at some point to provide state and configuration of a stage to the UI :shrug: 
Not sure, if so more fields will need to be transient.

lucille-core/src/main/java/com/kmwllc/lucille/stage/PythonStage.java
```diff
@@ -0,0 +1,565 @@
...
+        Thread.sleep(500);
+        log.info("Waiting for Py4J connection to be ready");
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new StageException("Interrupted while waiting for Py4J connection to be ready", e);
+      }
+    }
+    if (!ready) {
+      throw new StageException("Py4J connection was not established within 5 seconds");
+    }
```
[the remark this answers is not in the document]
mrwrench24: That makes sense. I guess I had just imagined the `processDocument` pulling the strings on the `Py4JClient` at first glance. (And maybe more).

lucille-core/src/main/java/com/kmwllc/lucille/stage/PythonStage.java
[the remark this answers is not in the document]
mrwrench24: Oh sorry. Didn't know that was something we were planning to do (at least w.r.t. state)!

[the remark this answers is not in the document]
rseitz: we haven't been marking stage instance variables transient so far as we haven't planned to serialize stages. If we're going to support serialization there will be significant work to do (including tests for serialization/deserialization of each stage) and I think we can wait on making fields transient until that work begins. For now, it would just be confusing for a few stages to use this keyword while others don't (even though they have nonserializable members).

lucille-core/src/main/java/com/kmwllc/lucille/stage/PythonStage.java
```diff
@@ -0,0 +1,666 @@
...
+  /**
+   * Waits for the Python process to terminate and logs its exit code.
+   * Notifies the parent PythonStage when the process ends.
+   */
+  public class WaitForProcess extends Thread {
+    public Integer exitCode;
+    public Process process;
+    public PythonStage pythonStage;
+
+    public WaitForProcess(PythonStage pythonStage, Process process) {
```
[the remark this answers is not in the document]
rseitz: This should be a standalone class. If it's going to be an inner class it should be static because it doesn't rely on the state of the enclosing class instance aside for the name (which could be passed in the constructor). But since this Stage is quite long and has other inner classes, it's best to break them out.

lucille-core/src/main/java/com/kmwllc/lucille/stage/util/StreamGobbler.java
```diff
@@ -0,0 +1,60 @@
...
+    this.name = name;
+  }
+
+  @Override
+  public void run() {
+    try {
+      InputStreamReader isr = new InputStreamReader(processOut);
+      BufferedReader br = new BufferedReader(isr);
+      String line = null;
+      while ((line = br.readLine()) != null) {
```
rseitz, line 53: where do we close the bufferedreader and the inputstreamreader?

lucille-core/src/main/java/com/kmwllc/lucille/stage/util/StreamGobbler.java
```diff
@@ -0,0 +1,60 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A general stream gobbler, useful when starting processes and handling their streams
+ */
+public class StreamGobbler extends Thread {
+  protected transient InputStream processOut;
+  protected transient OutputStream processIn;
+  protected String name;
```
rseitz, line 18: where is this name used? threads have names, and we're extending thread, so can call getName() if we need the name that was passed into the superconstructor

lucille-core/src/main/java/com/kmwllc/lucille/stage/util/Py4JRuntime.java
```diff
@@ -0,0 +1,422 @@
...
+      }
+      sysPathProc.waitFor();
+      log.info("Python sys.path entries:\n{}", sysPaths.toString().trim());
+    } catch (Exception e) {
+      log.error("Error logging Python environment", e);
+      throw new StageException("Failed to log Python environment: " + e.getMessage(), e);
+    }
+  }
+
+  private static int allocatePort(int configured) throws StageException {
```
rseitz: suggest just making this `private static synchronized` and removing the internal sychronized blocks for simplicity. Same with releasePort(). Can also remove `private static final Object portLock`
These methods are only called once upon stage startup so we do not need to optimize them for concurrency

lucille-core/src/test/java/com/kmwllc/lucille/stage/PythonStageTest.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+          for (int j = 0; j < numDocsPerThread; j++) {
+            stage.processDocument(docs[idx]);
+          }
+        } catch (Exception e) {
+          throw new RuntimeException(e);
+        }
+      });
+    }
+    for (Thread t : threads) t.start();
```
rseitz: put braces around statement body

lucille-core/src/main/java/com/kmwllc/lucille/stage/util/Py4jExecutorInterface.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Interface for executing Python code from Java via Py4J.
+ * <p>
+ * Implementations of this interface are expected to be exposed as entry points to the Py4J GatewayServer
+ * and provide a method for Java to send JSON-encoded messages or commands to Python for execution.
+ * The typical use case is to pass a JSON string representing a method call and its arguments, which the Python
+ * side will decode, dispatch, and execute, returning the result (often as a JSON string) back to Java.
+ * </p>
+ */
+public interface Py4jExecutorInterface  {
```
rseitz: no longer used; remove

lucille-core/src/test/java/com/kmwllc/lucille/stage/PythonStageTest.java
```diff
@@ -0,0 +1,90 @@
...
+  public void testPythonUpdateToDocMultiThreaded() throws Exception {
+    String confPath = "PythonStageTest/process_document_1.conf";
+    stage = factory.get(confPath);
+    int numThreads = 5;
+    int numDocsPerThread = 3;
+    Thread[] threads = new Thread[numThreads];
+    Document[] docs = new Document[numThreads];
+    for (int i = 0; i < numThreads; i++) {
+      final int idx = i;
+      docs[i] = Document.create("doc_mt_" + idx);
```
rseitz: looks like we're creating one doc per thread and running that same doc through processDocument multiple times in each thread... 
would seem more natural to create numThread*numDocsPerThread docs and not recycle them

lucille-core/src/main/java/com/kmwllc/lucille/stage/PythonStage.java
```diff
@@ -0,0 +1,128 @@
...
+ *   the virtual environment. Defaults to python3.</li>
+ *   <li>requirementsPath (String, Optional) : Path to a requirements.txt file whose dependencies will be
+ *   installed into the managed virtual environment before the Python script is started.</li>
+ *   <li>functionName (String, Optional) : Name of the Python function to call for each document. Defaults to
+ *   process_document.</li>
+ *   <li>port (Integer, Optional) : Explicit base port to use for the Py4J gateway. If omitted, a free port
+ *   pair is automatically selected starting from the default range.</li>
+ * </ul>
+ */
+public final class PythonStage extends Stage {
```
rseitz: Should rename to Py4JStage because we're also going to have a GraalPyStage

PR #324, opened 2025-05-28
LC-683: Custom File Handlers

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/BaseFileHandler.java
```diff
@@ -21,7 +23,9 @@ public abstract class BaseFileHandler implements FileHandler {
 
   protected String docIdPrefix;
```
rseitz, line 24: can this be final? do we ever need to modify it in a subclass?

lucille-core/src/main/java/com/kmwllc/lucille/connector/CSVConnector.java
```diff
@@ -37,7 +37,13 @@ public CSVConnector(Config config) {
             "idField", "docIdFormat", "separatorChar", "useTabs", "interpretQuotes", "ignoreEscapeChar", "lowercaseFields", "ignoredTerms"));
 
     this.pathStr = config.getString("path");
-    this.csvFileHandler = new CSVFileHandler(config);
+    this.csvFileHandler = new CSVFileHandler(config
```
rseitz, line 40: this approach of removing properties from the connector config to pass it to filehandler seems a little brittle (needs to be updated anytime the connector gains a new property that the filehandler doesn't recognize)

but 1) there doesn't seem to be a concise way to copy a subset of config properties into a new config, 2) connectors are basically deprecated at this point... so I think this can stay as is. This comment is just a note.

PR #325, opened 2025-05-30
LC-622: Improved tests for AddRandomString, ExtractEntities, ApplyRegex, and Print

mrwrench24:


lucille-core/src/test/java/com/kmwllc/lucille/stage/AddRandomStringTest.java
```diff
@@ -208,7 +208,38 @@ public void testNestedWithFile() throws StageException {
...
+
+    // defaults to min/max of 1
+    Document doc = Document.create("doc");
+
+    stage.processDocument(doc);
+
+    assertTrue(doc.has("data"));
+    int data = doc.getInt("data");
+    // lower is inclusive, upper is exclusive
+    assertTrue(data >= 0 && data < 100);
```
rseitz, line 222: the name of this test method suggests we're testing how many terms are generated -- should be one -- but where are we asserting that? I suppose doc.getInt would fail on a string that doesn't represent a single int... but the test would be clearer if we did doc.getString() and looked at the string as such

lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyRegexTest.java
```diff
@@ -62,6 +64,110 @@ public void testCapturingGroup() throws Exception {
...
   }
 
+  @Test
+  public void testIgnoreCase() throws Exception {
+    Stage stage = factory.get("ApplyRegexTest/ignoreCase.conf");
+
+    Document doc1 = Document.create("1");
+    doc1.setField("input", "Hi abcdef");
+
+    Document doc2 = Document.create("1");
```
rseitz, line 74: I know this test doesn't look at doc ID, but I'd suggest giving different docs different IDs (so doc2 would get "2")... same comment for other tests here.

lucille-core/src/test/java/com/kmwllc/lucille/stage/PrintTest.java
```diff
@@ -129,4 +129,41 @@ public void testAppendThrows() throws StageException {
...
   }
+
+  @Test
+  public void testBadOutputPath() {
+    assertThrows(StageException.class, () -> factory.get("PrintTest/badOutputPath.conf"));
+  }
+
+  @Test
+  public void testOverwriting() throws Exception {
+    Stage overwrite = factory.get("PrintTest/outputFileOverwrite.conf");
```
rseitz, line 140: suggest changing variable name to "stage" or "printStage"

lucille-core/src/test/java/com/kmwllc/lucille/stage/AddRandomStringTest.java
rseitz: after we do doc.getString() can we do something like a split the string on " " and assert there's only 1 element found... to be explicit about what we're looking for in this test?

rseitz (approved):

PR #327, opened 2025-06-02
LC-498: Print "AppendThreadNames" follow-up

mrwrench24:
- JavaDocs warning about the importance of `appendThreadName` in a multithreaded run
- `appendThreadName` defaults to `true`

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/Print.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/Print.java
@@ -29,7 +29,7 @@ import java.util.List;
  * </ul>
  *
  * <p> <b>Note:</b> If you have multiple worker threads, it is highly recommended that you keep <code>appendThreadName</code> enabled
- * to ensure thread-safe file writes and prevent data corruption.
+ * to ensure thread-safe file writes and prevent data corruption - <i>especially</i> if <code>overwriteFile</code> is enabled.
  */
 public class Print extends Stage {
 
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/Print.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/Print.java
@@ -7,14 +7,13 @@ import com.kmwllc.lucille.core.Stage;
 import com.kmwllc.lucille.core.StageException;
 import com.kmwllc.lucille.util.FileUtils;
 import com.typesafe.config.Config;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.BufferedWriter;
 import java.io.FileWriter;
 import java.io.IOException;
-import java.io.Writer;
-import java.lang.invoke.MethodHandles;
 import java.util.Iterator;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


rseitz (approved):

mrwrench24:
just going to do a sanity check that the old changes didn't work as I suspected they wouldn't. the new "thread local" changes are working and did a manual test w/ 5000 sequenceconnector docs and 10 worker threads.

rseitz (changes requested):
posted comment about the non-static queue

lucille-core/src/main/java/com/kmwllc/lucille/stage/Print.java
```diff
@@ -25,20 +24,26 @@
...
+  private final boolean overwriteFile;
   private final boolean appendThreadName;
+  private final List<String> excludeFields;
 
-  private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
-  private Writer writer = null;
-  private List<String> excludeFields;
+  // multithreaded support for handling appendThreadName + multithreaded runs involving an outputFile.
+  private ThreadLocal<BufferedWriter> threadLocalWriter;
+  private final ConcurrentLinkedQueue<BufferedWriter> allWriters = new ConcurrentLinkedQueue<>();
```
rseitz, line 46: lost this comment somehow on first attempt...
allWriters is not static, so each stage instance will have it's own ConcurrentLinkedQueue -- is that the intent?

mrwrench24: yes... we still need a way to close all the FileWriters we create as part of `ThreadLocal`. There didn't appear to be any way to just close all the ones created (in `threadLocalWriter`) in `stop()`. So we hold a stage-specific List (Queue) of the FileWriters created by all the threads, and then close all of them once stop() is called.

mrwrench24:
as I suspected the old version (without `ThreadLocal`) did not use `appendThreadName` correctly.

lucille-core/src/main/java/com/kmwllc/lucille/stage/Print.java
rseitz: each worker thread has its own instance of this stage which creates its own queue to which it adds its own writer... how would any of these allWriters intances every have more than one writer in it?

mrwrench24: ahhhhh... I got it a little mixed up here. construction and start still takes place in the main thread but there is still one Print stage being created/started per thread. will have to figure out a nice way to address this w.r.t. adding thread names / using a different fileWriter

mrwrench24:
Pushed an updated version after being reminded each thread gets its own pipeline instance. The writer gets created in `start()` if `appendThreadName: false`. Otherwise it'll be created in `processDocument`, if it hasn't been already.
We could also use some form of a lazy variable which would get us this "automatically" but I didn't know if that would just make it a bit more confusing.

lucille-core/src/main/java/com/kmwllc/lucille/stage/Print.java
```diff
@@ -48,30 +52,38 @@ public Print(Config config) {
...
+    // create the file writer, if needed, if we aren't using appendThreadName
+    if (outputFilePath != null && !appendThreadName) {
       try {
-        if (appendThreadName) {
-          String appendedOutputFilePath = FileUtils.appendStringToFileName(outputFilePath, Thread.currentThread().getName());
-          writer = new BufferedWriter(new FileWriter(appendedOutputFilePath, !overwriteFile));
-        } else {
-          writer = new BufferedWriter(new FileWriter(outputFilePath, !overwriteFile));
-        }
+        this.writer = new BufferedWriter(new FileWriter(outputFilePath, !overwriteFile));
```
rseitz, line 67: consider moving this into processDocument so the writer is always created in the same place regardless of whether we're using appendThreadName. Since we're forced to _sometimes_ do it in processDocument it might be cleaner to _always_ do it there

rseitz (approved):

PR #328, opened 2025-06-03
LC-678: Support for attributes & qualifiers in xmlIdPath for XMLFileHandler

mrwrench24:
I know there was a bit more refinement to do on this ticket, so I'm opening this PR with the knowledge that we might want to make some additional changes here. :)

Improves the `xmlIdPath` in the `XMLFileHandler`. You can now specify an attribute in the path. You can now specify a qualifier in the path. 

The `ChunkingXMLHandler` uses `xpath` to evaluate the `xmlIdPath`. A new parameter, `skipEmptyId`, allows users to decide whether they want to skip a document when `xpath` evaluates to `null` or `""`, or to just give the Document a `UUID`. (If the `xmlIdPath` contains a qualifier, this should be enabled.)

Some secondary effects of the PR: 
- The id path is now relative to the root path. For example, if the root is `/Company/Staff`, the idPath would previously be `/Company/Staff/id`, but would now be just `id`.
- "Nested" XML is no longer supported. Nested XML is invalid. We had unit tests making sure we handled this gracefully, but XPath does not allow for nested XML. (These unit tests were removed.)

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/XMLFileHandler.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/XMLFileHandler.java
@@ -33,8 +33,8 @@ import org.xml.sax.XMLReader;
  * <ul>
  *   <li><code>xmlRootPath</code> (String): The path to the root of the XML Element(s) you want to publish Documents for. For example, in a file
  *   that contains "&lt;Company&gt;" and, inside it, multiple "&lt;Staff&gt;", each of which you want to publish a Document for, your
- *   path should be <code>/Company/Staff</code>.
- *   <li><code>xmlIdPath</code></li> (String): The <i>absolute</i> path to the XML element you want to use as the Document ID. For example, <code>/Company/Staff/id</code>.
+ *   path should be <code>/Company/Staff</code>.</li>
+ *   <li><code>xmlIdPath</code> (String): The <i>absolute</i> path to the XML element you want to use as the Document ID. For example, <code>/Company/Staff/id</code>.</li>
  *   <li><code>xpathIdPath</code> (String): An xpath expression, <i>relative</i> to <code>xmlRootPath</code>, specifying the XML element
  *   or attribute that will be used to create Document IDs. This path can be a qualifier - be sure to enable <code>skipEmptyId</code>, if so!</li>
  *   <li><code>encoding</code> (String, Optional): The encoding the xml files use. Defaults to utf-8.</li>
--- a/lucille-core/src/test/resources/FileConnectorTest/example.conf
+++ b/lucille-core/src/test/resources/FileConnectorTest/example.conf
@@ -15,7 +15,7 @@
     }
     "xml" : {
       xmlRootPath : "/Company/staff"
-      xmlIdPath : "id"
+      xmlIdPath : "/Company/staff/id"
       encoding : "utf-8"
       outputField : "xml"
       docIdPrefix: "xmlHandled-"


lucille-core/src/main/java/com/kmwllc/lucille/connector/xml/ChunkingXMLHandler.java
```diff
@@ -80,26 +89,37 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
       String xml = "Malformed";
+      String id = null;
+
       try {
         xml = ris.returnUpTo("</" + qName + ">");
+
+        // we have the full XML object. run xpath to get the ID on just this XML.
+        try (InputStream xmlStream = new ByteArrayInputStream(xml.getBytes())) {
+          org.w3c.dom.Document xmlDoc = builder.parse(xmlStream);
+          id = docIdExpression.evaluate(xmlDoc.getDocumentElement());
```
rseitz, line 100: what is the performance impact of running an xpath to get the doc ID

we should be thinking about possibly huge xml files

might want to preserve the old way of getting the ID as an option if it is indeed significantly faster

lucille-core/src/test/resources/FileHandlerTest/XMLFileHandlerTest/staffIDAttribute.xml
```diff
@@ -1,18 +1,21 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Company>
-    <staff>
-        <id>1001</id>
```
rseitz, line 4: are we testing this scenario with a nested id element somewhere else?

lucille-core/src/main/java/com/kmwllc/lucille/connector/xml/ChunkingXMLHandler.java
mrwrench24:
```java
System.out.println("To extract ID. " + Instant.now());
        
// we have the full XML object. run xpath to get the ID on just this XML.
try (InputStream xmlStream = new ByteArrayInputStream(xml.getBytes())) {
  org.w3c.dom.Document xmlDoc = builder.parse(xmlStream);
  id = docIdExpression.evaluate(xmlDoc.getDocumentElement());
}

System.out.println("Done extracting ID. " + Instant.now());
```

```
To extract ID. 2025-06-03T19:51:13.251873Z
Done extracting ID. 2025-06-03T19:51:13.255376Z
```

On a small xml element:
```xml
<staff>
        <name>daniel</name>
        <salary currency="USD">3000</salary>
        <bio>I am from San Diego</bio>

        <info>
          <id>1001</id>
          <role>software engineer</role>
        </info>
</staff>
```

mrwrench24: On a large(r) XML file... each element is a little less than 50 lines. So not *massive*, but larger than before. 
```
Time (ms) to process: 6
Time (ms) to process: 2
Time (ms) to process: 1
Time (ms) to process: 0
Time (ms) to process: 1
```

mrwrench24: I also tested on a Wikidata dump that was ~250 MB in size. Each "page" had a few elements, but the chunk of the data size here is just the raw text for the wiki page. Still, each Document was taking between 0 ms and 1 ms to be published.

rseitz: I was thinking that a small slowdown in extracting the ID would add up if we're processing a huge piece of XML with many, many documents inside it. From testing so far, do we have a sense this?

mrwrench24:
```java
String filePath = "/Users/jsquatrito/Downloads/wikipedia-test.xml";
    File file = new File(filePath);

    FileInputStream fileStream = new FileInputStream(file);

    System.out.println(Instant.now());

    xmlHandler.processFileAndPublish(publisher, fileStream, filePath);

    System.out.println(Instant.now());
```

Old ID:
```
2025-06-11T13:08:40.490667Z
25/06/11 09:08:40  10 INFO PublisherImpl: First doc published after 163 ms
2025-06-11T13:08:47.933588Z
```

New (XPath) ID:
```
2025-06-11T13:10:25.294237Z
25/06/11 09:10:25  ade9a3e0-4f50-4d90-9b94-b6c82eb9ba32 INFO PublisherImpl: First doc published after 106 ms
2025-06-11T13:10:39.316645Z
```

So it looks like it does add some time. (This is to publish 27,000+ Documents).

PR #330, opened 2025-06-05
LC-690: Proof-of-Concept for Static, Typed Specs

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/core/Connector.java
```diff
@@ -77,6 +78,20 @@ public interface Connector extends AutoCloseable {
...
+   */
+  Spec getSpec() throws Exception;
+
+  /**
+   * @return The Spec describing the given Connector class's Config properties. Must be declared as <code>public static Spec SPEC</code>.
+   * @throws ReflectiveOperationException If this Connector does not have a <code>public static Spec SPEC</code>.
+   */
+  static Spec getSpecFor(Class<? extends Connector> connectorClass) throws Exception {
+    return (Spec) connectorClass.getDeclaredField("SPEC").get(null);
+  }
```
mrwrench24, line 93: Not in use at the moment, but imagining the UI or the API taking in a class name (that we know is a Connector) and calling this method just to get the spec which will then be serialized. The argument to this method certainly could need to change to just Class or just a String representing a class.

lucille-core/src/main/java/com/kmwllc/lucille/core/spec/AnyProperty.java
```diff
@@ -0,0 +1,15 @@
+package com.kmwllc.lucille.core.spec;
+
+import com.typesafe.config.Config;
+
+// TODO: Not sure if this is something we'd want to include in a final version. I made it to temporarily not need
+//  a full rewrite of every Spec, but not sure if there is a property that is of multiple types, sometimes.
+public class AnyProperty extends Property {
```
mrwrench24, line 7: Just again emphasizing the TODO above.
This allows us to bridge the gap in this PR, so `withRequiredProperties` and `withOptionalProperties` are still valid method calls / still work, for now.

lucille-core/src/main/java/com/kmwllc/lucille/core/spec/Spec.java
```diff
@@ -0,0 +1,571 @@
...
+
+  /**
+   * Returns this Spec with the given properties added as optional properties.
+   * @param optionalProperties The optional properties you want to add to this Spec.
+   * @return This Spec with the given optional properties added.
+   */
+  public Spec withOptionalProperties(String... optionalProperties) {
+    Arrays.stream(optionalProperties).forEach(optionalPropertyName -> properties.add(new AnyProperty(optionalPropertyName, false)));
+    return this;
+  }
```
mrwrench24, line 117: these two methods are to eventually be removed, everything will be typed

lucille-core/src/main/java/com/kmwllc/lucille/connector/AbstractConnector.java
```diff
@@ -32,13 +32,21 @@ public abstract class AbstractConnector implements Connector {
...
     this.pipelineName = config.hasPath("pipeline") ? config.getString("pipeline") : null;
     this.docIdPrefix = config.hasPath("docIdPrefix") ? config.getString("docIdPrefix") : "";
     this.collapse = config.hasPath("collapse") ? config.getBoolean("collapse") : false;
 
+    Spec spec;
+
+    try {
+      spec = getSpec();
+    } catch (Exception e) {
+      throw new RuntimeException("Error accessing " + getClass() + " Spec. Is it publicly and statically available under \"SPEC\"?", e);
```
rseitz, line 47: Suggest moving this try/catch into getSpec() 
could remove "throws Exception" from getSpec() since we'd be throwing an unchecked RuntimeException

lucille-core/src/main/java/com/kmwllc/lucille/core/spec/ObjectProperty.java
```diff
@@ -0,0 +1,67 @@
...
+
+    if (description != null) {
+      node.put("description", description);
+    }
+
+    node.put("type", "OBJECT");
+    node.set("child", parentSpec.serialize());
+
+    return node;
+  }
```
mrwrench24, line 51: also something that wouldn't be ideal in a final implementation

lucille-core/src/main/java/com/kmwllc/lucille/core/spec/Spec.java
```diff
@@ -0,0 +1,572 @@
...
+
+    node.set("fields", fieldsArray);
+
+    return node;
+  }
+
+  // *****************************
+
+  /** Represents a Spec for a Parent in a Config. */
+  public static class ParentSpec extends Spec {
```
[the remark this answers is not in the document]
mrwrench24: https://stackoverflow.com/questions/2165613/java-generic-type on the "Curiously Recurring Template Pattern"

lucille-core/src/main/java/com/kmwllc/lucille/connector/CSVConnector.java
```diff
@@ -30,11 +30,13 @@ public class CSVConnector extends AbstractConnector {
   private final String moveToAfterProcessing;
   private final String moveToErrorFolder;
 
+  public static final Spec SPEC = Spec.connector()
```
rseitz, line 33: may want to add a method to Spec to make it unmodifiable -- to call at the end of chain
by using all uppercase SPEC we're indicating that the object is a "constant" which means it should be unmodifiable

mrwrench24: Would you want to require this method call? 

in other words, do we want to have a formalized builder pattern, where we have to call `build()` to get an actual `Spec` instead of just a `SpecBuilder`?

It would be a decent sized change, but it could also remedy some of the issues I've been having on "ParentSpec", I think - would have to research it more

lucille-core/src/main/java/com/kmwllc/lucille/core/spec/Spec.java
```diff
@@ -0,0 +1,572 @@
+package com.kmwllc.lucille.core.spec;
```
mrwrench24, line 1: TODO:

If we move forward with this, in a finalized PR, bring back some of the stylistic changes to CLI validation / Spec Validation made in LC-411

lucille-core/src/main/java/com/kmwllc/lucille/connector/CSVConnector.java
rseitz: eventually yes, I think we should require it, but if it adds complexity to this PR let's wait to do this in a second pass

mrwrench24:
All specs are now static. 
Now, to make then all typed.

lucille-core/src/main/java/com/kmwllc/lucille/core/spec/Spec.java
```diff
@@ -0,0 +1,571 @@
...
+   * @return a Spec without any default, legal properties.
+   */
+  public static Spec withoutDefaults() { return new Spec(); }
+
+  /**
+   * Creates a ParentSpec with the given name. Has no default legal properties.
+   * @param parentName The name of the parent you are creating a spec for. Must not be null.
+   * @return A ParentSpec with the given name.
+   */
+  public static ParentSpec parent(String parentName) {
```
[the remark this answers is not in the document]
mrwrench24: seems more reasonable to just get rid of this since it is a bit clunkier to declare a list / object in general now, but 1.) don't want to add much more to this PR and 2.) would make configs like the file connector harder to declare

lucille-core/src/test/java/com/kmwllc/lucille/stage/QueryOpensearchTest.java
```diff
@@ -239,4 +239,9 @@ public void testBadResponsePath() throws Exception {
       assertEquals("", testDoc.getJson("response").toString());
     }
   }
+
+  @Test
+  public void sandbox() throws Exception {
+    System.out.println(QueryOpensearch.SPEC.toJson().toPrettyString());
```
rseitz, line 245: Remove System.out.println
Do we need this test?

spencersolomon6 (approved):

PR #337, opened 2025-06-23
LC-702: Support "Any Matching" in DictionaryLookup

mrwrench24:
When you use `set_only`, you can control whether it uses any or all matching with the optional `use_any_match` parameter. It defaults to `false`, so existing configs will not be changed.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
@@ -38,8 +38,9 @@ import org.slf4j.LoggerFactory;
  * <p>  - use_payloads (Boolean, Optional) : denotes whether payloads from the dictionary should be used or not. Defaults to true.
  * <p>  - update_mode (String, Optional) : Determines how writing will be handling if the destination field is already populated.
  *      Can be 'overwrite', 'append' or 'skip'. Defaults to 'overwrite'.
- * <p>  - set_only (Boolean, Optional) : If true, the destination field will be set to true if all values in the source field
- *      are present in the dictionary.
+ * <p>  - set_only (Boolean, Optional) : If true, the destination field will be set to true when a match is found (instead of
+ *      outputting the match itself). You can control this behavior with <code>use_any_match</code> and <code>ignore_missing_source</code>.
+ *      <code>set_only</code> defaults to false.
  * <p>  - use_any_match (Boolean, Optional) : Use in combination with set_only. If true, the destination field will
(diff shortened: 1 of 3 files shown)


rseitz (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -34,12 +35,15 @@
  *       for a 1-1 mapping of results or supply one destination field for all of the source fields to be mapped into.
  * <p>  - dict_path (String) : The path the dictionary to use for matching. If the dict_path begins with "classpath:" the classpath
  *       will be searched for the file. Otherwise, the local file system will be searched.
- * <p>  - use_payloads (Boolean, Optional) : denotes whether paylaods from the dictionary should be used or not. Defaults to true.
+ * <p>  - use_payloads (Boolean, Optional) : denotes whether payloads from the dictionary should be used or not. Defaults to true.
  * <p>  - update_mode (String, Optional) : Determines how writing will be handling if the destination field is already populated.
  *      Can be 'overwrite', 'append' or 'skip'. Defaults to 'overwrite'.
  * <p>  - set_only (Boolean, Optional) : If true, the destination field will be set to true if all values in the source field
  *      are present in the dictionary.
```
spencer-solomon-tfs, line 42: We should improve the documentation around set_only to be more specific about what it is doing and/or include references to the use_any_match/ignore_missing_source params. ie:

```
- set_only (Boolean, Optional) : If true, will set the dest field to `true` when a match is found, instead of outputting the match. Specific behavior can be controlled with `use_any_match` and `ignore_missing_source`. Defaults to false.
```

lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -34,12 +35,15 @@
...
- * <p>  - ignore_missing_source (Boolean, Optional) : Intended to be used in combination with set_only. If true, the destination field
+ * <p>  - use_any_match (Boolean, Optional) : Use in combination with set_only. If true, the destination field will
```
spencersolomon6, line 43: We should improve the documentation around set_only to be more specific about what it is doing and/or include references to the use_any_match/ignore_missing_source params. ie:

```
- set_only (Boolean, Optional) : If true, will set the dest field to `true` when a match is found, instead of outputting the match itself. Specific behavior can be controlled via `use_any_match` and `ignore_missing_source`. Defaults to false.
```

PR #338, opened 2025-06-23
LC-702 followup: DictionaryLookup "any match" is true for any source field

mrwrench24:


lucille-core/src/main/java/com/kmwllc/lucille/stage/DictionaryLookup.java
```diff
@@ -130,14 +131,15 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
               .map(ignoreCase ? String::toLowerCase : String::toString);
 
           if (useAnyMatch) {
```
rseitz, line 133: can we just do the doc.update here and return null if a match is found?

kiratraynor (approved):

spencersolomon6 (approved):

PR #349, opened 2025-08-05
LC-719: JsonFileHandler Id Fields

HenryCaldwell:
Added functionality to JsonFileHandler to declare a single or composite id field and an id format similar to that of CsvFileHandler.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/JsonFileHandler.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/JsonFileHandler.java
@@ -1,15 +1,22 @@
 package com.kmwllc.lucille.core.fileHandler;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/JsonFileHandler.java
```diff
@@ -70,12 +91,28 @@ public Document next() {
...
+            ObjectNode node = (ObjectNode) mapper.readTree(line);
+            List<String> parts = new ArrayList<>(idFields.size());
+
+            for (String f : idFields) {
+              JsonNode v = node.get(f);
+              parts.add((v != null && !v.isNull()) ? v.asText() : "");
+            }
+
+            String rawId = (docIdFormat != null) ? String.format(docIdFormat, parts.toArray()) : String.join("_", parts);
+            node.put(Document.ID_FIELD, docIdPrefix + rawId);
```
rseitz, line 106: there's an edge case here where the json has a field called "id" but the user configures the filehandler to use a different field as the id

in this case, we're overwriting the "id" that's in the source json

this may be OK but we should document the behavior (in the class level javadocs for this file handler)

PR #354, opened 2025-08-12
LC-722: FileHandler warning on empty or null idFields

HenryCaldwell:
Added warnings to CSV and JSON file handlers for when an id field is either empty or null. Not implemented for XML since it doesn't support multiple idFields.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/JsonFileHandler.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/JsonFileHandler.java
@@ -112,10 +112,20 @@ public class JsonFileHandler extends BaseFileHandler {
           } else {
             ObjectNode node = (ObjectNode) mapper.readTree(line);
             List<String> parts = new ArrayList<>(idFields.size());
+            List<String> blankIdFields = new ArrayList<>();
 
             for (String fieldName : idFields) {
               JsonNode valueNode = node.get(fieldName);
               parts.add((valueNode != null && !valueNode.isNull()) ? valueNode.asText() : "");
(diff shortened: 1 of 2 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/JsonFileHandler.java
```diff
@@ -112,10 +112,19 @@ public Document next() {
           } else {
             ObjectNode node = (ObjectNode) mapper.readTree(line);
             List<String> parts = new ArrayList<>(idFields.size());
+            List<String> blankIdFields = new ArrayList<>();
```
rseitz, line 115: We're doing this for every single document so we should try to be economical about creating new objects. In this case, I'd recommend not keeping a blankIdFields array but just logging a warning for each individual blank field as it is encountered.

lucille-core/src/main/java/com/kmwllc/lucille/core/fileHandler/JsonFileHandler.java
```diff
@@ -112,10 +112,19 @@ public Document next() {
...
 
             for (String fieldName : idFields) {
               JsonNode valueNode = node.get(fieldName);
               parts.add((valueNode != null && !valueNode.isNull()) ? valueNode.asText() : "");
+              String value = (valueNode != null && !valueNode.isNull()) ? valueNode.asText() : null;
+              if (StringUtils.isBlank(value)) {
+                blankIdFields.add(fieldName);
```
rseitz, line 122: log warning here, don't save the fieldName

rseitz (approved):

PR #358, opened 2025-08-15
LC-723: testing for drop deadlock

HenryCaldwell:


HenryCaldwell:
Need to add testing for percentage field in DopDocument

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
```diff
@@ -856,4 +856,18 @@ public void testStringifyValidationNoExceptions() {
...
+   * Attempt to reproduce a deadlock that was seen in a pipeline that dropped a high percentage of docs, but not all docs.
+   * The deadlock is also sensitive to how many docs are processed in total, the batch size, and the logging configuration.
+   *
+   * This test will possibly hang if the deadlock potential still exists in the code. It should pass in several seconds
+   * assuming the deadlock has been fixed, otherwise it will time out and fail after 30 seconds.
+   */
+  @Test(timeout = 30000)
+  public void testDropDeadlock() throws Exception {
+    RunResult result =
+        Runner.run(ConfigFactory.load("RunnerTest/dropDeadlock.conf"), Runner.RunType.TEST);
```
rseitz, line 870: if we switch from Runner.RunType.TEST to Runner.RunType.LOCAL can we reduce the number of docs required to consistently reproduce this issue any further?

the main difference between RunType.TEST and RunType.LOCAL is that TEST saves all the messages that are sent during the run for later inspection. We don't need those saved messages here because we're not looking at them. It's not clear that switching to LOCAL will change the scenario meaningfully, just wondering if it happens to make a difference...

lucille-core/src/test/java/com/kmwllc/lucille/stage/DropDocumentTest.java
```diff
@@ -119,4 +119,46 @@ public void testNumberConditional() throws StageException {
...
+
+      if (doc.isDropped()) {
+        dropped++;
+      } else {
+        kept++;
+      }
+    }
+
+    assertTrue(dropped > 0);
+    assertTrue(kept > 0);
```
spencersolomon6, line 143: This test isn't deterministic and there's a chance it can fail despite the Stage working properly... there's a chance, albeit very very small (7.8886091e-31), that we could drop or keep all of the documents. 

This may be ok since the chances are so small but want to call it out.

spencersolomon6: We might want to just check that dropped + kept == 100 instead... that is guaranteed to be true

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
```diff
@@ -856,4 +856,18 @@ public void testStringifyValidationNoExceptions() {
...
+   *
+   * This test will possibly hang if the deadlock potential still exists in the code. It should pass in several seconds
+   * assuming the deadlock has been fixed, otherwise it will time out and fail after 30 seconds.
+   */
+  @Test(timeout = 30000)
+  public void testDropDeadlock() throws Exception {
+    RunResult result =
+        Runner.run(ConfigFactory.load("RunnerTest/dropDeadlock.conf"), Runner.RunType.LOCAL);
+    assertTrue(result.getStatus());
+  }
```
spencersolomon6, line 872: [curiosity] did this test always produce the deadlock when it existed?

spencersolomon6 (changes requested):

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
HenryCaldwell: Yes, with a lower batch size it was slightly inconsistent, but bumping it up to 10k it always hangs.

spencersolomon6 (approved):

PR #361, opened 2025-08-18
LC-724: Full config validation at the beginning of each run

HenryCaldwell:
Updated runner to run config validation on every run as opposed to just on -validate.

Updated "name" in connectors to be an optional field through validation.

Removed old pollIntervalMs from kafka config file.

### DIFF
--- a/application-example.conf
+++ b/application-example.conf
@@ -267,9 +267,6 @@ zookeeper {
 kafka {
   bootstrapServers: "localhost:9092"
 
-  # how long a kafka poll should wait for data before returning an empty record set
-  pollIntervalMs: 250
-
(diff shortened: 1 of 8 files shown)


lucille-core/src/test/resources/KafkaTest/twoConnectors.conf
```diff
@@ -46,9 +46,6 @@ indexer {
 kafka {
   bootstrapServers: "localhost:9090"
 
-  # how long a kafka poll should wait for data before returning an empty record set
-  pollIntervalMs: 250
```
rseitz, line 50: there are several other configs in the codebase that have pollIntervalMs. Let's remove all instances as part of this PR.

rseitz (approved):

PR #362, opened 2025-08-21
LC-714: Pipeline stage for adding nested fields

HenryCaldwell:
Added a stage for creating nested fields with a variable number of objects within. Objects values can be configured with pre defined values from existing fields or random values created by generators using another stage.

Example conf:

```
{
  name: "test_nested_mixed"
  class: "com.kmwllc.lucille.stage.BuildNested"
  target_field: "test_nested"

  # Range of objects within each nested field
  min_num_objects: 1
  max_num_objects: 3

  # Entries within each nested object
  entries {
    "metrics.random_score" = "intGen" # Random int generator
    "content.title" = "test_source" # Previously defined document field
    "timestamps.created" = "dateGen" # Random date generator
  }

  # Random int and date generators
  generators {
    intGen {
      class = "com.kmwllc.lucille.stage.AddRandomInt"
      range_start = 0
      range_end   = 1000000
    }

    dateGen {
      class = "com.kmwllc.lucille.stage.AddRandomDate"
      range_start_date = "2015-01-01"
      range_end_date   = "2025-12-31"
    }
  }
}
```


lucille-core/src/main/java/com/kmwllc/lucille/stage/BuildNested.java
```diff
@@ -0,0 +1,332 @@
...
+
+    if (minNumObjects != null) {
+      return ThreadLocalRandom.current().nextInt(minNumObjects, maxNumObjects + 1);
+    }
+
+    return 1;
+  }
+
+  // Generate a value with the provided stage
+  private Object generateWith(String genKey, Document doc) throws StageException {
```
rseitz, line 267: Suggest that we don't run the generator on the doc being processed
All we need is one reserved doc that we can keep around in an instance variable
we can use this reserved doc to run the generator on, then get the field value, then remove the output field
the name of the output field could be fixed, don't need to take it as param

lucille-core/src/main/java/com/kmwllc/lucille/stage/BuildNested.java
```diff
@@ -0,0 +1,332 @@
...
+      out.put(key, trimmedSrc);
+    }
+
+    return out;
+  }
+
+  @Override
+  public Iterator<Document> processDocument(Document doc) throws StageException {
+    // Get doc as map so we can use previous field values
+    Map<String, Object> map = doc.asMap();
```
rseitz, line 202: can you give an example of why this is necessary?
where would be looking for a value in this map that we can't get directly from the doc?

lucille-core/src/main/java/com/kmwllc/lucille/stage/BuildNested.java
```diff
@@ -0,0 +1,330 @@
...
+    if (gen == null) {
+      return null;
+    }
+
+    // Run the generator on the current doc once
+    gen.processDocument(genDoc);
+    JsonNode value = genDoc.getJson(GEN_OUT_FIELD);
+
+    // Clean up temp field
+    if (genDoc.has(GEN_OUT_FIELD)) {
```
rseitz, line 281: might be safer to do this before calling gen.processDocument instead of after
that way if the generator has append instead of overrwrite behavior, we can be sure the GEN_OUT_FIELD wont be there

HenryCaldwell:
> By naming convention, we might want to rename this stage to AddRandomNestedField .. so it's a little more aligned with the other random data generation stages...

It does however also allow for nesting previously defined fields, not just random ones.

It could also be split in to two different stages for clarity if that would make sense.

For example right now if you already had 2 previously defined fields in each document called company_name and product you could have;

```
{
  name: "nested_details"
  class: "com.kmwllc.lucille.stage.BuildNested"
  target_field: "nested_details"

  min_num_objects: 1
  max_num_objects: 1

  # Entries within each nested object
  entries {
    "name" = "company_name" # Previously defined name
    "product" = "product" # Previously defined product
  }
}
```

rseitz (approved):
approving... see one minor pending comment re: stage SPEC

PR #363, opened 2025-08-22
LC-680-v2: API endpoints for stages, connectors, indexers

HenryCaldwell:
Json files now generate on mvn clean install, endpoints return and cache results for stage, connector, and indexer specs, field descriptions, and class names.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/doclet/JsonDoclet.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/doclet/JsonDoclet.java
@@ -1,0 +1,440 @@
+package com.kmwllc.lucille.doclet;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.kmwllc.lucille.util.ClassUtils.ClassDescriptor;
+import com.kmwllc.lucille.util.ClassUtils.MethodDescriptor;
+
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.SourceVersion;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.Modifier;
+import javax.tools.Diagnostic;
+import java.io.Writer;
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.HashMap;
+import java.util.stream.Collectors;
+
+import jdk.javadoc.doclet.Doclet;
+import jdk.javadoc.doclet.DocletEnvironment;
+import jdk.javadoc.doclet.Reporter;
+
+/**
+ * A modern JDK-9+ Doclet that generates a JSON file of ClassDescriptor and MethodDescriptor.
+ * This doclet extracts documentation from source code structure rather than loading runtime classes.
+ */
+public class JsonDoclet implements Doclet {
+
+  private Reporter reporter;
+  private String outputFile = "javadocs.json"; // Default output file
+  private String outputPath = "target"; // Default output directory
+  private boolean helpWanted = false;
+
+  @Override
+  public void init(java.util.Locale locale, jdk.javadoc.doclet.Reporter reporter) {
+    this.reporter = reporter;
+  }
+
+  @Override
+  public String getName() {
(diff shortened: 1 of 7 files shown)


run-jsondoclet.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
+# Directory paths
+PROJECT_ROOT=$(pwd)
+TARGET_DIR="$PROJECT_ROOT/lucille-core/target"
+LIB_DIR="$TARGET_DIR/lib"
+CLASSES_DIR="$TARGET_DIR/classes"
+OUTPUT_DIR="$PROJECT_ROOT/lucille-plugins/lucille-api/src/main/resources"
```
rseitz, line 8: Looks like we're outputting directly into the source tree.
I think it would be cleaner to output into target.
If there's a need to output directly into source, we likely need to add these generated files to .gitignore

lucille-plugins/lucille-api/src/test/java/endpoints/ConfigInfoTest.java
```diff
@@ -0,0 +1,74 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.kmwllc.lucille.AuthHandler;
+import com.kmwllc.lucille.endpoints.ConfigInfo;
+
+import jakarta.ws.rs.core.Response;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ConfigInfoTest {
```
rseitz, line 17: should add a test that actually looks at the json returned for a specific stage and confirms that the fields and field descriptions are as we would expect

lucille-core/src/main/java/com/kmwllc/lucille/doclet/JsonDoclet.java
```diff
@@ -0,0 +1,426 @@
...
+        if (enclosedElement instanceof TypeElement) {
+          TypeElement innerType = (TypeElement) enclosedElement;
+          Set<Modifier> modifiers = innerType.getModifiers();
+
+          // Check if it's a public static inner class/enum
+          if ((innerType.getKind() == ElementKind.CLASS || innerType.getKind() == ElementKind.ENUM) &&
+              modifiers.contains(Modifier.PUBLIC) &&
+              modifiers.contains(Modifier.STATIC)) {
+
+            // Check if it follows the naming convention: OuterClassNameConfig
```
rseitz, line 298: can remove everything relating to inner *Config classes...

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/ConfigInfo.java
```diff
@@ -226,15 +223,12 @@ public Response getStages(@Parameter(hidden = true) @Auth Optional<PrincipalImpl
       return authResponse;
     }
 
-    if (cachedStageListJson != null) {
-      return Response.ok(cachedStageListJson, MediaType.APPLICATION_JSON).build();
+    if (cachedStageListJson == null) {
+      Map<String, ComponentDoc> ComponentDocs = loadDocs("stage-javadocs.json");
+      cachedStageListJson = buildSpecArrayForSubclasses("com.kmwllc.lucille.core.Stage", ComponentDocs);
```
rseitz, line 228: could use Stage.class.getName() as is being done with connectors and indexers

run-jsondoclet.sh
```diff
@@ -0,0 +1,68 @@
...
+if [ ! -d "$LIB_DIR" ]; then
+  echo "Library directory not found at $LIB_DIR. Running Maven to download dependencies..."
+  mvn dependency:copy-dependencies -DoutputDirectory="$LIB_DIR" -f "$PROJECT_ROOT/lucille-core/pom.xml"
+fi
+
+# Build the classpath with all dependencies
+CLASSPATH="$CLASSES_DIR"
+
+# Add all JARs in the lib directory
+for JAR in "$LIB_DIR"/*.jar; do
```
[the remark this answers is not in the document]
HenryCaldwell: It seems like we need to keep the loop because javadoc wont expand lib/* on -docletpath, so listing each JAR is the only way it finds the doclet's deps like Gson. Without listing each JAR we hit a NoClassDefFoundError.

rseitz (approved):

PR #364, opened 2025-09-03
LC-727: Extract entities FST based stage implementation

HenryCaldwell:
Added an FST based entity extraction stage that has ~2.3x better memory performance in comparison to the Aho-Corasick implementation.


lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
```diff
@@ -0,0 +1,483 @@
...
+ *   <li>entity_field (String, Optional) : When set and use_payloads=true, also writes the matched normalized surface terms
+ *   to this field.</li>
+ *   <li>s3 (Map, Optional) : If your dictionary files are held in S3. See FileConnector for the appropriate arguments to provide.</li>
+ *   <li>azure (Map, Optional) : If your dictionary files are held in Azure. See FileConnector for the appropriate arguments to provide.</li>
+ *   <li>gcp (Map, Optional) : If your dictionary files are held in Google Cloud. See FileConnector for the appropriate arguments to provide.</li>
+ * </ul>
+ */
+public class ExtractEntitiesFST extends Stage {
+
+  public static final Spec SPEC = SpecBuilder.stage()
```
rseitz, line 52: should add ignore_case as an option
to be consistent with the existing ExtractEntities, it would have to default to false

lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
```diff
@@ -0,0 +1,484 @@
...
+            break;
+          }
+
+          // Skip empty rows
+          if (line.length == 0 || line[0] == null) {
+            continue;
+          }
+
+          // Normalize term with trim and lowercase
+          String term = normalizeKey(line[0].trim());
```
rseitz, line 154: normalizeKey trims the key so we don't have to do it here

lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
```diff
@@ -0,0 +1,484 @@
...
+    } catch (IOException e) {
+      // StringReader -> shouldn't happen
+      throw new RuntimeException(e);
+    }
+
+    return out;
+  }
+
+  // Check if a sequence of tokens forms a complete dictionary match
+  private boolean isCompleteMatch(List<String> tokens) {
```
rseitz, line 406: seems like this method is never called anymore?

lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
```diff
@@ -0,0 +1,484 @@
...
+      }
+    }
+
+    return termToPayload;
+  }
+
+  private void buildFSTNoPayloads() throws IOException, StageException {
+    // Sort terms lexicographically
+    List<String> terms = new ArrayList<>(loadTermsAndPayloads().keySet());
+    terms.sort(Comparator.naturalOrder());
```
rseitz, line 175: sorting could be expensive for a huge dictionary -- might want to have a flag like dictionariesArePresorted -- defaults to false but let the user set it to true if they want to save startup time for dictionaries that are already sorted

lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
```diff
@@ -0,0 +1,469 @@
...
+  private void buildFSTNoPayloads() throws IOException, StageException {
+    // Sort terms lexicographically
+    List<String> terms = new ArrayList<>(loadTermsAndPayloads().keySet());
+
+    if (!dictsSorted) {
+      terms.sort(Comparator.naturalOrder());
+    }
+
+    // Initialize FST builder with no payloads
+    var rw = FSTCompiler.getOnHeapReaderWriter(15);
```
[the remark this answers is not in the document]
rseitz: but let's add a comment saying where we got that 15 from -- e.g. where is it suggested?

lucille-core/src/test/java/com/kmwllc/lucille/stage/ExtractEntitiesFSTTest.java
```diff
@@ -0,0 +1,221 @@
...
+    doc.setField("input1", "I live in the United States.");
+    Stage stage = factory.get("ExtractEntitiesFSTTest/configEntityAndPayload.conf");
+    stage.processDocument(doc);
+    assertEquals("North America", doc.getString("payload"));
+    assertEquals("United States", doc.getString("entity"));
+  }
+
+  @Test
+  public void testExtractEntitiesSkipsMalformedLine() throws Exception {
+    Document doc = Document.create("doc");
```
kwatters, line 51: What is this test case intending to validate?

PR #365, opened 2025-09-04
LC-478: update CopyFields to allow nested fields.

kiratraynor:


lucille-core/src/test/java/com/kmwllc/lucille/stage/CopyFieldsTest.java
```diff
@@ -62,11 +66,52 @@ public void testCopyFieldsSkip() throws Exception {
     assertEquals("input3 should be skipped.", doc.getStringList("output3").get(0));
   }
 
+  @Test
+  public void testCopyFieldsNested() throws Exception {
```
rseitz, line 70: suggest adding a test where we try to copy a dotted field to another dotted field with isNested=false and confirm that the field names are treated as literals
.e.g we have a field named "a.b.c" and we try to copy its value into "x.y.z"

lucille-core/src/main/java/com/kmwllc/lucille/stage/CopyFields.java
```diff
@@ -1,70 +1,125 @@
...
-        continue;
-      }
+      if (!this.isNested) {
+        if (!doc.has(source)) {
+          continue;
+        }
+        doc.update(dest, updateMode, doc.getStringList(source).toArray(new String[0]));
+      } else {
+        // deal with nested case
+        String[] nestedSourceField = source.split("[.]");
```
rseitz, line 83: suggest renaming to sourceFieldParts b/c calling it "Field" seems to suggest it's a single String

lucille-core/src/main/java/com/kmwllc/lucille/stage/CopyFields.java
```diff
@@ -1,70 +1,125 @@
...
+          }
+        }
 
-      doc.update(destField, updateMode, doc.getStringList(sourceField).toArray(new String[0]));
+        if (sourceVal == null) {
+          continue;
+        }
+
+        // get dest json field value, if exists
+        String[] nestedDestField = dest.split("[.]");
```
rseitz, line 106: suggest renaming to destFieldParts

lucille-core/src/main/java/com/kmwllc/lucille/util/JSONUtils.java
```diff
@@ -0,0 +1,112 @@
+package com.kmwllc.lucille.util;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.IOException;
+
+public class JSONUtils {
```
rseitz, line 9: took another look at this util class and thinking maybe we don't need it... instead, we could put a new method on the Document interface like this:

`  default JsonNode getNestedJson(String name) {
    String[] pathSegments = name.split("\\.");

    if (pathSegments.length == 0 || !has(pathSegments[0])) {
      return null;
    }
    JsonNode node = getJson(pathSegments[0]);

    for (int i = 1; i < pathSegments.length; i++) {
      String segment = pathSegments[i];

      if (!node.has(segment)) {
        return null;
      }

      node = node.get(segment);
    }

    return node;
  };`

along with a similar Document.setNestedJson method...

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -570,4 +574,131 @@ static Document createFromJson(String json, UnaryOperator<String> idUpdater) thr
...
+    }
+
+    // set last node
+    setFieldSegment(currentNode, pathSegments[pathSegments.length-1], value);
+
+    setField(pathSegments[0], node);
+  }
+
+  // if the given segment is an integer, it will create a new ArrayNode, otherwise an ObjectNode
+  private JsonNode createNewNode(String segment) {
```
rseitz, line 655: all the private helper methods added here can be static

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -570,4 +574,131 @@ static Document createFromJson(String json, UnaryOperator<String> idUpdater) thr
     return JsonDocument.fromJsonString(json, idUpdater);
   }
 
+  /* --- JSON Getters/Setters --- */
+
+  /**
+   * Gets a nested JsonNode at a path like "a.b.c.d" where the path is split on '.' and each part is treated as a level of nesting.
+   * This also works for nested values that contain a list, such as "a.b.2.c" where 'b' is an ArrayNode.
```
rseitz, line 581: should mention that array indices are 0 based

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -570,4 +574,131 @@ static Document createFromJson(String json, UnaryOperator<String> idUpdater) thr
...
+      return;
+    }
+    if (pathSegments.length == 1) {
+      setField(pathSegments[0], value);
+      return;
+    }
+
+    // create a default fresh node where the type is based off of the child field, either ArrayNode or ObjectNode
+    JsonNode defaultFreshNode = createNewNode(pathSegments[1]);
+    JsonNode node = has(pathSegments[0]) ? getJson(pathSegments[0]) : defaultFreshNode;
```
rseitz, line 633: I would remove l632 and do
JsonNode node = has(pathSegments[0]) ? getJson(pathSegments[0]) : createNewNode(pathSegments[1]);

so we are not creating a new node we don't need to

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -570,4 +574,131 @@ static Document createFromJson(String json, UnaryOperator<String> idUpdater) thr
...
+  /* --- JSON Getters/Setters --- */
+
+  /**
+   * Gets a nested JsonNode at a path like "a.b.c.d" where the path is split on '.' and each part is treated as a level of nesting.
+   * This also works for nested values that contain a list, such as "a.b.2.c" where 'b' is an ArrayNode.
+   *
+   * @param name the nested field path to get the JsonNode from
+   * @return the JsonNode at the nested path or null if not found
+   */
+  default JsonNode getNestedJson(String name) {
```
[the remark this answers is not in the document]
rseitz: further refinement would be to take pathSegments as Object[] where we're expecting the objects to be Strings or Integers
that way we could do all the integer parsing in one pass instead of having to do it in setFieldSegment and getFieldSegment

lucille-core/src/test/java/com/kmwllc/lucille/core/DocumentTest.java
```diff
@@ -1654,6 +1654,60 @@ public void testRemoveChildren() {
     assertFalse(doc.hasChildren());
   }
```
rseitz, line 1656: some smaller tests that illustrate how array indices work...

```
  @Test
  public void testNestedArrayUpdate() {
    Document doc = Document.create("id1");
    doc.setNestedJson("a.b.0", IntNode.valueOf(50));
    doc.setNestedJson("a.b.0", IntNode.valueOf(60));
    assertEquals(1, doc.getNestedJson("a.b").size());
    assertEquals(60, doc.getNestedJson("a.b.0").intValue());
    assertEquals(null, doc.getNestedJson("a.b.1"));
  }

  @Test
  public void testNestedArrayUpdateConfusingArrayIndexScenario() {
    Document doc = Document.create("id1");
    doc.setNestedJson("a.b.1", IntNode.valueOf(50)); // since a.b doesn't exist, an array is created there and added at 0th position
    doc.setNestedJson("a.b.1", IntNode.valueOf(60)); // now a.b exists and we are adding a new value at the 1th position, giving 2 values
    assertEquals(2, doc.getNestedJson("a.b").size());
    assertEquals(50, doc.getNestedJson("a.b.0").intValue());
    assertEquals(60, doc.getNestedJson("a.b.1").intValue());
    assertEquals(null, doc.getNestedJson("a.b.2"));
  }
```

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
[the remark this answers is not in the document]
kiratraynor: That's a good idea! I'll add those methods and see if it would be beneficial to use them in CopyFields or AddRandomNestedField

lucille-core/src/test/java/com/kmwllc/lucille/core/DocumentTest.java
kiratraynor: As discussed, I will add these tests to add more detail about the array logic. And I am updating the setter to error if you try to set an index of an array that does not exist yet or is not the next index in the array.

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -652,7 +670,7 @@ default void setNestedJson(String name, JsonNode value) {
   }
 
   // if the given segment is an integer, it will create a new ArrayNode, otherwise an ObjectNode
-  private JsonNode createNewNode(String segment) {
+  static private JsonNode createNewNode(String segment) {
```
rseitz, line 673: "private static" is our convention, instead of "static private" (even though they mean the same)

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -627,7 +627,7 @@ default JsonNode getNestedJson(String[] pathSegments) {
    * @param name the nested field path to set the JsonNode at
    * @param value the JsonNode to set at the nested path
    */
-  default void setNestedJson(String name, JsonNode value) {
+  default void setNestedJson(String name, JsonNode value) throws IllegalArgumentException {
```
rseitz, line 630: IllegalArgumentException is an unchecked exception so you don't need to declare it
you still can declare it to be explicit, but I think it's more common not to declare

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomNestedField.java
```diff
@@ -245,7 +243,12 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
         }
 
         // Write value at the destination
-        doc.setNestedJson(targetFieldPath + "." + destPathStr, valNode);
+        try {
+          doc.setNestedJson(targetFieldPath + "." + destPathStr, valNode);
+        } catch(IllegalArgumentException ex) {
```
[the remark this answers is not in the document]
kiratraynor: I was thinking that it would be good to have the actual details of the docid and field here so it is easier to debug what went wrong. I can still remove it if you think that is unnecessary though.

In terms of CopyFields, I was thinking about that a little more. I think with that stage, it should not actually throw an error because if one of the fields is invalid I think the user would likely still want the document to still be processed and continue on with other fields in this stage as well as other stages. Let me know thoughts about this as well.

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -570,4 +574,150 @@ static Document createFromJson(String json, UnaryOperator<String> idUpdater) thr
...
+  private static void setFieldSegment(JsonNode node, String segment, JsonNode value) {
+    if (node.isArray()) {
+      int index = Integer.parseInt(segment);
+      // check if index is within the existing array and overwrite if so
+      if (index >= 0 && index < node.size()) {
+        ((ArrayNode) node).set(index, value);
+      } else if (index == node.size()) { // add to end of array (including creating as array if index==0)
+        ((ArrayNode) node).add(value);
+      } else {
+        throw new IllegalArgumentException("Cannot set index " + index + " on array of size " + node.size());
```
rseitz, line 693: thinking about this more... I would change this to ArrayIndexOutOfBoundsException because it's more descriptive. It's also an unchecked exception so you don't need to declare it.

I agree with what you did in AddRandomNestedField (catch & rethrow with more info) and CopyFields (only log an error)

lucille-core/src/main/java/com/kmwllc/lucille/stage/CopyFields.java
```diff
@@ -1,70 +1,93 @@
...
+    }
   }
 
   @Override
   public Iterator<Document> processDocument(Document doc) throws StageException {
-    for (int i = 0; i < sourceFields.size(); i++) {
-      // If there is only one source or dest, use it. Otherwise, use the current source/dest.
-      String sourceField = sourceFields.get(i);
-      String destField = destFields.size() == 1 ? destFields.get(0) : destFields.get(i);
+    for (Entry<String, Object> fieldPair : this.fieldMapping.entrySet()) {
```
rseitz, line 70: an optimization for the isNested case would be to convert the source and dest fields to String[] in the start() method and use those String[]'s when calling doc.getNestedJson() and doc.setNestedJson()

this would avoid having to repeatedly String.split the same source and dest fields in getNestedJson/setNestedJson for every single doc

lucille-core/src/main/java/com/kmwllc/lucille/stage/CopyFields.java
```diff
@@ -1,70 +1,102 @@
...
+    if (this.fieldMapping.size() == 0) {
+      throw new StageException("fieldMapping must have at least one source-dest pair for Copy Fields");
+    }
+
+    if (this.isNested && this.updateMode != null) {
+      log.info("Cannot set both isNested and update_mode in Copy Fields at the same time. Ignoring update_mode, fields will be overwritten.");
+    }
+
+    // create set of destination field parts so we don't have to split them up for every doc
+    for (Entry<String, Object> fieldPair : this.fieldMapping.entrySet()) {
```
rseitz, line 70: suggest that when handling the nested case, we should iterate over a list of Pair<String[], String[]> where both the source and dest have been pre-split
when handling the no nested case, we can just iterate over the entries in the fieldMapping.entrySet()
this way, handling nested, we never have to split source or dest inside processDocument(), and when handling non-nested, we never have to join source or dest inside processDocument()

will share a code example separately

lucille-core/src/main/java/com/kmwllc/lucille/stage/CopyFields.java
```diff
@@ -1,70 +1,102 @@
...
-      // If there is only one source or dest, use it. Otherwise, use the current source/dest.
-      String sourceField = sourceFields.get(i);
-      String destField = destFields.size() == 1 ? destFields.get(0) : destFields.get(i);
+    for (Entry<String, String[]> fieldPair : this.sourceToDestParts.entrySet()) {
+      String source = fieldPair.getKey();
+      String[] dest = fieldPair.getValue();
 
-      if (!doc.has(sourceField)) {
-        continue;
+      if (!this.isNested) {
```
rseitz, line 83: you don't have to use `this.` before instance variable all the time
it's a convention to do that in the constructor because constructors often take arguments that have the same name as the instance variable
inside a regular method body, you can use `this.` to disambiguate, if the method happens to take an arg with the same name as an instance variable, but it's generally easier to name the method argument so that this doesn't happen

lucille-core/src/test/java/com/kmwllc/lucille/core/DocumentTest.java
```diff
@@ -1654,6 +1655,83 @@ public void testRemoveChildren() {
...
+    Document doc = Document.create("id1");
+    doc.setNestedJson("a.b.0", IntNode.valueOf(50));
+    doc.setNestedJson("a.b.0", IntNode.valueOf(60));
+    assertEquals(1, doc.getNestedJson("a.b").size());
+    assertEquals(60, doc.getNestedJson("a.b.0").intValue());
+    assertNull(doc.getNestedJson("a.b.1"));
+  }
+
+  @Test
+  public void testNestedArrayUpdateConfusingArrayIndexScenario() throws IllegalArgumentException {
```
rseitz, line 1723: minor point re the name of this test -- I'm not sure this scenario is "confusing" any more... it was confusing back when two updates at index 1 would result in an array with two values...

maybe testNestedArrayUpdateIndexOutOfBounds() .. or roll these tests into the above testNestedArrayUpdate()

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomNestedField.java
```diff
@@ -130,32 +132,48 @@ public AddRandomNestedField(Config config) throws StageException {
...
+      }
+    }
 
-      try {
-        gen = Stage.fromConfig(injected);
-      } catch (Exception e) {
-        throw new StageException("Failed to instantiate generator '" + key + "'", e);
+    // create set of destination field parts so we don't have to split them on every doc
+    final int n = pickNumObjects();
+    int index = 0;
```
rseitz, line 163: why do we have to maintain this `index` separately from the `i` in the for loop?

rseitz (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomNestedField.java
```diff
@@ -208,53 +224,45 @@ private static Map<List<String>, String> parseEntries(Config entriesCfg) throws
 
   @Override
   public Iterator<Document> processDocument(Document doc) throws StageException {
-    final int n = pickNumObjects();
-    ArrayNode arr = mapper.createArrayNode();
+    // set new array node on doc
+    doc.setNestedJson(targetField, mapper.createArrayNode());
```
HenryCaldwell, line 228: This pre writes an empty array to the target field, so if something midway through throws (Missing value no generator), the doc ends up being left with an empty array in the target field. In the previous implementation the array was set at the end so failures didn't leave any mutations.

HenryCaldwell (changes requested):

lucille-core/src/main/java/com/kmwllc/lucille/stage/AddRandomNestedField.java
HenryCaldwell: I think that for clarity, we shouldn’t write an empty array unless the stage completes successfully. If we run into an error, the document should remain unchanged. I think it's cleaner to not touch the document unless the entire operation is successful, but I don't think it's anything breaking or significant either way so I'm more or less indifferent.

PR #367, opened 2025-09-11
LC-728: Stage for applying javascript

HenryCaldwell:


lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJavascript.java
```diff
@@ -0,0 +1,408 @@
...
+          .build();
+    } catch (Exception e) {
+      throw new StageException("Failed to initialize JavaScript context.", e);
+    }
+  }
+
+  @Override
+  public void stop() throws StageException {
+    if (context != null) {
+      context.close();
```
rseitz, line 85: the docs don't say that closing the context will close the engine it's using
and Engine has it's own close method
so I think we have to close the engine after closing the context

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJavascript.java
```diff
@@ -0,0 +1,408 @@
...
+      context.eval(source);
+      return null;
+    } catch (Exception e) {
+      throw new StageException("JavaScript failed for doc '" + doc.getId() + "' using '" + scriptPath + "'.", e);
+    }
+  }
+
+  // Proxy implementation exposed to JS
+
+  static final class JsDocProxy implements ProxyObject {
```
spencersolomon6, line 107: Should this class be private?

lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -669,6 +669,36 @@ default void setNestedJson(String[] pathSegments, JsonNode value) {
...
+
+    if (parts.length == 0) {
+      return;
+    }
+    if (parts.length == 1) {
+      removeField(parts[0]);
+      return;
+    }
+
+    JsonNode parent = getNestedJson(String.join(".",
```
rseitz, line 683: should use the version of getNestedJson that takes String[] so we can avoid having to do a String.join here

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJavascript.java
```diff
@@ -115,8 +109,10 @@ public void stop() throws StageException {
       context.close();
       context = null;
     }
-    if (scriptPath != null) {
-      fileFetcher.shutdown();
+
+    if (engine != null) {
+      engine.close();
+      engine = null;
```
rseitz, line 115: not necessary to set to null here

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/ApplyJavascript.java
```diff
@@ -0,0 +1,432 @@
...
+      JsonNode json = doc.getJson(key);
+
+      if (json != null) {
+        if (json.isNull()) return null;
+        if (json.isArray()) return jsonArrayToProxyArray(json, key);
+        if (json.isContainerNode()) return new JsDocProxy(doc, key);
+        return jsonNodeToJsValue(json);
+      }
+
+      Object raw = doc.asMap().get(key);
```
rseitz, line 172: how could we ever reach this line?
if doc.has(key) then doc.getJson(key) will always return non-null for JsonDocument -- looking at the implementation it just returns data.get(name)

so maybe this line can only be reached with HashMapDocument?
I'm going to suggest we limit ApplyJavascript to being used with JsonDocument because HashMapDocument is not really in use in the codebase and should be deprecated

would that save us some code here?

lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyJavascriptTest.java
```diff
@@ -0,0 +1,381 @@
...
+
+    Document d = Document.create("d");
+    d.setNestedJson("meta.x", TextNode.valueOf("existing"));
+
+    stage.processDocument(d);
+
+    assertEquals("existing", d.getNestedJson("meta.x").asText());
+    assertNull(d.getNestedJson("meta.missing"));
+  }
+}
```
rseitz, line 381: should add a test that shows the creation of a nested path via javascript:

```
  @Test
  public void testCreateNestedPath() throws StageException {
    Stage stage = stageWithInlineScript("""
      doc.val1 = doc.tags[1];
      doc.a = {"b": {"c": {"d": doc.val1}}};      
      delete doc.val1;      
    """);

    Document d = Document.create("doc1");
    d.update("tags", UpdateMode.OVERWRITE, "x", "y", "z");
    stage.processDocument(d);

    assertEquals("y", d.getNestedJson("a.b.c.d").asText());
  }
```

Where do we stand on supporting the convenience syntax `doc.a.b.c.d = 1` ?

The following test fails -- if indeed we don't support this syntax, should add the test and expect and exception, adding a comment saying we don't support this (yet?)

```
  @Test
  public void testCreateNestePathConvenientSyntax() throws StageException {
    Stage stage = stageWithInlineScript("""
      doc.val1 = doc.tags[1];
      doc.a.b.c.d = doc.val1;
      delete doc.val1;      
    """);

    Document d = Document.create("doc1");
    d.update("tags", UpdateMode.OVERWRITE, "x", "y", "z");
    stage.processDocument(d);

    assertEquals("y", d.getNestedJson("a.b.c.d").asText());
  }
```

PR #369, opened 2025-09-16
LC-730 Fix useKafka casing issue and make param case insensitive

spencersolomon6:
useKafka or usekafka are both accepted but documentation says useKafka

### DIFF
--- a/lucille-examples/lucille-distributed-example/runner/scripts/run.sh
+++ b/lucille-examples/lucille-distributed-example/runner/scripts/run.sh
@@ -1,2 +1,2 @@
 #!/bin/sh
-java -Dconfig.file=conf/main.conf -cp 'target/lib/*' com.kmwllc.lucille.core.Runner -useKafka
+java -Dconfig.file=/conf/main.conf -cp '/target/lib/*' com.kmwllc.lucille.core.Runner -useKafka


lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -80,20 +80,21 @@ private Runner() {
...
    * <p>
    * -render: prints out the effective/actual config in the exact form it will be seen by Lucille during the run
    */
   public static void main(String[] args) throws Exception {
     Options cliOptions = new Options()
-        .addOption(Option.builder("usekafka").hasArg(false)
-            .desc("Use Kafka for inter-component communication and don't execute pipelines locally").build())
+        .addOption(Option.builder("usekafka").hasArg(false).build())
+        .addOption(Option.builder("useKafka").hasArg(false)
+            .desc("Use Kafka for inter-component communication and don't execute pipelines locally. NOTE: This param is not case sensitive.").build())
```
rseitz, line 95: I think the param is _mostly_ case-sensitive -- it's only the k that's not... :) so, could just remove the NOTE

rseitz: I think we could get case-insensitivity for all params by applying toLowercase() to all the args before we pass them to the parser here       `cli = new DefaultParser().parse(cliOptions, args);`

rseitz (approved):

PR #370, opened 2025-09-18
LC-734: support array syntax when getting/setting nested json

rseitz:


lucille-core/src/main/java/com/kmwllc/lucille/core/Document.java
```diff
@@ -577,39 +579,125 @@ static Document createFromJson(String json, UnaryOperator<String> idUpdater) thr
...
+      if (segments == null || segments.isEmpty()) {
+        return "";
+      }
+      if (segments.size() == 1) {
+        return segments.get(0).name;
+      }
+      StringBuffer result = new StringBuffer();
+      boolean first = true;
+      for (Segment segment : segments) {
+        if (segment.isIndex()) {
```
kiratraynor, line 660: I think if the first segment is an index or the only segment, it should throw an IllegalArgumentException since we're not allowing that in parse.

rseitz: I'm thinking of stringify as just a way to show what you have in a given array of Segments... it's not meant to enforce that the array of Segments is legal.  I.e. parse should enforce, stringify should just reflect what's there...  though let me know if this doesn't make sense

HenryCaldwell (approved):

spencersolomon6 (approved):

PR #372, opened 2025-09-25
LC-735: Add support to conditions for reading from a cloud provider

HenryCaldwell:


lucille-core/src/main/java/com/kmwllc/lucille/core/Condition.java
```diff
@@ -35,10 +38,36 @@ public static Operator get(String operator) {
   }
 
   public Condition(Config config) {
-    this(config.getStringList("fields"), createValueSet(config),
+    this(config.getStringList("fields"),
+        config.hasPath("values_path") ? loadValuesFromPath(config) : createValueSet(config),
         config.hasPath("operator") ? Operator.get(config.getString("operator")) : Operator.MUST);
   }
 
+  private static Set<String> loadValuesFromPath(Config config) {
```
rseitz, line 46: need some test coverage

kiratraynor (approved):
This looks good to me! 
I was just thinking it would be nice to have an example of using this with one of the cloud providers but wasn't sure where would really be best and it is already documented in the example of how to specify them. Maybe we could add a note to the DropDocument stage since it is meant to be used with Conditions but I am not sure if that is necessary.

PR #375, opened 2025-09-29
LC-739: S3 File Traversal

HenryCaldwell:


HenryCaldwell:
Important to note that pathsToStorages has to be pre encoded at the moment or the path won't be recognized. For example if you want to point to the directory "s3://test/folder with spaces", you would need pathsToStorage: ["s3://test/folder%20with%20spaces"]

HenryCaldwell:
Works without any issues with TextExtraction.

PR #377, opened 2025-10-04
LC-745: Add FST flag for custom regex stop characters

HenryCaldwell:
Defaults to whitespace and punctuation stop characters. Setting the custom regex flag to true defaults to only whitespace.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
@@ -9,10 +9,14 @@ import com.kmwllc.lucille.util.FileContentFetcher;
 import com.kmwllc.lucille.util.StageUtils;
 import com.opencsv.CSVReader;
 import com.typesafe.config.Config;
+import java.io.Reader;
+import java.util.regex.Pattern;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.TokenStream;
-import org.apache.lucene.analysis.standard.StandardAnalyzer;
 import org.apache.lucene.analysis.standard.StandardTokenizer;
+import org.apache.lucene.analysis.core.WhitespaceTokenizer;
+import org.apache.lucene.analysis.core.LowerCaseFilter;
+import org.apache.lucene.analysis.pattern.PatternReplaceCharFilter;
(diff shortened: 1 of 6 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
```diff
@@ -40,6 +44,10 @@
  *   all overlapping matches that start at that position. Defaults to false.</li>
  *   <li>stop_on_hit (Boolean, Optional) : Denotes whether this matcher should stop after one hit. Defaults to false.</li>
  *   <li>ignore_case (Boolean, Optional) : Denotes whether this Stage will ignore case determining when making matches. Defaults to false.</li>
+ *   <li>use_custom_breaks (Boolean, Optional) : If true, tokenization uses a whitespace tokenizer. Optionally, if token_break_regex is provided
```
rseitz, line 47: "use custom breaks" implies that something **custom** is going to be applied, but maybe it isn't

suggest changing this param to use_whitespace_tokenizer
mention that the default of false means "use standard tokenizer"

in constructor, should enforce that token_break_regex cannot be provided unless use_whitespace_tokenizer=true

lucille-core/src/main/java/com/kmwllc/lucille/stage/ExtractEntitiesFST.java
```diff
@@ -40,6 +44,10 @@
  *   all overlapping matches that start at that position. Defaults to false.</li>
  *   <li>stop_on_hit (Boolean, Optional) : Denotes whether this matcher should stop after one hit. Defaults to false.</li>
  *   <li>ignore_case (Boolean, Optional) : Denotes whether this Stage will ignore case determining when making matches. Defaults to false.</li>
+ *   <li>use_whitespace_tokenizer (Boolean, Optional) : If true, tokenization uses a whitespace tokenizer. If token_break_regex is provided
+ *   a pattern replace filter applies first to replace matches with a single space. Defaults to false.</li>
```
rseitz, line 48: should mention that false (or omitted) means the Lucene StandardTokenizer will be used

```
If true, tokenization uses Lucene's WhitespaceTokenizer. If token_break_regex is provided,
 a pattern replace filter applies first to replace matches with a single space. 
If false, tokenization uses Lucene's StandardTokenizer. Defaults to false.
```

kiratraynor (approved):

PR #378, opened 2025-10-07
LC-681: Graal python 

HenryCaldwell:


lucille-core/src/main/java/com/kmwllc/lucille/stage/EmbeddedPythonStage.java
```diff
@@ -0,0 +1,671 @@
...
+      context.eval(source);
+      return null;
+    } catch (Exception e) {
+      throw new StageException("Python failed for doc '" + doc.getId() + "' using '" + scriptPath + "'.", e);
+    }
+  }
+
+  // Proxy implementation exposed to Python
+
+  private static final class PyDocProxy implements ProxyObject, ProxyHashMap {
```
rseitz: it looks like we are supporting two paradigms where the python doc can be treated as a dictionary (accessing fields via `doc['name']` or as an object where the fields are properties that you access via `doc.name`

I think it's OK to move forward supporting both but we should consider whether this might be confusing for a python developer who isn't expecting these paradigms to be combined?

in any case, we should include some documentation here about what we decided and why

lucille-core/src/main/java/com/kmwllc/lucille/stage/EmbeddedPythonStage.java
```diff
@@ -0,0 +1,671 @@
...
+import org.graalvm.polyglot.*;
+import org.graalvm.polyglot.proxy.ProxyArray;
+import org.graalvm.polyglot.proxy.ProxyObject;
+import org.graalvm.polyglot.proxy.ProxyHashMap;
+import org.graalvm.polyglot.proxy.ProxyIterator;
+
+import java.io.Reader;
+import java.util.*;
+
+public class EmbeddedPythonStage extends Stage {
```
rseitz: should be `EmbeddedPython` (no "Stage" in the name)

PR #424, opened 2026-01-20
FIX - FileConnector: it always pushes already published files even if they haven't been modified since last publish

sartian:
There seems to be basic incremental search logic missing from TraversalParams code to make sure a document is a candidate for index update due to having a newer modification date than last published date.

If a document was never published, it should automatically be a candidate for publishing, but if it was previously published and hasn't been modified since, it shouldn't need an update.

A full update can be forced by deleting state information before an index, which should cause all document to look new. 

Expected Behavior: 
1st Ingest: 6 documents ingested, 6 documents published
2nd ingest: 6 documents (or urls) ingested, not were modified, 0 documents published
3rd ingest: 6 documents (urls) ingestsed, one was modified, 1 document published

Behavior Seen:
Documents ingested, not cutoffs, docs always published


meghanboyd:
Closing this PR as it will be subsumed by https://github.com/kmwtechnology/lucille/pull/451

PR #451, opened 2026-03-13
LC-801: FileConnector incremental mode

Otto-Cline:
This is the larger of two PRs based on a large PR that is being broken into parts.

**Larger PR:**
https://github.com/kmwtechnology/lucille/pull/424

**Other branch from this PR:**
https://github.com/kmwtechnology/lucille/pull/450

This part introduces full support for incremental updates with the FileConnector, including proper tombstone and time zone handling.

Note: once the smaller branch is merged to main, this branch can be rebased onto main to ensure the changes from the smaller PR do not show up in this PR. The reason they show up in the first place is because this branch was created off of our original PR, which has both the smaller branch's changes along with this branch's changes.

Minor edits have been made on this branch with comparison to the relevant portion of the original PR, including some new test coverage and Javadoc.


lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -190,8 +190,11 @@ public void execute(Publisher publisher) throws ConnectorException {
         traverseStoragePath(publisher, resource);
       }
 
-      // find files no longer in datastore that need to be removed from index
-      sendExpiredFileTombstones(publisher);
+      if (config.hasPath("filterOptions.sendTombstones") &&
+          config.getString("filterOptions.sendTombstones").equals("true")) {
```
rseitz: for the second part of the &&, you can just do `config.getBoolean("filterOptions.sendTombstones")`

Otto-Cline:
To-do:
Doc site documentation for how tombstones work, toggle

lucille-core/src/test/java/com/kmwllc/lucille/connector/storageclient/TraversalParamsTest.java
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  @Test
+  public void oneMsNewerIncludes() {
+    // any forward movement in modified time should trigger inclusion.
+    assertTrue(incrementalParams().includeFile("doc.json", BASE_TIME.plusMillis(1), BASE_TIME));
+  }
+
+  @Test
+  public void skewedModifiedSkips() {
```
rseitz: is this testing something different from olderModifiedSkips() ?

both seem to be scenarios where last modified is earlier than last published

lucille-core/src/test/resources/FileConnectorTest/tombstonesOff.conf
```diff
@@ -0,0 +1,31 @@
+{
```
rseitz: this file seems to be identical to stateMultiplePaths.conf?
if there's a reason we need to maintain them separately, please add a comment explaining that somewhere

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
```diff
@@ -309,11 +381,335 @@ public void testTraversalWithStateAndMultiplePaths() throws Exception {
...
+        Files.delete(file2.toPath());
+      }
+      if (file3.exists()) {
+        Files.delete(file3.toPath());
+      }
+      if (tempDir.exists()) {
+        tempDir.delete();
+      }
+
+      File stateDir = new File("state");
```
rseitz: can use FileUtils.deleteDirectory() to simplify this

also, wondering, why don't we also need to delete the state directory after other tests?

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -165,6 +171,27 @@ public void shutdown() throws SQLException {
...
+   * transient upstream source flakiness doesn't immediately cause content to be
+   * deleted from index
+   * - expiredFileRetryCutoff [1..N]
+   *   file has to be missing N times before marked as expired)
+   * - expiredFileDurationCutoff [Duration]
+   *   Amount if time file must be missing before marked as expired)
+   */
+  public List<URI> listExpiredFiles() throws SQLException {
+    String selectSQL = "SELECT name FROM \"" + tableName + "\" WHERE encountered = FALSE";
+    var fileUris = new ArrayList<URI>();
```
rseitz: pls. avoid use of var

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -165,6 +171,27 @@ public void shutdown() throws SQLException {
     }
   }
 
+  /*
+   * Needed to support publishing document tombstones suggestion: we should add
```
rseitz: expecting to see javadoc saying what this method does, but the text here is a suggestion/todo. Do we need to act on the suggest? If so, create ticket for it; otherwise suggest removing or moving to an inline comment (not javadoc).

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseFileReference.java
```diff
@@ -98,24 +101,13 @@ public Document decompressedFileAsDoc(InputStream in, String decompressedFullPat
...
-  }
-
   /**
    * Creates an empty document for this FileReference. Uses the given full path and params to create an appropriate docId.
    * @param params Parameters for your storage traversal.
    * @param fullPathString The full path String to the file reference
    * @return An empty Document with an appropriate docId representing this file reference.
    */
-  protected Document createEmptyDocument(TraversalParams params, String fullPathString) {
+  protected static Document createEmptyDocument(TraversalParams params, String fullPathString) {
```
rseitz: remove this method and copy the two lines to the one place where we're calling it buildBaseDoc

it's overkill to have a separate method with its own extensive javadoc just to do this, and "protected static" is confusing is a potentially confusing combination

doc/site/content/en/docs/architecture/components/Connectors/file_connector.md
```diff
@@ -58,4 +58,22 @@ can be embedded, or it can be remote.
...
\ No newline at end of file
+3. Each database table should be used for only one connector configuration.
+
+### Publishing Options
+* `publishMode`: The connector can be configured to have a publishing mode of `FULL` or `INCREMENTAL`, which are relevant on the second and subsequent runs of Lucille. 
+`FULL` mode will republish every document your connector encounters. `INCREMENTAL` mode (requires state management with a JDBC database) will only publish files that are 
+new or have been modified since the last time the connector ran.
+* `sendTombstones`:  By default, if a file in the file system is deleted, documents from the file will still appear in the index until they are manually deleted.
+In order for these deletions to be recognized automatically, we can set `sendTombstones` to be `true`. This will create a tombstone with field `file_expired: true`
+for each deleted document. These can be deleted in the indexer using:
```
rseitz: replace "These can be deleted in the indexer using:" with something like

"The indexer can be configured to issue deletes against the search backend when tombstones are encountered. This only works with indexers such as SolrIndexer and OpenSearchIndexer that respect the deletionMarkerField configuration parameter. Example:"

lucille-core/src/test/resources/FileConnectorTest/EndToEndFileTraversal/file1.txt
```diff
@@ -0,0 +1 @@
+File 1 text
```
rseitz: what tests are referencing this file and/or the others in the new EndToEndFileTraversal directory? I'm not seeing any references.

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
Otto-Cline: For this test I opted to use a config with state: { } to trigger the state db to be created on disk rather than in memory (like in other tests) as a more comprehensive test of tombstones, so we need to clean that up at the end. I can see why this might be inconsistent; added comments for clarification but if you think we should stick to in memory for this test I'll change it.

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
```diff
@@ -24,11 +19,21 @@
...
+import com.kmwllc.lucille.core.Connector;
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.Publisher;
+import com.kmwllc.lucille.core.PublisherImpl;
+import com.kmwllc.lucille.message.TestMessenger;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+import com.typesafe.config.ConfigValueFactory;
 
 public class FileConnectorStateManagerTest {
```
rseitz, line 39: there are 22 occurrences of "new FileConnector(" here

This class (FileConnectorStateManagerTest) should be limited to tests where we're instantiating FileConnectorStateManager and interacting with it directly

If we're creating a FileConnector and testing functionality that it exposes, the test should go in FileConnectorTest instead

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
rseitz: having some tests with state on disk is a good idea -- fine to leave this

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
Otto-Cline: I was halfway through transferring the tests over when I saw that several instances of this are tests that we had prior to this PR, which I remember is why I put the new tests using FileConnector here rather than in the FileConnectorTest class. I'm not opinionated either way but wanted to confirm we still want to move everything over knowing that

rseitz: Thanks for pointing that out -- I do see 3 refererences to new FileConnector() in FileConnectorStateManagerTest in main (i.e. before this PR).

It's a judgement call:

An end-to-end test of FileConnector should go in FileConnectorTest. If the test flexes the state-related functionality but never creates a FileConnectorStateManager and/or never looks inside the state database, again, it still belongs in FileConnectorTest.

A test that uses FileConnector to create state but then looks inside the state database and makes assertions about the database contents should go in FileConnectorStateManagerTest. At this point, the test is dependent on implementation details of FileConnectorStateManager.

Its a question of: are we testing FileConnector itself or are we just using FileConnector to set up a test for something detailed and specific about FileConnectorStateManager.

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
Otto-Cline: Ticket created

rseitz: thanks for creating a ticket. the other task is to remove the comment here

It's a bit unusual for a public method to have a multi-line comment above the declaration that's not actually javadoc (because it doesn't begin with /**) -- the expectation on a public method is that you'd have real javadoc above the declaration and it would start by explaining what the method does, not focus on todo & misc stuff

doc/site/content/en/docs/architecture/components/Connectors/file_connector.md
```diff
@@ -64,8 +64,17 @@ It's important to note that File Connector state is designed to be efficient and
...
-called tombstones. These documents will have a field `file_expired: true` making them easy to delete, and will automatically be set as skipped by all stages in the pipeline.
-So, if we wanted to have these tombstones automatically deleted once they get to the indexer, and we're using Solr or OpenSearch, we might do this in the indexer section of 
-our config: `indexer.deletionMarkerField: "file_expired"` and `indexer.deletionMarkerFieldValue: "true"`.
\ No newline at end of file
+* `sendTombstones`:  By default, if a file in the file system is deleted, documents from the file will still appear in the index until they are manually deleted.
+In order for these deletions to be recognized automatically, we can set `sendTombstones` to be `true`. This will create a tombstone with field `file_expired: true`
+for each deleted document. The indexer can be configured to issue deletes against the search backend when tombstones are encountered. This only works with indexers such 
+* as SolrIndexer and OpenSearchIndexer that respect the deletionMarkerField configuration parameter. Example:
```
spencersolomon6: extra `*` here

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/TraversalParams.java
```diff
@@ -90,6 +91,8 @@ public TraversalParams(Config config, URI pathToStorage, String docIdPrefix) {
         filterOptions.getStringList("excludes") : Collections.emptyList();
     this.excludes = excludeRegex.stream().map(Pattern::compile).collect(Collectors.toList());
 
+    this.publishMode = PublishMode.fromString(filterOptions.hasPath("publishMode") ? filterOptions.getString("publishMode") : "full");
```
rseitz: very minor but but you can avoid using the hardcoded string "full"

```
    this.publishMode = 
        filterOptions.hasPath("publishMode") ? PublishMode.fromString(filterOptions.getString("publishMode")) : PublishMode.FULL;
```

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -206,11 +247,10 @@ public Instant getLastPublished(String fullPathStr) {
...
         if (rs.next()) {
-          Timestamp timestamp = rs.getTimestamp("last_published");
-
-          if (timestamp != null) {
-            return timestamp.toInstant();
-          }
+          // Get timestamp as OffsetDateTime, which is stored in UTC
+          // Note: H2 may convert to local timezone on retrieval, but toInstant() returns the correct absolute point in time.
+          OffsetDateTime odt = rs.getObject("last_published", OffsetDateTime.class);
+          if (odt != null) { return odt.toInstant(); }
```
rseitz: very minor but per our code conventions the conditional body should always start on a separate line

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorTest.java
```diff
@@ -434,4 +439,237 @@ public void testLocalTraversal_DirWithSpacesAndSpecials_FromConf() throws Except
...
+      long firstRunTombstones = messenger1.getDocsSentForProcessing().stream()
+          .filter(doc -> Boolean.TRUE.equals(doc.getBoolean(FileConnector.EXPIRED)))
+          .count();
+      assertEquals(0, firstRunTombstones);
+
+      // Delete file2, modify file1, leave file3 alone.
+      // Set file1's modification time explicitly to a point after the first run's traversalInstant just in case
+      Files.delete(file2.toPath());
+      Files.writeString(file1.toPath(), "Modified Content 1");
+      Files.setLastModifiedTime(file1.toPath(), java.nio.file.attribute.FileTime.from(Instant.now().plusSeconds(5)));
```
rseitz: Instant.now() should be sufficient -- we shouldn't need to set the modification time to be 5 seconds in the future

rseitz: also, should import java.nio.file.attribute.FileTime

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorTest.java
```diff
@@ -434,4 +439,237 @@ public void testLocalTraversal_DirWithSpacesAndSpecials_FromConf() throws Except
...
+      assertTrue(tombstone.getString(FileConnector.FILE_PATH).contains("file2.txt"));
+      assertTrue(tombstone.isSkipped());
+
+      // Verify non-tombstone is for modified file1
+      Document modifiedDoc = secondRunDocs.stream()
+          .filter(doc -> !Boolean.TRUE.equals(doc.getBoolean(FileConnector.EXPIRED)))
+          .findFirst()
+          .orElseThrow();
+      assertTrue(modifiedDoc.getString(FileConnector.FILE_PATH).contains("file1.txt"));
+
```
rseitz, line 689: add a third run here (this only works if the modification time on l635 is not in the future)

```
      // Third run, should not publish anything
      TestMessenger messenger3 = new TestMessenger();
      Publisher publisher3 = new PublisherImpl(config, messenger3, "run1", "pipeline1");
      Connector connector3 = new FileConnector(config);
      connector3.execute(publisher3);
      assertEquals(0, messenger3.getDocsSentForProcessing().size());
```

rseitz: I realized to be more thorough we should restore file2 and make sure a non-tombstone is sent for it (confirming there isn't any problem with republishing files that have been deleted previously)... and then we should delete it a second time and confirm that a tombstone is sent again.

```
      // Third run, should not publish anything
      TestMessenger messenger3 = new TestMessenger();
      Publisher publisher3 = new PublisherImpl(config, messenger3, "run1", "pipeline1");
      Connector connector3 = new FileConnector(config);
      connector3.execute(publisher3);
      assertEquals(0, messenger3.getDocsSentForProcessing().size());

      Files.writeString(file2.toPath(), "Content 2 Restored");

      // Fourth run, should publish file2 which has now been restored after being deleted earlier
      TestMessenger messenger4 = new TestMessenger();
      Publisher publisher4 = new PublisherImpl(config, messenger4, "run1", "pipeline1");
      Connector connector4 = new FileConnector(config);
      connector4.execute(publisher4);
      List<Document> fourthRunDocs = messenger4.getDocsSentForProcessing();
      assertEquals(1, fourthRunDocs.size());

      // Verify non-tombstone is for restored file2
      Document restoredDoc = fourthRunDocs.stream()
          .filter(doc -> !Boolean.TRUE.equals(doc.getBoolean(FileConnector.EXPIRED)))
          .findFirst()
          .orElseThrow();
      assertTrue(restoredDoc.getString(FileConnector.FILE_PATH).contains("file2.txt"));

      Files.delete(file2.toPath());

      // Fifth run, should publish tombstone for file2 which has been deleted a second time
      TestMessenger messenger5 = new TestMessenger();
      Publisher publisher5 = new PublisherImpl(config, messenger5, "run1", "pipeline1");
      Connector connector5 = new FileConnector(config);
      connector5.execute(publisher5);
      List<Document> fifthRunDocs = messenger5.getDocsSentForProcessing();
      assertEquals(1, fifthRunDocs.size());

      // Verify tombstone is for file2
      Document tombstone2 = fifthRunDocs.stream()
          .filter(doc -> Boolean.TRUE.equals(doc.getBoolean(FileConnector.EXPIRED)))
          .findFirst()
          .orElseThrow();
      assertTrue(tombstone2.getString(FileConnector.FILE_PATH).contains("file2.txt"));
      assertTrue(tombstone2.isSkipped());
```

rseitz (approved):
approving
have left one more comment re: testTombstonesWithFileSystemChanges()
would like a review from @kyle-kmw before we merge this PR

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -166,56 +190,132 @@ public FileConnector(Config config) throws ConnectorException {
...
+    }
+  }
 
-      if (stateManager != null) {
-        try {
-          stateManager.shutdown();
-        } catch (SQLException e) {
-          log.warn("Error occurred while shutting down FileConnectorStateManager.", e);
-        }
+  private void shutdown() {
```
kyle-kmw: Should the shutdown be wired in to the Connector close() method?

rseitz: Looks like we should rename shutdown() as

```
  @Override
  public void close() 
```

and can then remove the finally block inside execute()

That finally block was in place before this PR, so I think it's reasonable if we want to do this refactor as part of a separate/followup PR. But it could also be convenient to do it here in this PR because we're creating the shutdown method in this PR. 

@Otto-Cline please try this refactor and see if all tests still pass. [Hmm... tests will probably fail because now we'll have to call close() explicitly after we call execute() in tests. Up to you if you want to work through the necessary changes here or separately.]

kyle-kmw (approved):

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
Otto-Cline: Yes working on it rn, it does cause the tests to fail; I will pick it back up later tonight. Would like to include it as part of this PR since we're already updating some decisions around the FileConnector / FCTest that were in the codebase prior to this anyways

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -165,57 +189,129 @@ public FileConnector(Config config) throws ConnectorException {
...
   public void execute(Publisher publisher) throws ConnectorException {
-    try {
-      try {
-        for (StorageClient client : storageClientMap.values()) {
-          client.init();
-        }
-      } catch (IOException e) {
-        throw new ConnectorException("Error initializing a StorageClient.", e);
-      }
+      initialize();
```
rseitz: update indentation now that you removed the try/catch

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorTest.java
```diff
@@ -301,7 +312,7 @@ public void testMultiplePathsSameClient() throws Exception {
     Config config = ConfigFactory.parseResourcesAnySyntax("FileConnectorTest/multiplePathsLocal.conf");
     TestMessenger messenger = new TestMessenger();
     Publisher publisher = new PublisherImpl(config, messenger, "run", "pipeline1");
-    Connector connector = new FileConnector(config);
+    connector = new FileConnector(config);
```
rseitz, line 315: I realize this last Connector will be closed in `@After`
But since this test is creating and closing multiple Connectors I would opt to handle all of them in a consistent way

First declaration should be `Connector connector = ` so we are using a local variable
then call connector.close() on the last connector like we do on all the others
Same comment holds for testMultiplePathsDifferentClients() and anywhere else we're creating/closing multiple connectors

To avoid confusion between the instance variable `connector` and a local variable of the same name, you could change the local variable to be `conn`

PR #459, opened 2026-04-12
LC-836: automated versioned docs

Otto-Cline:
We would like to have our doc site provide a version selector with the docs for every release of Lucille.

This requires updating our hugo.yml file with a step in our build job which:
- Iterates through all release tags
- Adds the doc site from each to our checked out repo

And another step that updates the toml


Otto-Cline:
Need to remove docs-0.8.0 from content folder

doc/site/versions.sh
```diff
@@ -2,8 +2,11 @@
 rm -rf content/en/docs-*
 git checkout HEAD -- hugo.toml
 
-# Get the latest tag so we can skip it, its docs are already the main /docs/.
-latest_tag=$(git tag --list --sort=-version:refname | grep -v '^v' | head -1)
+tags=$(git tag --list)
+tags_descending=$(git tag --list --sort=-version:refname)
```
spencersolomon6: I think we can just always use the `tags_descending` instead of calling `git tag --list` twice here

.github/workflows/hugo.yml
```diff
@@ -52,6 +54,117 @@ jobs:
       - name: Install postcss
         run: npm install postcss-cli
         working-directory: doc/site
+
+      - name: Generate versioned docs and update hugo.toml
+        working-directory: doc/site
+        env:
+          BASE_URL: ${{ steps.pages.outputs.base_url }}
+        run: |
```
spencersolomon6: I would like to see this big wall of bash split up into smaller logical pieces, ideally as small as possible. We should have many small steps, as it makes these files easier to read, is more inline with how our other Github Actions files are formatted and will give more granular logging in the case of an error occurring somewhere in the action.

.github/workflows/hugo.yml
```diff
@@ -11,7 +11,8 @@ on:
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
-  contents: read
+  # change contents to write access so we can add the version docs
+  contents: write
```
spencersolomon6, line 15: [curiosity] Why was this not necessary before?

.github/workflows/hugo.yml
```diff
@@ -43,6 +44,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           submodules: recursive
+          fetch-depth: 0
```
spencersolomon6, line 48: [curiosity] What is the purpose of this?

.github/workflows/hugo.yml
Otto-Cline: No code in the working directory was changed before. There could be something to be said about the build generating new files with the public folder, but I'm assuming since it's done indirectly by hugo itself write access wasn't needed

.github/workflows/hugo.yml
Otto-Cline: I believe it just says that we have access to the full history of commits and tags; actions/checkout defaults to fetch-depth: 1, which only fetches the current commit

Otto-Cline: To elaborate, the workflow clones the repo to the vm every time it runs, so the default fetch-depth: 1 is only getting the latest commit to save resources. Since we need everything to loop through our tags, we need full commit history

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,146 @@
...
+
+# Keep in mind that the latest version of the docs here will not reflect your changes to the doc site content
+# folder in the working directory, because it pulls from the latest release. Your changes will not appear until
+# the next release.
+
+# Also keep in mind that this script will alter the hugo.toml file.
+# MAKE SURE TO REVERT THE hugo.toml BEFORE COMMITTING IF YOU RUN THIS SCRIPT.
+
+# Pre-run cleanup if running in local mode, not necessary for production
+cd ${GITHUB_WORKSPACE}/doc/site
```
spencersolomon6: This cd should only be performed in production mode, can be removed here

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,146 @@
...
+  url="https://kmwtechnology.github.io/lucille/docs"
+else
+  echo "Invalid mode specified: $mode"
+  exit 1
+fi
+
+# Remove any version folders from previous runs
+rm -rf content/en/docs-*
+# Restore hugo.toml in case there are any leftover changes from last run
+git checkout HEAD -- hugo.toml
```
spencersolomon6: These can be removed, only need to happen on local run

doc/site/versions.sh
spencersolomon6, line 1: This file can be removed right? We should instead add some instructions in the doc/site README about how to build the full versioned docs locally using the workflow script

doc/site/hugo.toml
```diff
@@ -135,7 +127,7 @@ offlineSearch = true
 # Enable syntax highlighting and copy buttons on code blocks with Prism
 prism_syntax_highlighting = false
 
-# User interface configuration
```
spencersolomon6, line 138: Why was this removed?

.github/scripts/versionsworkflow.sh
Otto-Cline: sorry dont know how I missed that and the last comment haha

doc/site/hugo.toml
Otto-Cline: Found it a bit weird since the versions get put directly above [params.ui] (and thus underneath the comment) so it no longer makes sense when you run the script, but I guess since it would be rare that someone runs it locally, we could just keep it or put it right below [params.ui]

spencersolomon6: Lets try to just append the versions block at the start or end of the script, without anchoring it to any code. If that doesn't work, lets add a comment whose only purpose is to be used as a flag for where to place the versions block. MAKE IT VERY EXPLICIT THE PURPOSE OF THIS COMMENT.

spencersolomon6 (approved):

doc/site/layouts/partials/sidebar.html
```diff
@@ -0,0 +1,101 @@
+{{/* Notes:
```
rseitz, line 1: Looks like this is a copy of the default docsy sidebar, presumaby with some customizations, but I can't tell what the customizations are. It would be helpful if you made a commit with the original/unmodified version of this file from docsy, then made your changes in a second commit. The commit comment would ideally say what version of docsy you got the file from and/or include a link to the raw file in the docsy repo in the appropriate branch/tag. Right now I am looking here: https://github.com/google/docsy/blob/main/layouts/_partials/sidebar.html

I notice that a lot of indentation present in the docsy repo has been lost here, which makes the file hard to read. Please make sure to keep the indentation.

This comment applies to navbar.html as well

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,140 @@
...
+else
+  echo "Invalid mode specified: $mode"
+  exit 1
+fi
+
+# Minimum version to expose on the docs site. Tags older than this are
+# skipped entirely, they get no docs-<tag> folder and no dropdown entry.
+min_version="0.5.10"
+
+# Helper which returns success if $1 a tag is at or above $min_version. Strips leading v for v0.2.4
```
rseitz: let's rename the v0.2.4 tag instead of adding special handling for it in this script

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,140 @@
...
+# [[params.versions]] blocks are left alone).
+
+awk '
+  /^# PLACEHOLDER/{while((getline line < "/tmp/versions_block.txt") > 0) print line; next}
+  /^version = /{next}
+  /^url_latest_version = /{next}
+  {print}
+' hugo.toml > hugo.toml.tmp && mv hugo.toml.tmp hugo.toml
+
+# We can skip the cleanup if deploying on our production VM
```
rseitz: I wouldn't say "our production VM"
we don't have a production VM

.github/scripts/versionsworkflow.sh
Otto-Cline: Tags unfortunately cannot be renamed in github/git

Otto-Cline: Resolved after v0.2.4 was copied and deleted to fix naming issue, relevant code change made as well

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,132 @@
...
+  # Check if docs exist in this tag
+  git ls-tree "$tag" content/en/docs >/dev/null 2>&1 || { echo "Skipping $tag - docs not found"; continue; }
+
+  # Make temporary directory for git archive to use, then archive the tag into it
+  tmpdir=$(mktemp -d)
+  git archive "$tag" content/en/docs | tar -x -C "$tmpdir"
+
+  if [ "$tag" = "$latest_tag" ]; then
+    # Latest tag replaces the main docs/ folder
+    rm -rf content/en/docs
```
rseitz, line 71: there's an "rm -rf" here as we as lines 22 and 130
comment at the top of this file should be ultra-clear that this is a destructive script which will delete any changes a developer might have made inside content/en/docs
it should ONLY be run by a developer if they are working in a fresh checkout that does not have any local changes they want to keep

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,136 @@
...
+
+# Filtered list of tags at or above min_version. sort -V puts tags in version
+# order, then awk emits everything from min_version onward.
+tags=$(git tag --list | sort -V | awk -v min="$min_version" '$0 == min { found=1 } found')
+
+# Descending (newest-first) list used for the dropdown selector order
+tags_descending=$(echo "$tags" | sort -rV)
+
+# Get the latest tag; its docs will replace the working directory's docs/
+latest_tag=$(echo "$tags_descending" | grep -v '^v' | head -1)
```
rseitz: this is still filtering out tags that start with v

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,136 @@
...
+# [[params.versions]] blocks are left alone).
+
+awk '
+  /^# PLACEHOLDER/{while((getline line < "/tmp/versions_block.txt") > 0) print line; next}
+  /^version = /{next}
+  /^url_latest_version = /{next}
+  {print}
+' hugo.toml > hugo.toml.tmp && mv hugo.toml.tmp hugo.toml
+
+# We can skip the cleanup if deploying to prod
```
rseitz: This comment still really threw me off. It's not that we "can" skip the cleanup if we're deploying to prod. We HAVE to skip cleanup here if deploying to prod because the docs-* directories will be needed later in hugo.yml when the hugo build actually happens.

I would suggest not doing the build here in local mode and also not doing the cleanup in local mode. Let the developer run this versionworkflow.sh script, then run "hugo build" and decide if they want to look over the "docs-*" directories or delete them manually. We know they will already have to revert hugo.toml so this was never going to be a fully "clean" process to begin with.

doc/site/hugo.toml
```diff
@@ -135,6 +127,10 @@ offlineSearch = true
 # Enable syntax highlighting and copy buttons on code blocks with Prism
 prism_syntax_highlighting = false
 
+
+# The following line is a placeholder for where the version generation script will place the version blocks. Do not change or remove it.
+# PLACEHOLDER
```
rseitz: placeholder is a pretty common word to use in a file like this. 
It's unlikely that someone would try to put it in another place in this file in all caps, but conceivable
I would change it to something more specific like 
VERSION_PLACEHOLDER or 
VERSION_GENERATION_SCRIPT_PLACEHOLDER

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,136 @@
...
+
+# Keep in mind that the latest version of the docs here will not reflect your changes to the doc site content
+# folder in the working directory, because it pulls from the latest release. Your changes will not appear until
+# the next release.
+
+# Also keep in mind that this script will alter the hugo.toml file.
+# MAKE SURE TO REVERT THE hugo.toml BEFORE COMMITTING IF YOU RUN THIS SCRIPT.
+
+# Pre-run cleanup if running in local mode, not necessary for production
+mode=${1:-"local"}
```
rseitz, line 28: I would just generally try to make local mode and production mode as similar as possible.
Looks like in local vs. production we need to cd into different directories and set urls, but those are the only necessary differences.

In local mode I would leave out "rm -rf content/en/docs-*" and "git checkout HEAD -- hugo.toml" here and I would leave out the entire local mode block at the end. This script should just be focused on ONE THING -- taking an existing environment and setting it up (downloading tagged docs, updating hugo.toml) so that someone who runs "hugo build" afterwords will see the versioned docs. The pre and post cleanup can be left to the developer. In practice I don't think devs are going to be running this very often so it's not that much of a burden. And it's safer to leave anything with "rm -rf" to a developer to execute manually when they want (even though we did up the readme to warn about this, it would be even better to not have to warn about it but just tell the developer when they should do it on their own)

.github/scripts/versionsworkflow.sh
Otto-Cline: Acknowledged for the first paragraph, but I'm not sure about not including the build and cleanup; I generally think it's a good idea to have that. Maybe @spencersolomon6 has thoughts

.github/scripts/versionsworkflow.sh
Otto-Cline: Reading this comment I see where you're coming from more, will update

.github/scripts/versionsworkflow.sh
```diff
@@ -0,0 +1,127 @@
+# NOTE: This script will delete any changes made inside doc/site/content/en/docs.
```
rseitz, line 9: First thing I'd want to see here is a summary of what this script does. 

This script updates the working tree with all the content needed for "hugo build" to generate versioned docs. To do this:

- It deletes doc/site/content/en/docs and replaces it with docs from the latest tagged release.
- It uses git to get the docs from prior releases and places them in versioned fulders under doc/site/content/en
- It updates hugo.toml with information about the versions

rseitz (approved):
Approving with one last comment about the explanation at the top of versionworkflow.sh

.github/scripts/versionsworkflow.sh
Otto-Cline: Updated, lmk if too verbose / long. Wanted to err on the side of descriptiveness

rseitz: looks good, thx

PR #463, opened 2026-04-24
LC-843: Adding incremental backoff logic to OpenSearchUtils

Otto-Cline:
If an OpenSearch cluster is under heavy load we will sometimes get a error code when trying to send batches from the Lucille indexer:
`429 too many requests`

When this error is received today, the Lucille indexer just fails the current batch and sends a new batch of content. This behavior often just results in more failed batches.

This PR tells the indexer to engage in incremental backoffs when a 429 or 503 error is received using the setRetryStrategy() that comes with HttpClientBuilder.

We would like to test this using a local OpenSearch instance, as well as a customer instance for real verifiability.


Otto-Cline:
Closed, superseded by [LC-862](https://github.com/kmwtechnology/lucille/pull/468)

PR #464, opened 2026-04-26
LC-859: coding style improvements

rseitz:
This PR contains the result of several AI-assisted passes over the codebase to look for places where the Java coding style could be improved. The intent is that all these changes are minor and have no impact on functionality. The common changes include:

- declaring fields final where possible
- using string substitution instead of concatenation in log messages
- using ConfigUtils.getOrDefault instead of a conditional 
- using modern instanceof syntax
- declaring List instead of ArrayList
- preferring collection.isEmpty() over collection.size()==0
- avoiding wildcard imports


lucille-core/src/main/java/com/kmwllc/lucille/core/KafkaDocument.java
```diff
@@ -77,9 +77,7 @@ public boolean equals(Object other) {
     if (this == other) {
       return true;
     }
-    if (other instanceof KafkaDocument) {
-      KafkaDocument doc = (KafkaDocument) other;
-
+    if (other instanceof KafkaDocument doc) {
```
Otto-Cline, line 80: I notice that a main theme of this PR is using pattern matching for instanceof, but looking through the branch it seems like this isn't being applied everywhere it could. Might be something to consider

lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
```diff
@@ -39,6 +39,9 @@ public class PublisherImpl implements Publisher {
   private static final Logger log = LoggerFactory.getLogger(PublisherImpl.class);
   private static final Logger docLogger = LoggerFactory.getLogger("com.kmwllc.lucille.core.DocLogger");
 
+  /** Timeout in seconds for re-checking the pending-docs condition when maxPendingDocs is set. */
+  private static final int PENDING_DOCS_POLL_TIMEOUT_SECONDS = 10;
```
Otto-Cline, line 43: We only use this in one place, is this change necessary?

lucille-core/src/main/java/com/kmwllc/lucille/core/RunResult.java
```diff
@@ -75,9 +76,9 @@ private static String formatMessage(boolean status, List<Connector> connectors,
       sb.append(result.toString());
       sb.append("\n");
     }
-    if (connectors.size() > 0 && connectors.size() > connectorResults.size()) {
```
Otto-Cline, line 78: is there a reason we were checking the size is greater than 0 in the first place

Otto-Cline (approved):

PR #468, opened 2026-04-27
LC-862: Add ability for Indexer to retry a batch

rseitz:
- Indexer base class now wraps the abstract sendToIndex call in a retry handler implemented via a new dependency, resilience4j
- Indexer accepts new optional params maxRetries, retryWaitDurationMs, retryableStatusCodes; default behavior when these params are not set is no retries
- For a retry to occur, the Indexer subclass must throw IndexerRetryableException from sendToIndex() and the exception must contain one of the status codes configured as retryable
- OpenSearchIndexer is updated in this PR to throw IndexerRetryableException where appropriate; ElasticsearchIndexer and SolrIndexer have NOT been updated and this work should be tracked in separate tickets
- Test coverage includes testRetryAgainstLiveServer() which starts an embedded HttpServer, important for validating that retry logic is actually working, since mocking can give a false sense of security in this context

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
@@ -371,6 +371,10 @@ public abstract class Indexer implements Runnable {
     try {
       stopWatch.reset();
       stopWatch.start();
+      // Note: the retry wraps the entire sendToIndex() call. If sendToIndex() partially succeeds
+      // (e.g. some documents indexed before a subsequent delete-by-query fails), a retry will
+      // re-execute the entire method. This is considered safe because search engine upserts are idempotent —
+      // re-indexing an already-indexed document produces the same result.
       Set<Pair<Document, String>> failedDocPairs = retry != null
           ? Retry.decorateCheckedSupplier(retry, () -> sendToIndex(batchedDocs)).get()
           : sendToIndex(batchedDocs);
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
@@ -80,22 +80,24 @@ import sun.misc.Signal;
  *   <li>indexer.deleteByFieldValue (String, Optional) : Field whose value specifies which documents to delete by query. Must be set
  *   together with indexer.deleteByFieldField.</li>
  *   <li>indexer.maxRetries (Integer, Optional) : Maximum number of times to retry a failed batch before treating all documents
- *   in the batch as failures. Retries are disabled by default ({@value #DEFAULT_MAX_RETRIES}); set to a positive integer to opt in.</li>
+ *   in the batch as failures. Retries are disabled by default; set to a positive integer to opt in. A value of 0 or less
+ *   is rejected as invalid — remove the parameter entirely to disable retries.</li>
  *   <li>indexer.retryWaitDurationMs (Integer, Optional) : Initial wait duration in milliseconds before the first retry.
- *   Subsequent retries use exponential backoff, doubling the wait on each attempt. Only relevant when maxRetries is greater than 0.
+ *   Subsequent retries use exponential backoff, doubling the wait on each attempt. Only allowed when maxRetries is also set.
  *   Defaults to {@value #DEFAULT_RETRY_INITIAL_WAIT_DURATION_MS}.</li>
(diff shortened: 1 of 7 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -186,6 +208,48 @@ public Indexer(Config config, IndexerMessenger messenger, boolean bypass, String
...
+        retryableStatusCodes = config.getIntList("indexer.retryableStatusCodes");
+        if (retryableStatusCodes.isEmpty()) {
+          throw new IllegalArgumentException(
+              "indexer.retryableStatusCodes must not be empty when specified. "
+                  + "Remove the parameter to use the default, or provide at least one status code.");
+        }
+      }
+      this.retry = Retry.of("indexer-batch-retry", RetryConfig.custom()
+          .maxAttempts(maxRetries + 1) // maxAttempts includes the initial attempt
+          .intervalFunction(IntervalFunction.ofExponentialBackoff(retryInitialWaitDurationMs))
```
Otto-Cline: might be nice to have a config option where you can write a custom intervalfunction

Otto-Cline (approved):

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
kyle-kmw: Should this be wrapped with the same logic to wrap certain errors in IndexerRetryableException?

kyle-kmw: Should this be wrapped with the same logic to wrap certain errors in IndexerRetryableException?

kyle-kmw, line 388: I *think* that the item error's can be inspected for 429s at the item level and if the item has an error code should the batch be retried?

rseitz: yes, will do this

rseitz: yes, will do this

rseitz: genuinely not sure! would OpenSearch report a 429 for a specific document inside a batch rather than throwing the exception from the client.bulk() call itself? am thinking we shouldn't add handling for this unless we have evidence that it's a real scenario, otherwise it just complicates the logic. will hold off on this for now but let me know if you have more info on this

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
rseitz: yes, good idea for a followup

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
kyle-kmw: After reading this: https://docs.opensearch.org/latest/api-reference/document-apis/bulk/#partial-responses-and-shard-failures

I do think that there is a key piece missing which is that we are not checking the top level `response.errors()` boolean. "The top-level errors field indicates whether any operations encountered errors, allowing you to quickly determine if you need to examine individual operation results." 

So the bulk request can complete successfully in a scenario where there is nothing preventing opensearch from responding to the request but if there are failures at an individual shard level the request will succeed but the response.error(s) will be true

kyle-kmw: Based on this small test where I created a write block index and attempted to send a batch with an indexer we can see the top level errors boolean is true and the individual items have errors on them with a status code. 

Image


I think it would make sense to include logic similar to this:

```
for (BulkResponseItem item : response.items()) {
    if (item.error() != null) {
        if (retryableStatusCodes.contains(item.status())) {
            // At least one item has a retryable error — throw to trigger retry
            throw new IndexerRetryableException(item.status(),
                "Bulk item returned retryable status " + item.status() + ": " + item.error().reason(),
                null);
        }
        // ... handle non-retryable errors as before
    }
}
```

rseitz:
closing this PR because it is superseded by https://github.com/kmwtechnology/lucille/pull/488

PR #466, opened 2026-04-27
Add PostgresIndexer plugin with upsert, jsonb, and pgvector

seanoc5:
## Summary

- Adds `lucille-postgres` plugin: a JDBC-based indexer that writes Lucille documents to PostgreSQL
- Supports upsert (`ON CONFLICT DO UPDATE`), jsonb, pgvector embedding storage, and array types
- Follows the same plugin architecture as `lucille-pinecone` and `lucille-weaviate` (separate module, shade plugin for fat JAR)
- Single line added to `lucille-plugins/pom.xml` to register the module

## Features

| Feature | Details |
|---------|---------|
| **Upsert** | `INSERT ... ON CONFLICT (id) DO UPDATE` for idempotent writes |
| **Column types** | text, varchar, int, bigint, double, float, boolean, timestamp |
| **jsonb** | Store arbitrary JSON payloads with GIN indexing |
| **pgvector** | `vector(N)` columns with configurable dimensions for embedding storage |
| **Arrays** | text[], int[], bigint[], double[] |
| **Batching** | PreparedStatement batch operations |
| **Safety** | SQL identifier validation rejects unsafe table/column names |

## Config example

```hocon
indexer {
  type: "postgres"
  class: "com.kmwllc.lucille.postgres.indexer.PostgresIndexer"
  batchSize: 100
}

postgres {
  jdbcUrl: "jdbc:postgresql://localhost:5432/mydb"
  user: "lucille"
  password: ${?PG_PASSWORD}
  table: "documents"
  idColumn: "id"
  upsert: true
  columns: [
    { field: "title",      column: "title",     type: "text" }
    { field: "body",       column: "content",   type: "text" }
    { field: "metadata",   column: "meta",      type: "jsonb" }
    { field: "embeddings", column: "embedding", type: "vector", dim: 1024 }
    { field: "tags",       column: "tags",      type: "text[]" }
  ]
}
```

## Dependencies

- `org.postgresql:postgresql:42.7.4` (JDBC driver)
- `com.pgvector:pgvector:0.1.6` (pgvector Java type support)

Both are scoped to the plugin JAR only — no impact on lucille-core.

## Test plan

- [x] 4 unit tests: upsert SQL generation, insert-only SQL, vector dimension validation, unsafe identifier rejection
- [x] All tests pass (`mvn test -pl lucille-plugins/lucille-postgres`)
- [x] Google Java Style compliant (2-space indent, 132-char lines)
[...]

PR #470, opened 2026-04-29
LC-517: HTTPS on Lucille API

Otto-Cline:
This PR overhauls the README for the `lucille-api` folder and changes the example `api.yml` file to have HTTPS and basic auth on by default.

PR #471, opened 2026-04-30
Fix ClassGraph OOM in ConfigInfo by narrowing scan scope

seanoc5:
## Summary

- `ConfigInfo.buildSpecArrayForSubclasses` was using `new ClassGraph().enableAllInfo().scan()`, which scans every annotation, method, and field across **all JARs** on the classpath
- This caused an `OutOfMemoryError: Java heap space` in `ConfigInfoTest.testStages_basicShape` on CI (observed in PR #466's run)
- `getSubclasses()` only needs class hierarchy info — `enableClassInfo()` is sufficient
- Limiting the scan to `com.kmwllc.lucille` packages further reduces memory and scan time

**Before:**
```java
new ClassGraph().enableAllInfo().scan()
```
**After:**
```java
new ClassGraph().enableClassInfo().acceptPackages("com.kmwllc.lucille").scan()
```

> Note: happy for this to be closed and absorbed into other ongoing work if you prefer — just flagging the root cause of the intermittent CI OOM.

## Test plan
- [ ] `ConfigInfoTest` passes in CI without OOM
- [ ] Connector, stage, and indexer lists still return correct results

### DIFF
--- a/lucille-plugins/lucille-api/README.md
+++ b/lucille-plugins/lucille-api/README.md
@@ -47,6 +47,19 @@ Dropwizard can be configured via the `conf/api.yml` file. Details about availabl
 If configuration beyond what Dropwizard supplies is required, the functionality can be extended via the LucilleAPIConfiguration class.
 Details on how can be found [here](https://www.dropwizard.io/en/stable/manual/configuration.html#man-configuration).
 
+### Scanning Custom Components
+
+The `/v1/config-info/*` endpoints discover Connector, Stage, and Indexer implementations by scanning the classpath under
+`com.kmwllc.lucille`. To expose custom implementations that live in other packages, supply a comma-separated list via the
+`scan.extra.packages` system property at startup:
+
+```bash
+java -Dscan.extra.packages=com.example.lucille.custom,org.acme.lucille \
+     -jar target/lucille-api-X.X.X-SNAPSHOT.jar server ${DROPWIZARD_CONF}
+```
+
+Whitespace around entries is trimmed. The built-in `com.kmwllc.lucille` package is always scanned and does not need to be listed.
+
 ### Auth Configuration
 
 Authentication is configured under the `auth` JSONProperty.
--- a/lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/ConfigInfo.java
+++ b/lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/ConfigInfo.java
@@ -176,7 +176,7 @@ public class ConfigInfo {
       throws NoSuchFieldException, IllegalAccessException {
     ArrayNode array = mapper.createArrayNode();
     // Use ClassGraph to scan the classpath
-    try (ScanResult scanResult = new ClassGraph().enableAllInfo().scan()) {
+    try (ScanResult scanResult = new ClassGraph().enableClassInfo().acceptPackages("com.kmwllc.lucille").scan()) {
       // Find every class that is a subclass of the provided class name
       ClassInfoList classes = scanResult.getSubclasses(baseClassName);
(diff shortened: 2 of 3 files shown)


seanoc5:
Note: claude's suggested one-liner change **seems** reasonable, but I did not probe any deeper. It basically only scans the kmw classes, which sounds legit, but I'm going on the llm suggestion more than explicit insight.

rseitz:
> Note: claude's suggested one-liner change **seems** reasonable, but I did not probe any deeper. It basically only scans the kmw classes, which sounds legit, but I'm going on the llm suggestion more than explicit insight.

The change makes sense and also has a performance benefit in the unit test and api lazy-initialization when /stage-list is first called.

However, we want the API to know about any custom Stage, Indexer, and Connector implementations that might be on the classpath but not defined underneath com.kmwllc.lucille

We need a way for a user to specify additional packages that should be scanned. These packages should passed in where we call `acceptPackages("com.kmwllc.lucille")`

Where is a good place for the user to specify these additional packages?

The current README.md says that a lucille config file should be provided when launching the API via `-Dconfig.file=${LUCILLE_CONF}` however this conf file isn't being used by the API currently and another in-flight PR actually edits the readme to remove it from the example invocation. See https://github.com/kmwtechnology/lucille/pull/470/changes#diff-cd371a596f5de79ecc883c34d7d9613c82f7c3bc2a42784ca1260893acf6a8b7R13

It should work to have ConfigInfo take any additional user-specified packages to scan from a Java system property which would be specified via -D at API startup. I'm good with this Claude suggestion of how that approach could be fleshed out:

```
String extraPackages = System.getProperty("scan.extra.packages", "");

String[] packages = Stream.concat(
        Stream.of("com.kmwllc.lucille"),
        Arrays.stream(extraPackages.split(","))
              .map(String::trim)
              .filter(s -> !s.isEmpty())
).toArray(String[]::new);

try (ScanResult scanResult = new ClassGraph()
        .enableClassInfo()
        .acceptPackages(packages)
        .scan()) {
    // ...
}
```

The readme would need to be updated to explain "scan.extra.packages"

We'd also want some test coverage showing that "scan.extra.packages" is respected but we should avoid overkill there... would rather not introduce any Stages etc. in packages outside "com.kmwllc.lucille" into the codebase just to test this unless there's a compact, self-contained way to do it all in a unit test.

seanoc5:
Pushed `1c38af11` taking your suggested approach:

- `ConfigInfo.resolveScanPackages(String)` static helper merges `com.kmwllc.lucille` with comma-separated extras parsed from `-Dscan.extra.packages` (whitespace-trimmed, empty entries skipped)
- `buildSpecArrayForSubclasses` reads the system property and passes the resolved array to `ClassGraph.acceptPackages(...)`
- README section *Scanning Custom Components* documents the `-D` flag with an example invocation
- Tests (9/9 passing): null / empty / whitespace-only / single / multi-with-padding cases for the resolver, plus a smoke test that sets the property to a non-matching package and confirms the default `com.kmwllc.lucille` scan still returns in-tree stages

One **open question** on test coverage: 
the smoke test proves the wiring doesn't break the default scan, but it doesn't *positively* prove that a class in a foreign package would be discovered when listed in `scan.extra.packages`. 

The cleanest way to assert that would be a tiny stub `Stage` under `src/test/java/org/example/extstages/` — test-classpath only, never in the shipped artifact. 
Reading your earlier note _("would rather not introduce any Stages etc. in packages outside com.kmwllc.lucille into the codebase")_ I wasn't sure whether test-scope counts as "in the codebase." 
Happy to add it if you're OK with that, otherwise the resolver unit tests + smoke test seem like a reasonable stopping point.

BTW: sorry for the delay, just noticed this last night, and realized it might be pointing loosely in my direction for follow up 
:- )

PR #475, opened 2026-05-04
LC-861: Config option to setSkipped in SetStaticValues.java

Otto-Cline:
Allows us to automatically mark a document for to be skipped by our pipeline as a config option in SetStaticValues. We want to do this because SetStaticValues is often used to mark docs to be deleted, so we don't need / want those to be processed by other stages. Originally we would have to use the SkipDocument stage after SetStaticValues using the same conditions for both

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/SetStaticValues.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/SetStaticValues.java
@@ -20,7 +20,7 @@ import java.util.Map;
  * <ul>
  *   <li>staticValues (Map&lt;String, Object&gt;) : A mapping from the field to the value.</li>
  *   <li>updateMode (UpdateMode) : The update mode to use when updating the fields.</li>
- *   <li>skipDoc (boolean) : Sets the document to be skipped by the rest of the pipeline. Can be used in place of the {@link SkipDocument}
+ *   <li>skipDocument (boolean) : Sets the document to be skipped by the rest of the pipeline. Can be used in place of the {@link SkipDocument}
  *   stage when we want to both skip and delete documents, and don't want to add another stage with duplicate conditions.</li>
  * </ul>
  */
@@ -28,25 +28,25 @@ public class SetStaticValues extends Stage {
 
   public static final Spec SPEC = SpecBuilder.stage()
       .optionalString("updateMode")
-      .optionalBoolean("skipDoc")
+      .optionalBoolean("skipDocument")
       .requiredParent("staticValues", new TypeReference<Map<String, Object>>() {}).build();
 
   private final Map<String, Object> staticValues;
   private final UpdateMode updateMode;
-  private final boolean skipDoc;
+  private final boolean skipDocument;
 
   public SetStaticValues(Config config) {
     super(config);
 
     this.staticValues = config.getConfig("staticValues").root().unwrapped();
     this.updateMode = UpdateMode.fromConfig(config);
-    this.skipDoc = ConfigUtils.getOrDefault(config, "skipDoc", false);
+    this.skipDocument = ConfigUtils.getOrDefault(config, "skipDocument", false);
   }
 
   @Override
   public Iterator<Document> processDocument(Document doc) throws StageException {
     staticValues.forEach((fieldName, staticValue) -> doc.update(fieldName, updateMode, staticValue));
-    if (skipDoc) {
+    if (skipDocument) {
       doc.setSkipped(true);
     }
     return null;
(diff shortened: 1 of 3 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/stage/SetStaticValues.java
spencersolomon6, line 22: Make sure to add javadoc whenever we add a new param

lucille-core/src/test/java/com/kmwllc/lucille/stage/SetStaticValuesTest.java
```diff
@@ -36,9 +37,19 @@ public void testSetValues() throws Exception {
...
   }
 
+  @Test
+  public void testSkip() throws Exception {
+    Stage stage = factory.get("SetStaticValuesTest/skipConfig.conf");
+
+    Document doc = Document.create("doc");
+    stage.processDocument(doc);
+
+    assertTrue(doc.getBoolean(".skipped"));
```
spencersolomon6: Lets add an assertion to make sure that the other static values (or at least 1) are properly set before we set the doc to be skipped

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/SetStaticValues.java
```diff
@@ -19,27 +20,32 @@
...
-    updateMode = UpdateMode.fromConfig(config);
+    this.staticValues = config.getConfig("staticValues").root().unwrapped();
+    this.updateMode = UpdateMode.fromConfig(config);
+    this.skip = ConfigUtils.getOrDefault(config, "skip", false);
   }
 
   @Override
   public Iterator<Document> processDocument(Document doc) throws StageException {
     staticValues.forEach((fieldName, staticValue) -> doc.update(fieldName, updateMode, staticValue));
+    doc.setSkipped(skip);
```
rseitz: might be worth wrapping this in a conditional and only calling setSkipped when skip==true

lucille-core/src/test/java/com/kmwllc/lucille/stage/SetStaticValuesTest.java
```diff
@@ -36,9 +37,23 @@ public void testSetValues() throws Exception {
...
+    Stage stage = factory.get("SetStaticValuesTest/skipConfig.conf");
+
+    Document doc = Document.create("doc");
+    stage.processDocument(doc);
+
+    assertEquals("File", doc.getString("myString"));
+    assertEquals(1, doc.getInt("myInt").intValue());
+    assertEquals(true, doc.getBoolean("myBool"));
+
+    assertTrue(doc.getBoolean(".skipped"));
```
rseitz: can use doc.isSkipped()

lucille-core/src/test/java/com/kmwllc/lucille/stage/SetStaticValuesTest.java
rseitz, line 19: add an assertion that this document is NOT skipped

lucille-core/src/main/java/com/kmwllc/lucille/stage/SetStaticValues.java
```diff
@@ -19,27 +20,32 @@
  * <ul>
  *   <li>staticValues (Map&lt;String, Object&gt;) : A mapping from the field to the value.</li>
  *   <li>updateMode (UpdateMode) : The update mode to use when updating the fields.</li>
+ *   <li>skip (Boolean) : Sets the document to be skipped by the rest of the pipeline.</li>
```
rseitz: this new functionality looks like it duplicates SkipDocument so the comment should explain why it is provided here

a user should prefer SkipDocument if they only want to skip the document

a user would want to use SetStaticValues with skip=true if they are specifically setting a deletion marker field and don't want to duplicate the stage conditions that they are using on SetStaticValues on a separate SkipDocument stage

lucille-core/src/main/java/com/kmwllc/lucille/stage/SetStaticValues.java
```diff
@@ -19,27 +20,35 @@
...
 public class SetStaticValues extends Stage {
 
   public static final Spec SPEC = SpecBuilder.stage()
       .optionalString("updateMode")
+      .optionalBoolean("skip")
       .requiredParent("staticValues", new TypeReference<Map<String, Object>>() {}).build();
 
   private final Map<String, Object> staticValues;
   private final UpdateMode updateMode;
+  private final Boolean skip;
```
rseitz: minor but this can be a boolean primitive (not a Boolean object) because we're never setting it to null

I'm also wondering if it would be better to name the parameter "skipDocument" because "skip" is already one of the possible UpdateModes and it's slightly confusing for it to be used in two different ways... this is also minor though & I'll leave the final decision to you

rseitz (approved):
approving but with a comment about using a boolean primitive for the skip field and possibly renaming the parameter to skipDocument

PR #476, opened 2026-05-05
LC-850: Set runs before file marked as expired

Otto-Cline:
A file might be marked as not encountered in our state db during traversal due to a transient error. This PR adds a config option in the file connector to wait a number of runs before marking a file as expired (and thus a tombstone being created / deletion happening) for the sake of safety and consistency.

Each run that a file is not marked as encountered, we increment up a column in the state db. Once that increment reaches the runsBeforeExpiration from our configuration, the file will be marked as expired.


lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -130,6 +131,7 @@ public class FileConnector extends AbstractConnector {
           SpecBuilder.parent("state")
               .optionalString("driver", "connectionString", "jdbcUser", "jdbcPassword", "tableName")
               .optionalBoolean("performDeletions")
+              .optionalNumber("runsBeforeExpiration")
```
spencersolomon6: We should add this to the existing optionalNumber(...)  method call

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -162,6 +179,14 @@ public void shutdown() throws SQLException {
       }
     }
 
+    if (resetStatement != null) {
+      try {
+        resetStatement.close();
+      } catch (SQLException e) {
+        log.warn("Couldn't close update statement (PreparedStatement).", e);
```
spencersolomon6: Update this log to be unique

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -172,9 +197,11 @@ public void shutdown() throws SQLException {
   }
   
   public List<URI> listExpiredFiles() throws SQLException {
-    String selectSQL = "SELECT name FROM \"" + tableName + "\" WHERE encountered = FALSE";
+    String selectSQL = "SELECT name FROM \"" + tableName + "\" WHERE encountered=FALSE AND runs_not_encountered >= ?";
```
spencersolomon6, line 195: Since the definition of an "expired" file is now solely based on the runs_not_encountered value, we should remove the check on encountered here? (unless there is some edge case I am missing). If not, same for the method below

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
```diff
@@ -440,6 +440,99 @@ public void testTraversalMultipleFileTypesWithIncrementalMode() throws Exception
...
+    manager.markFileEncountered(infoFile);
+    manager.incrementRunsNotEncountered();
+    manager.shutdown();
+
+    try (Connection connection = DriverManager.getConnection("jdbc:h2:mem:test", "", "");
+        ResultSet rs = RunScript.execute(connection, new StringReader(secretsQuery))) {
+      assertFalse(rs.next());
+    }
+  }
+
```
spencersolomon6, line 582: Let's add one more test to confirm that ListExpiredFiles respects runsBeforeExpiration being configured to be greater than 1 as well

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
Otto-Cline: Just wanted to be careful in case anything happens- it's better to not delete than to delete if something goes wrong somehow. One thing I can think of is if we set runsBeforeExpiration=0, this prevents everything from immediately being expired. Maybe we want to have some other handling for that instead

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
Otto-Cline: It's set to 2 in testDeletionDeferredByRunsBeforeExpiration, but not super obvious

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
spencersolomon6: RusBeforeExpiration=0 is a good one... I think we should handle that in the constructor and throw an Exception if they try to configure that

spencersolomon6 (changes requested):

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorStateManagerTest.java
spencersolomon6: Where is listExpiredFiles tested in that method?

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -196,6 +197,15 @@ public void execute(Publisher publisher) throws ConnectorException {
       traverseStoragePath(publisher, resource);
     }
 
+    // bump runs_not_encountered so listExpiredFiles and shutdown's deletion see up-to-date counters
+    if (stateManager != null) {
```
rseitz, line 206: since FileConnector is changing here we need some test coverage in this PR for FileConnector itself (current test changes in this PR are only for FileConnectorStageManager). See if there are any existing tests inside FileConnectorTest that can be adjusted to flex the new runsBeforeExpiration feature, or add a new dedicated one.

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
rseitz, line 37: add the new column here

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -172,9 +200,11 @@ public void shutdown() throws SQLException {
   }
   
   public List<URI> listExpiredFiles() throws SQLException {
-    String selectSQL = "SELECT name FROM \"" + tableName + "\" WHERE encountered = FALSE";
+    String selectSQL = "SELECT name FROM \"" + tableName + "\" WHERE runs_not_encountered >= ?";
```
rseitz: would we gain any additional safety by making this:

WHERE encountered = FALSE AND runs_not_encountered >= ?

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -183,10 +213,11 @@ public List<URI> listExpiredFiles() throws SQLException {
   }
 
   private void deleteFilesNotEncounteredAndResetTable() throws SQLException {
-    String deleteSQL = "DELETE FROM \"" + tableName + "\" WHERE encountered=FALSE";
+    String deleteSQL = "DELETE FROM \"" + tableName + "\" WHERE runs_not_encountered >= ?";
```
rseitz: rename method to deleteExpiredFilesAndResetTable()

same comment as above, would we gain any additional safety by making this:

WHERE encountered = FALSE AND runs_not_encountered >= ?

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -125,10 +131,24 @@ public void init() throws ClassNotFoundException, SQLException {
...
+    resetStatement = jdbcConnection.prepareStatement(resetSQL);
+
+    String insertNewFileSQL = "INSERT INTO \"" + tableName + "\" VALUES (?, NULL, TRUE, 0)";
     insertNewFileStatement = jdbcConnection.prepareStatement(insertNewFileSQL);
   }
 
+  /**
+   * Increments the runs_not_encountered counter for every file that was not seen during the current traversal.
+   */
+  public void incrementRunsNotEncountered() throws SQLException {
```
rseitz, line 144: when is it safe to call this method?
what needs to be called before it?
what should be called after it?
add this info to the method javadoc

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
Otto-Cline: me and Spencer discussed this, in support of having both I originally said:
"Just wanted to be careful in case anything happens- it's better to not delete than to delete if something goes wrong somehow. One thing I can think of is if we set runsBeforeExpiration=0, this prevents everything from immediately being expired. Maybe we want to have some other handling for that instead"
Ended up handling that in the constructor with an exception and removing the encountered = FALSE, because it doesn't add any other additional safety.

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
rseitz: here the points that need to be included in the javadoc -- wording is up to you: 

- Must be called exactly once per run
- All markFileEncountered calls must be complete before calling this
- Must be called before shutdown() and before listExpiredFiles(), since both rely on the updated counters
- Calling it before marking is finished will prematurely increment counters for files still being traversed

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -125,17 +132,34 @@ public void init() throws ClassNotFoundException, SQLException {
     String updateSQL = "UPDATE \"" + tableName + "\" SET encountered=true WHERE name=?";
     updateStatement = jdbcConnection.prepareStatement(updateSQL);
 
-    String insertNewFileSQL = "INSERT INTO \"" + tableName + "\" VALUES (?, NULL, TRUE)";
+    String resetSQL = "UPDATE \"" + tableName + "\" SET runs_not_encountered = 0 WHERE name=?";
+    resetStatement = jdbcConnection.prepareStatement(resetSQL);
```
mrwrench24: I know that there were similar comments earlier about consolidating encountered / runs not encountered in a `WHERE` statement earlier, but is _this_ a case where we can consolidate these two? We could have just an `updateStatement` that sets `encountered=true` and `runs_not_encountered=0`?

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -62,6 +62,7 @@
  *   <li>state.jdbcPassword (String, Optional) : Database password. Defaults to "".</li>
  *   <li>state.tableName (String, Optional) : Table name for state. Defaults to the connector name.</li>
  *   <li>state.performDeletions (Boolean, Optional) : Delete rows for files removed from storage. Defaults to true.</li>
+ *   <li>state.runsBeforeExpiration (Int, Optional) : The number of runs a file is not encountered consecutively before it is marked as expired. Defaults to 1.</li>
```
mrwrench24: I would recommend rephrasing to something like:

> After a file is not encountered for this number of runs, it will be marked as expired. Must be at least 1. Defaults to 1.

We want to mention the valid values but also I think it is good to help specify for users the "one-off" case - does 0 mean delete on unencountered, or 1, etc...

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -125,17 +132,34 @@ public void init() throws ClassNotFoundException, SQLException {
...
+    String insertNewFileSQL = "INSERT INTO \"" + tableName + "\" VALUES (?, NULL, TRUE, 0)";
     insertNewFileStatement = jdbcConnection.prepareStatement(insertNewFileSQL);
   }
 
+  /**
+   * Increments the runs_not_encountered counter for every file that was not seen during the current traversal. This method must
+   * be called exactly once per run, after all markFileEncountered calls are complete.
+   * Must be called before shutdown() and before listExpiredFiles(), since both rely on the updated counters.
+   * Calling it before marking is finished will prematurely increment counters for files still being traversed.
+   */
```
mrwrench24, line 143: This isn't necessarily something to address in this PR / immediately, but I will just point out we want to be _very_ mindful about the changes in [LC-815](https://github.com/kmwtechnology/lucille/pull/486) as well.

Each traversal happens on its own thread, so the "intersection" of these PRs is _hopefully_ straightforward, but again just something to be aware of as the FileConnector grows in complexity :)

mrwrench24 (approved):
LGTM with just some small things to look at :)

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
Otto-Cline: this is a good point; initially I believe I added the extra statement to separate the handling of runs_not_encountered from encountered in case of edge cases, but it seems that that is not necessary looking back

meghanboyd:
@rseitz do you want to look at this PR one more time before its merged?

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorTest.java
```diff
@@ -740,4 +740,106 @@ public void testTombstonesWithFileSystemChanges() throws Exception {
...
+      Document tombstone1 = thirdRunDocs.stream()
+          .filter(doc -> Boolean.TRUE.equals(doc.getBoolean(FileConnector.EXPIRED)))
+          .findFirst()
+          .orElseThrow();
+      assertTrue(tombstone1.getString(FileConnector.FILE_PATH).contains("file1.txt"));
+      assertTrue(tombstone1.isSkipped());
+
+      // Fourth run
+      TestMessenger messenger4 = new TestMessenger();
+      Publisher publisher4 = new PublisherImpl(config, messenger4, "run3", "pipeline1");
```
rseitz: should be "run4"?

rseitz (approved):

PR #478, opened 2026-05-06
LC-865: Consume HOCON when uploading config to API and add general documentation

Otto-Cline:
Updates the API so that it takes in HOCON files rather than JSON for the Lucille config. Also provides documentation on how to use HTTPS with the API, and adds up to date info to the README on endpoints and configuring HTTPS


lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LucilleResource.java
```diff
@@ -70,28 +70,28 @@ public LucilleResource(RunnerManager runnerManager, AuthHandler authHandler) {
...
   @POST
   @Tag(name = "Config")
   @Path("/config")
-  @Consumes(MediaType.APPLICATION_JSON)
+  @Consumes({MediaType.APPLICATION_JSON, "application/hocon"})
   @Operation(summary = "Create a new Lucille config to be run later",
       description = "Creates a new Lucille config, which can be used later by its referenced uuid.")
   public Response createConfig(@Parameter(hidden = true) @Auth Optional<PrincipalImpl> user,
       @RequestBody(description = "Run configuration as a key-value map", required = true,
           content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(
```
spencersolomon6: Let's make sure we update the request body here with the same changes. The description and mediaType specifically.

lucille-plugins/lucille-api/src/test/java/endpoints/LucilleResourceTest.java
```diff
@@ -80,8 +76,7 @@ public void testStartRun_InvalidConfigId() {
   @Test
```
spencersolomon6, line 76: Let's add a test to make sure that valid JSON is still parsed properly when input as a single string

Otto-Cline: Do you mean having the configBody be a full config in json and parsing it?

spencersolomon6: It doesn't have to be a full config, it can be a snippet like the existing tests but I see that all of the existing tests were changed to HOCON format (key = /"value/") and just want to see one in JSON (\"key\": \"value\")

spencersolomon6 (approved):

rseitz:
PR title and ticket says this is about consuming HOCON but I see https-related changes and doc. Please update the PR description to clarify the goal of this PR.

lucille-plugins/lucille-api/conf/https_api.yml
```diff
@@ -0,0 +1,33 @@
+# This is the DropWizard configuration file. We are configuring HTTPS and basicAuth.
+# Refer to lucille-api/README.md for information on using HTTPS. More configuration options can be found at:
+# https://www.dropwizard.io/en/stable/manual/configuration.html#man-configuration
+server:
+  type: default
+  applicationConnectors:
+        - type: https
```
rseitz: indentation issue

rseitz (approved):

PR #479, opened 2026-05-07
LC-848: Logging for indexers where deletionMarkerField isn't supported

Otto-Cline:
A user can specify deletionMarkerField in their indexer config and it will not trigger a validation error.
However, the only Indexer implementations that respect this field are SolrIndexer, OpenSearchIndexer, and PineconeIndexer.
If the user is using ElasticsearchIndexer or others, the setting will have no effect. The user will never receive feedback that it isn’t being considered.

This PR adds logging for those indexers where the setting has no effect

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/CSVIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/CSVIndexer.java
@@ -56,7 +56,7 @@ public class CSVIndexer extends Indexer {
       throw new IllegalArgumentException(
           "Cannot create CSVIndexer. Config setting 'indexer.indexOverrideField' is not supported by CSVIndexer.");
     }
-    if (this.deletionMarkerField != null || this.deleteByFieldField != null || this.deleteByFieldValue != null) {
+    if (this.deletionMarkerField != null || this.deleteByFieldField != null) {
       log.warn("Deletion is not supported for this indexer. Documents marked for deletion will be written as regular rows.");
     }
     
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/CSVIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/CSVIndexer.java
@@ -56,8 +56,8 @@ public class CSVIndexer extends Indexer {
       throw new IllegalArgumentException(
           "Cannot create CSVIndexer. Config setting 'indexer.indexOverrideField' is not supported by CSVIndexer.");
     }
-    if (this.deletionMarkerField != null) {
-      log.warn("indexer.deletionMarkerField is set but is not supported by CSVIndexer; documents marked for deletion will be written as regular rows.");
+    if (this.deletionMarkerField != null || this.deleteByFieldField != null || this.deleteByFieldValue != null) {
+      log.warn("Deletion is not supported for this indexer. Documents marked for deletion will be written as regular rows.");
     }
     
     this.writer = writer;
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/CSVIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/CSVIndexer.java
@@ -56,6 +56,10 @@ public class CSVIndexer extends Indexer {
       throw new IllegalArgumentException(
           "Cannot create CSVIndexer. Config setting 'indexer.indexOverrideField' is not supported by CSVIndexer.");
     }
+    if (this.deletionMarkerField != null) {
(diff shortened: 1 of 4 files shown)


spencersolomon6:
I've spent some time looking over how this is implemented now and I think the best way to resolve this would be to make some further reaching changes to the Indexer, and the classes which actually support this functionality. Couple motivators for this

1. In general we should not have configuration in the shared Indexer Class which is not implemented by all Indexers
2. If this configuration was only made available where it is used, our SPEC validation would handle this for us, instead of needing one off logging for this case.

As such, I propose we move the deletionMarkerField and deletionMarkerValue out of the Indexer class and into the classes which implement it (SolrIndexer, OpenSearchIndexer and PineconeIndexer). This will result in some duplicated code, but we already have precedent for that with fields like url and acceptInvalidCerts.

One thing to note with this is this would be a breaking configuration change, so we should consider this extra carefully and make sure to document it clearly in release notes if we move forward. @rseitz any thoughts are welcome

Otto-Cline:
Is there any reason we don't have this functionality in the other indexer impls in the first place? Might be easier just to add it to them and then we don't have a breaking change @spencersolomon6

spencersolomon6:
@Otto-Cline  Met with Rudi and discussed this and we decided the best course of action is to merge this PR as is and add follow up tickets to implement deletion for the ElasticSearch and Weaviate Indexers. This is motivated by a couple factors:

1. Desire to keep the deletionMarkerField config within the indexer block
2. Concept of deletes don't really make sense for the CSV or Nop Indexers but these are edge cases

As such, we will keep this functionality in the shared Indexer class and log when its not supported, as you have already done.

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/indexer/CSVIndexer.java
```diff
@@ -56,6 +56,10 @@ public CSVIndexer(Config config, IndexerMessenger messenger, ICSVWriter writer,
       throw new IllegalArgumentException(
           "Cannot create CSVIndexer. Config setting 'indexer.indexOverrideField' is not supported by CSVIndexer.");
     }
+    if (this.deletionMarkerField != null) {
```
spencersolomon6: We should check deleteByFieldField here as well. Can make the logging more general to say deletion is not supported

spencersolomon6 (changes requested):

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
```diff
@@ -79,6 +79,10 @@ public ElasticsearchIndexer(Config config, IndexerMessenger messenger, boolean b
       throw new IllegalArgumentException(
           "Cannot create ElasticsearchIndexer. Config setting 'indexer.indexOverrideField' is not supported by ElasticsearchIndexer.");
     }
+    if (this.deletionMarkerField != null || this.deleteByFieldField != null) {
+      log.warn("Deletion is not supported for this indexer. Documents marked for deletion will be written as regular rows.");
```
spencersolomon6: I think the old language like `indexed as regular docs` was more descriptive for Elstaic/Weaviate Indexers

lucille-core/src/main/java/com/kmwllc/lucille/indexer/NopIndexer.java
```diff
@@ -18,8 +20,14 @@ public class NopIndexer extends Indexer {
...
   public static final Spec SPEC = SpecBuilder.indexer().build();
 
+  private static final Logger log = LoggerFactory.getLogger(NopIndexer.class);
+
   public NopIndexer(Config config, IndexerMessenger messenger, boolean bypass, String metricsPrefix, String localRunId) {
     super(config, messenger, bypass, metricsPrefix, localRunId);
+
+    if (this.deletionMarkerField != null || this.deleteByFieldField != null) {
+      log.warn("Deletion is not supported for this indexer. Documents marked for deletion will be written as regular rows.");
+    }
```
spencersolomon6: Thinking about this a bit more... I think we can remove this. The NopIndexer handles deletions the same way it handles any other index operation; by doing nothing. So I think its fine to say it does support deletions and you shouldn't expect an error when using it

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/indexer/NopIndexer.java
rseitz: Agree -- yes! It has very efficient and consistent support for deletes! It handles them faster than any other indexer. :)

rseitz (approved):

PR #480, opened 2026-05-07
LC-855: Add integration tests for config and run in api

Otto-Cline:
We were missing tests for

v1/config-info/connector-list
v1/config-info/stage-list
v1/config-info/indexer-list
v1/config POST
v1/config/{id}
v1/run POST
v1/run GET
v1/run/{id}


Otto-Cline:
I took a substring of the responses to get the actual IDs which is hacky but didn't see another way around it; lmk if there's a better solution

Otto-Cline:
The dropwizard test client defaults to a 5s timeout, and having all the config-info/* endpoints in one test was timing it out in the CI because those endpoints use ClassGraph().enableAllInfo().scan(), which is expensive. Maybe something to look at but not super solvable unless we decided to maybe move the scan to happen during the run and have it cached

spencersolomon6 (approved):

spencersolomon6:
Let's wait to merge this until we complete LC-868. Hopefully we can remove the timeout increases after that work.

Otto-Cline:
PR closed, incorporated into [LC-868](https://github.com/kmwtechnology/lucille/pull/481)

PR #485, opened 2026-05-18
Quick draft implementation for base lucille Dockerfile.

kyle-kmw:


mrwrench24:
Ran:

```
jakesquatrito@KMW-MacBook-Pro lucille % docker run \                                 
  -v "$(pwd)/lucille-examples/lucille-rss-example/conf:/lucille/conf" \
  -e LUCILLE_CONF=/lucille/conf/single.conf \
  lucille
```

Did require me to add the following line to the `Dockerfile`:
`COPY lucille-core/target/*.jar /lucille/lib/`
Since the actual Lucille jars weren't being copied over (But the dependencies were!)

PR #486, opened 2026-05-18
LC-815: Multithreading for FileConnector traversals

Otto-Cline:
Adds a multithreading option in the config for the FileConnector. This will allow resources to be traversed in parallel.

Some notes:
- I extend the Thread class in this first pass implementation, though I believe it is more common in the present day to use the java.util.concurrent package, which gives you an ExecutorService. We do this in WorkerPool.java, though we also use Thread in several places. My current implementation seems to work fine, so maybe this is overkill
- The Thread class extension, FileConnectorThread, interacts with the FileConnector slightly weirdly (at least in my eyes), because the overridden run() method can't directly bubble up a ConnectorException. The workaround is to catch the exception and then give it to the FileConnector using a getter.
- We might want to manually test this on a large amount of paths with a good amount of data
- I synchronized the StateManager methods that get touched indirectly by what we are multithreading (traverseStoragePath() drills down to processAndPublishFileIfValid(), which calls a few methods in the StateManager).


Otto-Cline:
Currently working on
- make "multithreaded" a parent
- has options "enabled" and "numberOfThreads"
- both optional, no numberOfThreads defaults to 1 and enabled defaults to true
- threads pull from a queue of paths

questions:
- if 1 thread is very similar to our old implementation without threads, can we just always enable multithreading?

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnectorStateManager.java
```diff
@@ -39,12 +39,6 @@
  *
  * <p> Lucille includes H2 as a dependency. You are welcome to configure the FileConnectorStateManager to use an embedded
  * H2 instance - use the <code>org.h2.Driver</code>, and something like <code>jdbc:h2:{LOCAL_FILE_PATH}</code> as your connectionString.
- *
- * <p> <b>Note:</b> This class is operating under two key assumptions about FileConnector / Connectors:
- * <ol>
- *   <li>Connectors run sequentially.</li>
```
mrwrench24, line 45: I would leave up the first bullet point - I think this is still a valid assumption the class is working off of

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -281,7 +360,7 @@ public void close() {
     }
   }
 
-  private void traverseStoragePath(Publisher publisher, URI pathToTraverse) throws ConnectorException {
+  void traverseStoragePath(Publisher publisher, URI pathToTraverse) throws ConnectorException {
```
mrwrench24: bring back to be private?

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -181,23 +201,37 @@ public FileConnector(Config config) throws ConnectorException {
       throw new IllegalArgumentException("publishMode must be set and be incremental to use the sendTombstones toggle.");
     }
 
-    // Cannot specify multiple storage paths and a moveTo of some kind
+    // Cannot specify multiple storage paths and a moveTo of some kind. Also a thread safety check:
+    // LocalStorageClient.moveFile has a Files.exists/createDirectory time of check time of use race, and restricting moveTo to
```
mrwrench24: a very small thing, but I think this comment applies to all of the storage clients - users are prevented from having some nasty side effects because we (implicitly) prevent `moveTo` calls in a multithreaded traversal

I may have misunderstood the intent of the original comment though

lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorTest.java
```diff
@@ -740,4 +739,116 @@ public void testTombstonesWithFileSystemChanges() throws Exception {
...
+    Publisher publisher = new PublisherImpl(config, messenger, "run", "pipeline1");
+    Connector conn = new FileConnector(config);
+    conn.execute(publisher);
+    conn.close();
+    assertEquals(21, messenger.getDocsSentForProcessing().size());
+
+    // Second run: every row already exists in the state DB, so every update goes through the updateStatement path rather than the
+    // insert path, different code paths under the lock.
+    messenger = new TestMessenger();
+    publisher = new PublisherImpl(config, messenger, "run", "pipeline1");
```
mrwrench24, line 788: maybe a small thing but it might be worth running in incremental mode to further flex that everything in the database was updated appropriately / we correctly don't publish anything

mrwrench24 (commented):
just a few small comments and suggestions

Otto-Cline:
Illustration of code components touched by threads (in black. Checks indicate component is thread-safe given the current implementation)
IMG_2305

rseitz:
superseded by https://github.com/kmwtechnology/lucille/pull/561

PR #487, opened 2026-05-19
LC-866: update reserved fields names

kiratraynor:
The reserved fields names used to start with a period. This could cause some confusion when dealing with nested json. We are replacing the names to begin with three underscores. This is a breaking change.

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyJSONataTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyJSONataTest.java
@@ -230,4 +230,22 @@ public class ApplyJSONataTest {
     assertEquals("doc1", doc.getId());
     assertNull(doc.getJson("destination"));
   }
+
+  @Test
+  public void testApplyReservedFields() throws StageException {
+    Stage stage = factory.get("ApplyJSONataTest/reservedFields.conf");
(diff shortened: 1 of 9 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyJSONataTest.java
```diff
@@ -230,4 +230,22 @@ public void testExceptionDuringTransformation() throws StageException {
     assertEquals("doc1", doc.getId());
     assertNull(doc.getJson("destination"));
   }
+
+  @Test
+  public void testApplyReservedFields() throws StageException {
```
rseitz, line 235: would help to have a brief comment explaining the regression we're avoiding here... this test would break if reserved fields like the attached children field began with a "." which used to be the case

rseitz (approved):
approving with small comment about documenting the reason behind testApplyReservedFields

lucille-core/src/test/java/com/kmwllc/lucille/stage/ApplyJSONataTest.java
kiratraynor: Will do. That reminds me though, I wanted to add a test for the CopyFields stage as well with reserved fields and isNested set to true. Updating the PR now.

spencersolomon6 (approved):

PR #489, opened 2026-05-21
LC-834: remove unnecessary helper in QueryDatabaseTest

Otto-Cline:


lucille-core/src/test/java/com/kmwllc/lucille/stage/QueryDatabaseTest.java
```diff
@@ -326,8 +326,7 @@ public void testUnsupportedTypeTime() throws Exception {
     Document d = Document.create("id");
     d.setField("get_row", 1);
     // JDBCUtils should log warning
-    assertEquals("no Error found", getUnsupportedMsg(stage, d));;
-
+    stage.processDocument(d);
```
rseitz, line 330: the change here is good
but while we're modifying this test we should take the chance to improve what it's asserting

the current test shows that no exception is thrown when we get back an unsupported type from the sql query

should at least also verify that no new field was added to the document

even better: what if the query returns two columns where one has an unsupported type and the other has a supported type. verify that the field for the unsupported type is not set, and the field for the supported type is set. this would require changing time_type.conf to return two colums

rseitz (approved):

spencersolomon6 (approved):

PR #490, opened 2026-05-21
LC-839: add fieldNamesField config option

Otto-Cline:
We often want to know what properties are available to be on a document as a result of text extraction. This PR implements a "fieldNamesField" config option on the TextExtractor stage, which lets the user specify a new multi-valued field to be added to the document which holds all of text extraction properties as a string list. This way, we can more easily query the index after an ingest run to understand what properties are available and should be maintained.


spencersolomon6 (approved):

lucille-plugins/lucille-tika/src/test/java/com/kmwllc/lucille/tika/stage/TextExtractorTest.java
```diff
@@ -343,6 +343,42 @@ public void testExtractionWithURI() throws Exception {
...
+    List<String> fieldNames = doc.getStringList("tika_property_names");
+    assertTrue(fieldNames.contains("tika_content_type"));
+    assertTrue(fieldNames.contains("tika_content_encoding"));
+    assertTrue(fieldNames.contains("tika_x_tika_parsed_by"));
+
+    // make sure every field appears exactly once and that we got everything
+    Set<String> metadataFieldsOnDoc = doc.getFieldNames().stream()
+        .filter(name -> name.startsWith("tika_"))
+        .filter(name -> !name.equals("tika_property_names"))
+        .collect(java.util.stream.Collectors.toSet());
```
rseitz: import java.util.stream.Collectors

Otto-Cline: Have a rule against this in CLAUDE.md and it got past my own review oh man

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -48,14 +48,16 @@
  * parseTimeout (Long, Optional) : timeout for parsing in milliseconds
  * whitelist (StringList, Optional) : list of metadata names that are to be included in document
  * blacklist (StringList, Optional) : list of metadata names that are not to be included in document
+ * fieldNamesField (String, Optional) : if set, the stage writes the (cleaned, tika_ prefixed) name of every metadata field added
```
rseitz: suggest:

if set, each extracted metadata field's prefixed name is added as a separate value to this multi-valued field

rseitz (approved):

PR #496, opened 2026-05-25
LC-878: log json of failed documents via dedicated logger

rseitz:


Otto-Cline (approved):

mrwrench24:
Everything looks good to me here...

To summarize from yesterday, for my own sanity / understanding: a lot of the discussion around this PR largely stemmed from the fact that this feature would be not so intuitive in Distributed mode. A "failed" documents file could show up in the separate processes. Users would have to do something on their own to reconcile that. Our discussion was largely about - in a future PR - how we could make something like this more intuitive via the failure topic. 

All of this to say - this PR is not too controversial on its own (and is nevertheless, additive!) - our discussion was largely about _future_ work.

mrwrench24 (approved):
LGTM as long as my summary above is accurate

kyle-kmw (approved):

rseitz:
> LGTM as long as my summary above is accurate

Yes, it's correct, and thanks for sharing it! :)

PR #498, opened 2026-05-26
LC-880: Change site to include version option for pre-release docs

Otto-Cline:
Adds a dropdown option for accessing the doc site as it exists in main, also allows devs to run versionsworkflow.sh script and see their changes.


.github/workflows/hugo.yml
```diff
@@ -4,7 +4,7 @@ name: Deploy Hugo site to Pages
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ["main"]
+    branches: ["LC-880-docs-version-for-main"]
```
spencersolomon6: Let's set this back to main, I don't think we want things to go live on the site as we test on this branch

Otto-Cline: Deployment gets blocked at branch level

spencersolomon6: Can we switch this back now that we have checked the build works?

spencersolomon6 (approved):

rseitz (approved):

PR #500, opened 2026-05-28
LC-881: doc site preview folder

Otto-Cline:
Alters our versionsworkflow.sh script so that a copy of the doc/site folder is made when running in local mode. This lets the developer make changes to the doc site and rerun the script to see them, without the generation interfering with the ability to easily commit their changes. Also significantly simplifies the process of reverting changes made by the versioned site generation (only need to delete the site-preview folder now). Added site-preview to .gitignore for extra convenience.


.github/scripts/versionsworkflow.sh
```diff
@@ -1,33 +1,45 @@
-# This script updates the working tree with all the content needed for "hugo build" to generate versioned docs. To do this:
+# This script prepares a versioned doc site in doc/site-preview, illustrating what the site will look like when deployed.
```
spencersolomon6: Lets add some information to the doc/site README about what this script is and how to use it

spencersolomon6: This top level comment isn't totally accurate. This script only prepares the site in doc/site-preview in local mode. Let's update the top comment to say that the script runs both locally and in the actions and generates a versioned doc site. Then in the section below about local mode we can specify where that happens

Otto-Cline:
reminder to point at main once 880 merged

.github/workflows/hugo.yml
```diff
@@ -4,7 +4,7 @@ name: Deploy Hugo site to Pages
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ["main"]
+    branches: ["LC-880-docs-version-for-main"]
```
rseitz: weren't we going to change this back to main?

rseitz (approved):

PR #501, opened 2026-05-28
LC-851: doc site search returns correct versions

Otto-Cline:
We would like the search feature on the doc site to get back pages from the current version of the site that is selected. This required
- Removing exclude-search: true from the _index.md of each version
- Override offline-search-index.json so that it only emits pages where the .Section matches that of the version you are in
- Override search-input.html so that it builds one index JSON for each version of the docs.

### DIFF
--- a/.github/scripts/versionsworkflow.sh
+++ b/.github/scripts/versionsworkflow.sh
@@ -78,7 +78,6 @@ linkTitle: Docs-pre-release
 weight: 20
 cascade:
   type: docs
-  exclude_search: true
 ---
 EOF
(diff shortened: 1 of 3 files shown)


rseitz:
just noting that the indentation in the initial commit of search-input.html seems to be different from what's present in the source repo here: https://github.com/google/docsy/blob/v0.14.0/layouts/_partials/search-input.html

not necessary to fix that for this PR, but mentioning it here to be aware of when maintaining this file

rseitz (approved):

spencersolomon6 (approved):

PR #502, opened 2026-05-29
LC-864: add config option to apply RenameFields stage to child docs

Otto-Cline:
Implementation decision: made processDocument on this stage recursive to apply field renaming to child docs of child docs

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/stage/RenameFields.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/stage/RenameFields.java
@@ -24,8 +24,8 @@ import java.util.Map.Entry;
  *   <li>fieldMapping (Map&lt;String, String&gt;) : A 1-1 mapping of original field names to new field names.</li>
  *   <li>updateMode (String, Optional) : Determines how writing will be handling if the destination field is already populated. Can be
  *   'overwrite', 'append' or 'skip'. Defaults to 'overwrite'.</li>
- *   <li>applyToChildren (Boolean, Optional) : Determines whether renaming of fields will be applied to children, grandchildren, etc.
- *   Defaults to false.</li>
+ *   <li>applyToChildren (Boolean, Optional) : Determines whether renaming of fields will be applied to attached child docs (the stage
+ *   is automatically applied to emitted children). Applies to all attached children, irrespective of stage conditions.</li>
  * </ul>
  */
 public class RenameFields extends Stage {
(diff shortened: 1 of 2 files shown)


spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/RenameFields.java
```diff
@@ -63,6 +70,21 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
     }
 
+    if (applyToChildren) {
+      List<Document> children = doc.getChildren();
+      if (!children.isEmpty()) {
+        for (Document childDoc : children) {
+          processDocument(childDoc);
+        }
+
+        doc.removeChildren();
```
mrwrench24, line 82: the only thing I am curious about is why we have to remove the children and then add them all back. I can see that removing these last two operations means the child document is not actually _updated_ within the parent. I am mostly just curious as to why that is.

mrwrench24: NVM I understand now (of course ~3 min after I leave a comment wondering about this.) 
`addChild` is more static than I imagined it being  

I can see part of our discussion on this is about standardizing these operations, maybe just something to consider how desirable this is - future stage devs would have to know / follow a similar pattern repeatedly when operating on child docs

lucille-core/src/main/java/com/kmwllc/lucille/stage/RenameFields.java
```diff
@@ -22,22 +24,27 @@
  *   <li>fieldMapping (Map&lt;String, String&gt;) : A 1-1 mapping of original field names to new field names.</li>
  *   <li>updateMode (String, Optional) : Determines how writing will be handling if the destination field is already populated. Can be
  *   'overwrite', 'append' or 'skip'. Defaults to 'overwrite'.</li>
+ *   <li>applyToChildren (Boolean, Optional) : Determines whether renaming of fields will be applied to attached child docs (the stage
```
rseitz: `Determines whether renaming of fields will be applied to attached children. If this stage has conditions, they will control the stage's own execution as normal, but will not be re-evaluated per child: when this stage runs, renaming is applied to all attached children. This stage has no special handling for emitted children; it sees them as first-class documents flowing through the pipeline.`

rseitz (approved):
approving with a request to update comment before merging

PR #504, opened 2026-06-01
LC-882: Action for Core + API Docker Image Releases

mrwrench24:
I am not sure what triggers GitHub Actions to eventually show up a "Run Workflow" button for "release to docker", which is currently implemented as requiring a manual run so we can create updated versions at will. So, I used this command for now - perhaps it will show up after we merge this PR hopefully?

`gh workflow run release-to-docker.yml --ref LC-882-core-and-api-docker-images -f release_version=TEST -f docker_revision=0`

**Note** that this PR removes some stale / potentially incorrect information and functionality regarding the use of a Lucille Config when launching the API.


Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM eclipse-temurin:17-jre
```
kyle-kmw: use java 21?

.github/workflows/release-to-docker.yml
```diff
@@ -0,0 +1,59 @@
...
+        uses: actions/setup-java@v4
+        with:
+          java-version: |
+            21
+            17
+          distribution: temurin
+          cache: maven
+
+      - name: Build with Maven
+        run: mvn -B package --file pom.xml
```
kyle-kmw: I wonder if we should pull down a released version of lucille from maven rather than building it here

mrwrench24: We may be able to, but I am not quite sure what we would do about Lucille's (lucille-core) dependencies.

if I am understanding this correctly, we _could_ get away with this for the API - everything is in one jar - but then we have two different approaches which could be a bit confusing.

this isn't my strong suit so let me know if I am off base

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ && useradd -r -g lucille -d /lucille -s /usr/sbin/nologin lucille \
+ && chown -R lucille:lucille /lucille
+
+# Environment variables with sensible defaults
+# LUCILLE_CONF: path to the Lucille config file (required)
+# JAVA_OPTS: JVM flags such as heap size and GC settings
+# LUCILLE_OPTS: Lucille Runner CLI flags (e.g. -local, -usekafka)
+ENV LUCILLE_CONF=""
+ENV LUCILLE_OPTS=""
+ENV JAVA_OPTS="-Xms256m -Xmx1g"
```
rseitz: when Xms and Xmx are set to different values you can run into delays while the heap is being resized (and if Xmx is too high for the system you may not find out until the resize actually happens). So a best practice is to set them the same

kyle-kmw (approved):

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM eclipse-temurin:21-jre
+
+LABEL maintainer="kmwtechnology"
+LABEL description="Lucille Search ETL base image"
+
+# Create a standard working directory
+WORKDIR /lucille
+
+# Copy the built Lucille artifacts
+COPY lucille-core/target/*.jar /lucille/lib/
```
rseitz: if we've build via "mvn clean install" there will be some jars in target that we probably don't need in the docker image:

```
lucille-core-<version>-SNAPSHOT-javadoc.jar
lucille-core-<version>-SNAPSHOT-sources.jar
lucille-tests.jar
```

lucille-plugins/lucille-api/docker-entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+exec java \
+  ${JAVA_OPTS} \
+  -Dconfig.file="${LUCILLE_CONF}" \
```
rseitz: should we check that the conf variable is set via `if [ -z "$LUCILLE_CONF" ];` like in the other entrypoint?

.github/workflows/release-to-docker.yml
```diff
@@ -0,0 +1,79 @@
...
+        type: string
+
+env:
+  release: ${{ inputs.release_version }}.${{ inputs.docker_revision }}
+
+jobs:
+  publish:
+    # run when manually triggered or triggered by *successful* sonatype
+    if: |
+      github.event_name == 'workflow_dispatch' ||
```
rseitz: I'm not entirely sure but "workflow_call" might be another github event name that could come up that we'd want to accept here

mrwrench24: good catch. I think we can remove because this is only called manually or by an _already successful_ release-to-sonatype.

Dockerfile
mrwrench24: updated to copy over only `Lucille.jar`

lucille-plugins/lucille-api/docker-entrypoint.sh
mrwrench24: this has been updated to require the Dropwizard config, which IS used. as for the Lucille conf, I think I will make a follow-up ticket to scrub mentions of this seemingly unused Lucille config file from the API documentation?

mrwrench24: never mind... just did it now

.github/workflows/release-to-docker.yml
```diff
@@ -0,0 +1,74 @@
...
+      - run: echo "Release version ${{ env.release }}!"
+
+      - uses: actions/checkout@v4
+
+      - name: Set up Java
+        uses: actions/setup-java@v4
+        with:
+          java-version: |
+            21
+            17
```
rseitz: remove 17 per recent changes

doc/docker.md
```diff
@@ -0,0 +1,120 @@
...
+```dockerfile
+FROM lucille:latest
+
+# Copy in your config and any additional resources
+COPY my-config.conf /lucille/conf/my-config.conf
+COPY data/ /lucille/data/
+
+# Set the config file path and JVM options
+ENV LUCILLE_CONF=/lucille/conf/my-config.conf
+ENV JAVA_OPTS="-Xms512m -Xmx4g"
```
rseitz: even though this is only doc, we don't want to set Xms and Xmx to different values

doc/docker.md
```diff
@@ -0,0 +1,120 @@
...
+## Configuring JVM Arguments
+
+Control JVM behavior through the `JAVA_OPTS` environment variable:
+
+```bash
+# Set heap size
+docker run --env JAVA_OPTS="-Xms1g -Xmx4g" ...
+
+# Use container-aware memory settings (adapts to container memory limits)
+docker run --env JAVA_OPTS="-XX:+UseContainerSupport -XX:MaxRAMPercentage=75.0" ...
```
rseitz: let's make sure any JAVA_OPTS we're proposing are actually needed. I'm seeing "UseContainerSupport has been on by default since Java 8u191/8u192 and Java 10. By Java 21 on any mainstream JVM (OpenJDK, Eclipse Temurin, Amazon Corretto, Microsoft, Azul) it is always enabled by default and explicitly passing it is a no-op."

doc/docker.md
```diff
@@ -0,0 +1,120 @@
...
+
+# Enable G1 GC tuning
+docker run --env JAVA_OPTS="-Xms1g -Xmx4g -XX:+UseG1GC -XX:MaxGCPauseMillis=200" ...
```
rseitz: again, make Xms == Xmx
make sure we understand what the other GC settings are doing and confirm that they are worth illustrating here; otherwise remove

lucille-examples/lucille-distributed-example/README.md
```diff
@@ -13,4 +13,11 @@
 
 Note: running ```docker-compose up --abort-on-container-exit```would run the internal test in runner, but it will not verify the test results.
 
-**Local runs:** After each run, remove containers to avoid a stale state, otherwise the test will error.
\ No newline at end of file
+**Local runs:** After each run, remove containers to avoid a stale state, otherwise the test will error.
+
+### A Note on Dockerfiles
+As you may have noted, there is a Dockerfile in the root of the project which is not
```
rseitz, line 19: but this means that we don't really have any test coverage for the "official" Dockerfile or the images that we publish to github, right? how can we maintain confidence that our published images are in good shape? OK to do this in a separate ticket/PR, but let's think how to acheive it.

doc/docker.md
mrwrench24: the incoming updated docs should not have any redundant jvm options...

lucille-examples/lucille-distributed-example/README.md
mrwrench24: correct... I agree we do want more testing for this (maybe as part of release-to-docker)

kiratraynor (approved):

doc/docker.md
```diff
@@ -0,0 +1,117 @@
...
+## Configuring JVM Arguments
+
+Control JVM behavior through the `JAVA_OPTS` environment variable:
+
+```bash
+# Example: Adjust heap size
+docker run --env JAVA_OPTS="-Xms1g -Xmx1g" ...
+
+# Example: Garbage collection tuning
+docker run --env JAVA_OPTS="-Xms1g -Xmx1g -XX:+UseG1GC -XX:MaxGCPauseMillis=200" ...
```
rseitz: I believe -XX:+UseG1GC is the default Java 21, so this is a no-op
When illustrating how you would tune GC, it would be ideal to show a setting that actually changes something
or just don't provide this example at all because if someone is working in this file they should know how decide what java options they want to use

doc/docker.md
```diff
@@ -0,0 +1,117 @@
...
+
+```dockerfile
+FROM lucille:0.5.0
+
+COPY conf/production.conf /lucille/conf/production.conf
+COPY conf/application.conf /lucille/conf/application.conf
+COPY certs/ /lucille/certs/
+
+ENV LUCILLE_CONF=/lucille/conf/production.conf
+ENV JAVA_OPTS="-XX:+UseContainerSupport -XX:MaxRAMPercentage=75.0 -XX:+UseG1GC"
```
rseitz: again, -XX:+UseContainerSupport and -XX:+UseG1GC are defaults...
and not sure why we're doing -XX:MaxRAMPercentage=75.0 here while we show -Xms1g -Xmx1g elsewhere in this file

rseitz (approved):

PR #505, opened 2026-06-02
LC-845: Retries for solrindexer

Otto-Cline:
Implements retries for the Solr indexer which is similar to OpenSearch retry implementation

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
@@ -256,7 +256,7 @@ public class SolrIndexer extends Indexer {
         solrClient.add(collection, solrDocs);
       }
     } catch (SolrException e) {
-      // Decide whether to throw based on http status code
+      // Decide error message based on http status code from SolrException
       throw toRetryableException(e);
     } catch (SolrServerException | IOException e) {
       // Communication / transport failure with no http status code available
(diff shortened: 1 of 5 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -243,58 +245,81 @@ protected Set<Pair<Document, Exception>> sendToIndex(List<Document> documents) t
...
-    } else {
-      solrClient.add(collection, solrDocs);
+    try {
+      if (collection == null) {
+        solrClient.add(solrDocs);
+      } else {
+        solrClient.add(collection, solrDocs);
+      }
+    } catch (SolrException e) {
+      // Decide whether to throw based on http status code
```
mrwrench24: small thing - this comment makes it sound like the method may or may not throw based on the status code. may be worth updating

lucille-core/src/main/java/com/kmwllc/lucille/indexer/SolrIndexer.java
```diff
@@ -243,58 +245,81 @@ protected Set<Pair<Document, Exception>> sendToIndex(List<Document> documents) t
...
-      solrClient.add(solrDocs);
-    } else {
-      solrClient.add(collection, solrDocs);
+    try {
+      if (collection == null) {
+        solrClient.add(solrDocs);
+      } else {
+        solrClient.add(collection, solrDocs);
+      }
+    } catch (SolrException e) {
```
mrwrench24, line 258: another small thing: the solr exception is not declared in the methods we are using - unlike open/elasticsearch. it may be worth also adding a comment about this just in case people get confused (which I was at first :) )

mrwrench24 (commented):
some small comments!

kyle-kmw (approved):

spencersolomon6 (approved):

PR #506, opened 2026-06-03
LC-883: Preset Configs for API

mrwrench24:
Allows API users to specify, in the API Configuration, a directory containing Configs (either in `.conf` or `.json`) which they would like to be loaded as part of the initialization of the Lucille API. These Configs are then stored / keyed by their filenames, including their extension. So, they would be run by passing the filename as the argument.


lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -100,6 +100,23 @@ public synchronized String createConfig(Config config) {
...
+   * Adds the provided Config keyed under the provided name. Returns whether the config
+   * was stored successfully. This method will <i>not</i> overwrite an existing config
+   * with the same name.
+   * @param config The configuration to store.
+   * @param name The name under which the configuration should be keyed / referenced.
+   * @return Whether the config was successfully stored and associated with the provided name.
+   */
+  public synchronized boolean createConfigWithName(Config config, String name) {
+    if (configMap.containsKey(name)) {
+      return false;
```
kyle-kmw, line 114: log a warning if this collision occurs

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/PresetConfigHandler.java
```diff
@@ -0,0 +1,58 @@
...
+
+    try (Stream<Path> files = Files.list(configDirectoryPath)) {
+      files
+          .filter(Files::isRegularFile)
+          .filter(path -> {
+            String extension = FilenameUtils.getExtension(path.getFileName().toString());
+            return extension.equalsIgnoreCase("conf") || extension.equalsIgnoreCase("json");
+          })
+          .forEach(path -> {
+            Config config = ConfigFactory.parseFile(path.toFile()).resolve();
```
kyle-kmw: It would be good to try/catch around this call to resolve the config and log a warning for any configs in the directory that can't be resolved. As is I think a single invalid config would bubble up into a runtime exception that causes the API to crash.

kyle-kmw (approved):

doc/site/content/en/docs/operations/api.md
```diff
@@ -73,3 +73,24 @@ All endpoints are under the `/v1` prefix.
...
 2. `POST /v1/run` with `{"configId": "<uuid>"}` — receive a `runId`.
 3. `GET /v1/run/{runId}` — poll for run status until complete.
+
+---
+
+## Configuring the API
+
+It is important to draw a distinction between configuration for the Lucille API and configuration for Lucille (a Lucille run).
+Configuration for the Lucille API takes the form of a `.yml` file and controls an instance of the API. Within the API,
+you'll upload your Lucille Configs (as JSON) to ultimately run them.
```
rseitz: re "as JSON" -- I think the API accepts HOCON

doc/site/content/en/docs/operations/api.md
```diff
@@ -73,3 +73,24 @@ All endpoints are under the `/v1` prefix.
...
+
+You can update your Lucille API configuration to define a directory containing preset Configs you would like to be loaded
+upon initialization of the Lucille API. 
+
+```yaml
+presetConfig:
+  configDirectoryPath: /path/to/my/configs
+```
+
+The provided path **must** be a directory. Only `.json` and `.conf` files in this directory will be considered. Instead of a UUID,
```
rseitz: should we add `.hocon` here? 
since `.conf` is the most common one, we should mention it first in the list

lucille-plugins/lucille-api/README.md
```diff
@@ -157,6 +157,20 @@ curl -k -u user:password https://localhost:8443/v1/config -H "Content-Type: appl
...
+### Preset (Lucille) Configuration
+In your API configuration, you can declare a path to a directory containing _Lucille_ configurations you would like to be loaded 
+during initialization of the API.
+
+```yaml
+presetConfig:
+  configDirectoryPath: /path/to/my/configs
+```
+
+The provided path **must** be a directory. Only `.json` and `.conf` files in this directory will be considered.
```
rseitz: mention .conf first
add .hocon

lucille-plugins/lucille-api/src/test/resources/presets/incremental.conf
```diff
@@ -0,0 +1,53 @@
+connectors: [
```
rseitz, line 2: if we're adding detailed configs that never get executed anywhere, please add a comment at the top of the config saying what its purpose is -- i.e. it's just to test that the API finds it and parses it properly

lucille-plugins/lucille-api/src/test/java/PresetConfigHandlerTest.java
```diff
@@ -0,0 +1,55 @@
...
+  Path config1Path = Paths.get("src/test/resources/test_presets/config1.conf");
+
+  @Test
+  public void testFetchConfigs() throws Exception {
+    PresetConfigHandler handler = new PresetConfigHandler(presetsPath);
+    Map<String, Config> presetMap = handler.fetchConfigs();
+
+    assertEquals(4, presetMap.size());
+
+    assertEquals(1, presetMap.get("config1.conf").getInt("id"));
```
rseitz: since the config library gives multiple ways of loading configs which handle env vars differently, one of the things that could go wrong is that environment variables referenced in the config as in `${?AWS_S3_REGION}` don't get properly resolved. We should add such an env var to at least one of the configs we're testing with and check that it gets resolved by loading the associated value from the config and asserting on it. To do this, we'd need to set the system property before loading the config, then reset it after the test to whatever its previous value might have been (or unset it if it wasn't present), which would require checking/storing its value before we set it.

rseitz: also, since configs can have includes, it would be good to know & document whether they work as expected in the scenario of the API loading the config

rseitz (changes requested):
code looks good; left some comments re: doc and testing

lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -134,6 +134,18 @@ public void testSimultaneousRuns() throws Exception {
     }
   }
 
+  @Test
+  public void testCreateConfigWithName() {
+    RunnerManager runnerManager = RunnerManager.getInstance();
+    Config config = ConfigFactory.load("RunnerManagerTest/sleep.conf");
+    boolean success = runnerManager.createConfigWithName(config, "sleep");
```
rseitz: we're registering a "sleep" config here that will be visible to any other tests running in the same JVM. We already had this kind of leakage in other tests that call `runnerManager.createConfig(config)` but in those cases, what gets registered is a random UUID that other tests are not likely to ever try to access

if we had a runnerManager.removeConfig() method we could clean up at the end of this test, but we should probably save that for a separate PR to keep this one manageable

so, how about replacing "sleep" in this test with a random UUID

mrwrench24:
running into some issues with env var resolution and test failures...

mrwrench24:
Particularly frustrating is the fact that this DOES work in practice with just a simple call to `.resolve()`. But not during the actual tests.
```
curl -X GET "http://localhost:8080/v1/config" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   247  100   247    0     0   5467      0 --:--:-- --:--:-- --:--:--  5488
{
  "config1.conf": {
    "id": 1,
    "environmentValue": "a_dummy",
    "an_object": {
      "key": "value"
    },
    "object": {
      "key": "value"
    }
  ...
}
```

mrwrench24:
okay... looks like the config issues have (hopefully) been resolved peacefully

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/PresetConfigHandler.java
```diff
@@ -0,0 +1,77 @@
...
+            return ACCEPTED_EXTENSIONS.contains(extension);
+          })
+          .forEach(path -> {
+            try {
+              // for the purposes of actually RUNNING the API, a simple call to
+              // .resolve() would suffice here. However, it does not work
+              // with unit tests. This more explicit call works with both, appears
+              // to have the same effect, and does not create disgustingly large configs.
+              Config parsed = ConfigFactory.parseFile(path.toFile());
+              Config resolved = parsed.resolveWith(
```
rseitz: wondering if the unit test issue was related to caching?
There's a ConfigFactory.invalidateCaches() method that primarily exists for use in unit tests
(I've had to use it in a different project that depends on lucille)
what happens if you go back to the simple resolve() call, but make sure the unit tests call     ConfigFactory.invalidateCaches() before loading configs?
or is the current approach better for other reasons that we want to preserve?

rseitz: just noticed ConfigFactory.invalidateCaches() is already being used in at least one place, so maybe you already considered this...

mrwrench24: yes... currently calling `invalidateCaches()` after setting the system property in the unit test which I believe is in line with how the method works. 

when we just use `ConfigFactory.parseFile().resolve()` the unit test does not succeed - both on its own and during the build. but environment variables were resolved appropriately in practice.

so really, the only advantage of this current approach is that it allows us to have a unit test for environment var resolution in preset configs. I'm not sure if there is something off with the test...?

lucille-plugins/lucille-api/README.md
```diff
@@ -157,6 +157,23 @@ curl -k -u user:password https://localhost:8443/v1/config -H "Content-Type: appl
...
+```yaml
+presetConfig:
+  configDirectoryPath: /path/to/my/configs
+```
+
+The provided path **must** be a directory. Only `.conf`, `.json`, and `.hocon` files in this directory will be considered.
+
+In lieu of a UUID, preset configs will be keyed by their filename, including their file extension (e.g. `config1.conf`).
+
+Preset configs can use environment variables / `include` other configs as normal. It may be good practice to place these "included"
```
rseitz: it's worth noting that this is different from how uploaded configs work
when a config is sent to the API, variables are NOT resolved; however, when they are provided as presets, variables are resolved when the API is started (and are not re-resolved later, even if environment values change on the system)
this difference is probably what we want. that is, while we definitely want variables in preset configs to be resolved, we probably don't want to resolve variables in config that are shipped to the API (this would effectively allow any user with API access to see any env var setting on the host machine). So, we should probably leave things as they are and just document this difference as intended. The doc should go here but there may be other places (like the docs for the API method that accepts configs) where we should also put a note about this.

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/PresetConfigHandler.java
rseitz: consider if this is slightly clearer: 
`Config resolved = ConfigFactory.defaultOverrides().withFallback(parsed).resolve();`

rseitz: config1.conf uses key names that are pretty generic like "id"
it's possible that resolve() was bringing in values from an "application.conf" or "reference.conf" in another library on the classpath that conflicted?
I think the approach you eventually took is better since, like you wrote, it gives smaller configs with less chance of conflict

rseitz (approved):
approving now
have made some comments about doc we should include and a possible minor simplification in the confg loading line
please address these before merging

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/PresetConfigHandler.java
mrwrench24: maybe... the issue was that there just was no resolution of `DUMMY_VAR` at all

PR #507, opened 2026-06-03
LC-875: fix doc site links via versionsworkflow script

Otto-Cline:
Adds code to our versionsworkflow.sh script that takes the broken links from each version of the site and corrects their paths to match their respective versions


.github/workflows/hugo.yml
```diff
@@ -4,7 +4,7 @@ name: Deploy Hugo site to Pages
 on:
   # Runs on pushes targeting the default branch
   push:
-    branches: ["main"]
+    branches: ["LC-880-docs-version-for-main"]
```
rseitz: change this back to main?

spencersolomon6 (approved):

PR #508, opened 2026-06-03
LC-885: Config "Locking" in API

mrwrench24:
This PR gives users the option to `preventConcurrentRuns` of the same Config in their API configuration. When enabled, attempts to create simultaneous runs with the same `configId` will fail until the first run completes. 

As part of this PR, `LucilleResource.startRun` now takes in a `RunRequest` (previously a `Map<String, String>`). This was because the first version added an extra parameter to this request (which was not a `String`). I figured I would leave it, we can revert back to the `Map` if we want!


lucille-plugins/lucille-api/src/test/java/APIPreventConcurrentIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+        .header(HttpHeaders.AUTHORIZATION, authHeader).post(Entity.entity(IMDB_JSON, MediaType.APPLICATION_JSON));
+    String configResponse2 = configStatus2.readEntity(String.class);
+    // same retrieval as in APIIntegrationTest
+    String configId2 = configResponse2.substring(13, configResponse2.length() - 2);
+
+    // There should be enough room to prevent race conditions here.
+    Response runPostStatus = client.target(url + "v1/run").request()
+        .header(HttpHeaders.AUTHORIZATION, authHeader).post(Entity.entity("{\"configId\": \"" + configId1 + "\"}", MediaType.APPLICATION_JSON));
+    assertEquals(200, runPostStatus.getStatus());
+    System.out.println("ENTITY 1: " + runPostStatus.readEntity(String.class));
```
rseitz: remove System.out.println

lucille-plugins/lucille-api/src/test/java/APIPreventConcurrentIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    // identical config, but referenced by a different ID, so it can run.
+    Response runPostStatus2 = client.target(url + "v1/run").request()
+        .header(HttpHeaders.AUTHORIZATION, authHeader).post(Entity.entity("{\"configId\": \"" + configId2 + "\"}", MediaType.APPLICATION_JSON));
+    assertEquals(200, runPostStatus2.getStatus());
+
+    Response rejectedStatus = client.target(url + "v1/run").request()
+        .header(HttpHeaders.AUTHORIZATION, authHeader).post(Entity.entity("{\"configId\": \"" + configId1 + "\"}", MediaType.APPLICATION_JSON));
+    System.out.println("ENTITY 2: " + rejectedStatus.readEntity(String.class));
```
rseitz: remove System.out.println

lucille-plugins/lucille-api/src/test/java/APIPreventConcurrentIntegrationTest.java
```diff
@@ -0,0 +1,145 @@
...
+  {
+    "connectors": [
+      {
+        "class": "com.kmwllc.lucille.connector.CSVConnector",
+        "name": "imdb-connector",
+        "pipeline": "imdb-pipeline",
+        "path": "classpath:APIPreventConcurrentIntegrationTest/imdb.csv"
+      }
+    ],
+    "pipelines": [
```
rseitz, line 28: does the pipeline need to be this elaborate for what we are testing here?

lucille-plugins/lucille-api/src/test/resources/APIPreventConcurrentIntegrationTest/imdb.csv
```diff
@@ -0,0 +1,17 @@
+title,vote_average,vote_count,status,release_date,revenue,runtime,adult,budget,imdb_id,original_language,original_title,overview,popularity,tagline,genres,production_companies,production_countries,spoken_languages,keywords
```
rseitz: do we really need all this test data for a test that never asserts anything about the pipeline output?

is the aim just to get the pipeline to take a certain amount of time. for that, we could look at using SleepConnector or adding a sleep option to NopStage (if we want the sleep to happen inside a stage, not a connector)

lucille-plugins/lucille-api/src/test/resources/LucilleResourceTest/imdb.csv
```diff
@@ -0,0 +1,17 @@
+title,vote_average,vote_count,status,release_date,revenue,runtime,adult,budget,imdb_id,original_language,original_title,overview,popularity,tagline,genres,production_companies,production_countries,spoken_languages,keywords
```
rseitz: same comment as in the other instance of this -- do we really need all this test data given what we are actually testing?

lucille-plugins/lucille-api/src/test/resources/APIPreventConcurrentIntegrationTest/imdb.csv
rseitz: regarding SleepConnector, I know it is currently in lucille-core's test tree, so it may not be accessible here. I propose moving it to lucille-core's src/main. There's a reasonable argument that someone might want to actually use it while doing some real-world prototyping, so it wouldn't be completely unwanted or out-of-place in src/main

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/objects/RunRequest.java
```diff
@@ -0,0 +1,21 @@
+package com.kmwllc.lucille.objects;
+
+import jakarta.validation.constraints.NotBlank;
+
+/**
+ * A request to start a run in the Lucille API.
+ */
+public class RunRequest {
+
+  @NotBlank
```
rseitz: looks like this NotBlank annotation is only used when jakarta bean validation is triggered, which we're not currently doing. suggest simply removing it because it's a bit misleading as is

rseitz (changes requested):
code looks good but left some comments about testing
main comment is there seems to be way too much test data and pipeline complexity for what we are actually testing here
if we just need a way to make a run take some time, let's move SleepConnector into lucille-core src/main so we can use it

lucille-plugins/lucille-api/src/test/resources/APIPreventConcurrentIntegrationTest/imdb.csv
mrwrench24: I will move the `SleepConnector` to `Lucille-core`. I was wondering whether we may want it there or not.

This config / run comes from the API's "non-trivial run" test I remember working on last year. But I initially made a `SleepConnector` config until I remembered/realized it was not exposed.

lucille-plugins/lucille-api/src/test/java/endpoints/LucilleResourceTest.java
```diff
@@ -16,6 +19,29 @@
...
 public class LucilleResourceTest {
 
+  private static final String SLEEP_JSON = """
+  {
+    "connectors": [
+      {
+        "class": "com.kmwllc.lucille.connector.SleepConnector",
+        "name": "sleep-connector",
+        "pipeline": "pipeline1",
+        "duration": 1000
```
rseitz: do we need a full second here? 
would rather set it too low, get an intermittent failure, and have to bump this up than to start out assuming 1s is the necessary cost of this test

lucille-plugins/lucille-api/src/test/java/APIPreventConcurrentIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+public class APIPreventConcurrentIntegrationTest {
+
+  private static final String SLEEP_JSON = """
+  {
+    "connectors": [
+      {
+        "class": "com.kmwllc.lucille.connector.SleepConnector",
+        "name": "sleep-connector",
+        "pipeline": "pipeline1",
+        "duration": 1000
```
rseitz: same comment as elsewhere:
do we need a full second here? 
would rather set it too low, get an intermittent failure, and have to bump this up than to start out assuming 1s is the necessary cost of this test

rseitz (approved):
approving but with a request to reduce the 1s sleep duration in tests

lucille-plugins/lucille-api/src/test/java/APIPreventConcurrentIntegrationTest.java
mrwrench24: I reduced the time. After a further look:

I will note that in both tests, as in other API tests, we don't wait / block for the runs to complete so there is not an effect on the test time. It really just affects how long the run is "out there" for. I think it would be pretty straightforward to block on them (using `RunnerManager`) in these tests + others if we want.

kyle-kmw (approved):

Otto-Cline (approved):

PR #510, opened 2026-06-04
LC-844: Retries for the Elasticsearch Indexer

Otto-Cline:


mrwrench24 (approved):
LGTM (but wait for others :) )

lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
kyle-kmw, line 225: deleteByQuery should have retries similar to OpenSearchIndexer

kyle-kmw, line 191: deleteById should have retries similar to OpenSearchIndexer

kyle-kmw (changes requested):
I think deleteById and deleteByQuery need to have retires wired in similar to the OpenSearchIndexer implementation and test cases should be expanded to cover those delete scenarios with retries.

kyle-kmw (approved):

PR #511, opened 2026-06-04
LC-888: Support configuring an external version field with the OpenSearchIndexer.

kyle-kmw:


lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -78,6 +79,7 @@ public OpenSearchIndexer(Config config, IndexerMessenger messenger, boolean bypa
     this.update = config.hasPath("opensearch.update") ? config.getBoolean("opensearch.update") : false;
     this.versionType =
         config.hasPath("indexer.versionType") ? VersionType.valueOf(config.getString("indexer.versionType")) : null;
+    this.versionField = config.hasPath("indexer.versionField") ? config.getString("indexer.versionField") : null;
```
rseitz, line 82: looks like this will be ignored if the user doesn't also specify versionType
should add validation that requires versionType if versionField is provided (or at least log a warning)

lucille-core/src/test/java/com/kmwllc/lucille/indexer/OpenSearchIndexerTest.java
```diff
@@ -760,6 +760,86 @@ public void testRoutingAndUpdate() throws Exception {
...
+    List<BulkOperation> requests = br.operations();
+    IndexOperation indexRequest = requests.get(0).index();
+
+    assertEquals("doc1", indexRequest.id());
+    // Should be null because field is missing and it's not a KafkaDocument
+    assertEquals(null, indexRequest.version());
+  }
+
+  @Test
+  public void testDocumentVersioningWithoutVersionField() throws Exception {
```
rseitz: seems very similar to testDocumentVersioning() -- same scenario with fewer assertions? consider removing or else document what this is covering that isn't already covered

rseitz (changes requested):
looks good overall but left some comments about param validation and possible duplicate test coverage

rseitz (approved):

PR #512, opened 2026-06-05
LC-871: RunDetails Map Cleanup

mrwrench24:
This PR updates the `RunnerManager` to only store the historical details of the last 10,000 runs to prevent unbounded heap usage when using the API. I did some "stress" testing locally and saw more reliable / appropriate heap usage with these changes allowed.

As currently implemented, this limit is not configurable or exposed to users. We can change the value easily. If we want this to be configurable, we would have to make changes regarding `RunnerManager` being a singleton.

This PR also includes changes from `LC-891` - limiting the capacity of the `configMap`


lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -54,7 +75,8 @@ public synchronized boolean isRunning(String runId) {
    * @throws Exception if the run throws an exception or is interrupted
    */
   public void waitForRunCompletion(String runId) throws Exception {
-    RunDetails details = runDetailsMap.get(runId);
+    RunDetails details = getRunDetails(runId);
```
mrwrench24, line 90: w/ the behavior / invariants of the two maps, it would be safe to just check the active map were it not for throwing an exception. this is also a function used exclusively in testing.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -180,4 +224,23 @@ public Set<String> getConfigKeys() {
   public Config getConfig(String configId) {
     return configMap.get(configId);
   }
+
+  /**
+   * Modify the maximum number of historical runs that will be stored for testing purposes.
+   * Resets the run history.
+   * Package access for unit testing.
+   */
+  static synchronized void setMaxHistoryForTesting(int maxHistory) {
```
rseitz: since this method is static, the "synchronized" keyword uses the class as the monitor, meaning other threads can't concurrently call this method or other static synchronized methods... but they _can_ call synchronized instance methods, because those are synchronized on the instance itself, not the class. So the way this is set up you could have threads calling RunnerManager.setMaxHistoryForTesting() and RunnerManager.getInstance().getRunDetails() at the same time, I think...

I'd suggest just removing "synchronized" here because it's confusing/misleading and add a line to the javadoc saying this isn't meant for multi-threaded use -- in a testing context with multiple threads, call it before the threads are started
same with resetMaxHistoryForTesting()

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -143,6 +165,17 @@ public synchronized RunDetails runWithConfig(String runId, String configId)
       } catch (Exception e) {
         log.error("Failed to run lucille with ID '{}' via the Runner Manager.", runId, e);
         runDetails.setThrowable(e);
+      } finally {
+        synchronized (instance) {
+          historicalDetailsMap.put(runId, runDetails);
+
+          if (historicalDetailsMap.size() > maxHistory) {
```
rseitz: this looks correct and is not that much code, but if we override removeEldestEntry() when we declare historicalDetailsMap we can make it automatically clean up old entries so we don't have to worry about the cleanup here 

that would look like this: 

```
private final Map<String, RunDetails> historicalDetailsMap =
    new LinkedHashMap<>() {
      @Override
      protected boolean removeEldestEntry(Map.Entry<String, RunDetails> eldest) {
        return size() > maxHistory;
      }
    };
```

it's more of a nicety... consider if it's worthwhile

rseitz (approved):

PR #513, opened 2026-06-05
LC-823: QueryOpensearch stage uses OpenSearchUtils

Otto-Cline:
The QueryOpensearch class does not currently make use of OpenSearchUtils, which handles cases where an OpenSearch cluster requires auth or TLS configured. This causes QueryOpensearch to fail to authenticate or on cert validation.

This PR allows the QueryOpensearch stage to properly interface with an auth/tls required instance. I also did some minor refactoring on the QueryOpensearchTest tests that flex this new behavior and mimic other in place testing patterns better


spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -148,18 +176,19 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+        .json(requestJson.toString())
         .build();
 
     JsonNode responseNode;
 
-    try {
-      HttpResponse<String> response = httpClient.send(httpRequest, HttpResponse.BodyHandlers.ofString());
-      responseNode = mapper.readTree(response.body());
+    try (Response response = client.generic().execute(request)) {
+      String responseBody = response.getBody().map(Body::bodyAsString).orElse("");
```
mrwrench24: the only thing I am curious about is our handling of empty response bodies. since the stage is all about executing search templates, I don't think we would expect empty response bodies. the current implementation would just place `null` in the `destinationField` upon an error, since we just parse an empty string.

```java
@Test
  public void testMissingResponseBody() throws Exception {
    Response mockResponse = mock(Response.class);
    when(mockResponse.getBody()).thenReturn(Optional.empty());

    OpenSearchGenericClient mockGeneric = mock(OpenSearchGenericClient.class);
    when(mockGeneric.execute(any())).thenReturn(mockResponse);

    OpenSearchClient mockClient = mock(OpenSearchClient.class);
    when(mockClient.generic()).thenReturn(mockGeneric);

    QueryOpensearch stage = (QueryOpensearch) factory.get("QueryOpensearchTest/searchTemplate.conf");
    stage.setClient(mockClient);

    Document testDoc = Document.create("neck_creek");
    testDoc.setField("park_name", "Neck Creek Preserve");
    stage.processDocument(testDoc);

    System.out.println(testDoc);

    assertEquals(Double.valueOf(6.7708125), testDoc.getDouble("response"));
  }
```

mrwrench24 (commented):
one small question

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
Otto-Cline: good catch; we actually weren't throwing when an error code is returned by the client either, so I updated the method to handle that

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
```diff
@@ -148,20 +176,36 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
...
+
+    try (Response response = client.generic().execute(request)) {
+      status = response.getStatus();
+      responseBody = response.getBody().map(Body::bodyAsString).orElse("");
+    } catch (Exception e) {
+      throw new StageException("Error occurred sending the Opensearch query.", e);
+    }
+
+    // The client will return errored status codes instead of throwing if there was an issue, so we need to catch these ourselves
+    if (status < 200 || status >= 300 || responseBody.isBlank()) {
```
mrwrench24: another small thing, but we could maybe move this in to the `try` block and change the `catch` to only catch the `IOException` from `.execute`. This would prevent us from having to use `orElse` and we could instead just check whether the body `isEmpty` or not.

mrwrench24 (approved):
lgtm with one very minor stylistic suggestion

lucille-core/src/main/java/com/kmwllc/lucille/stage/QueryOpensearch.java
Otto-Cline: I moved the conditional and mapper.readTree into the try catch for uniformity and since an ioexception applies to both the execute() and readTree() call (JsonProcessingException is a subclass of ioexception). I kept orElse() because I think a present but blank Body would still slip through if we used isEmpty, since it only checks that the Optional object itself is missing a Body object.

mrwrench24:
You don't have to copy this but this is a rough sketch of what I was thinking:

```java
try (Response response = client.generic().execute(request)) {
      int status = response.getStatus();

      if (status < 200 || status >= 300 || response.getBody().isEmpty()) {
        throw new StageException("OpenSearch returned an unexpected response to a search template query (HTTP " + status
            + "). Response body: " + response.getBody());
      }

      String responseBody = response.getBody().get().bodyAsString();
      // The client will return errored status codes instead of throwing if there was an issue, so we need to catch these ourselves

      responseNode = mapper.readTree(responseBody);
    } catch (IOException e) {
      throw new StageException("Error communicating with OpenSearch or parsing its response.", e);
    }
```

Basically just avoiding using an "orElse" value, where we are kind of using an empty string as a sentinel for an errored response. I think a "missing" body is an unexpected response. let me know if that makes sense

mrwrench24 (approved):
LGTM sorry for back and forth

PR #515, opened 2026-06-05
LC-890: Metrics Cleanup

mrwrench24:
Lucille Runs create `MetricRegistry`s under the `SharedMetricsRegistry`. In many cases (and for every run in the API), these are prefixed by the runID. They have never been cleaned up. This PR has the `RunnerManager` remove, after a run, all metrics starting with the runID. A straightforward unit test ensures this works, but it will also warn us if additional metrics are being introduced and not removed when needed.


mrwrench24:
will reopen w/ a similar PR w/ main as base since this was pretty straightforward.

PR #516, opened 2026-06-08
LC-891: Limited ConfigMap Capacity

mrwrench24:
Similar to run history, users can store up to 10,000 Configs when running the API. However, if they try to create their 10,001st config, their request will be rejected. So, there is now a `DELETE` endpoint in the API allowing you to do so.

Similar to a previous PR, the `CreateConfigResult` was created since the response in an earlier version of this PR had two fields instead one. Now, we are back to only one field, but I figured it would be worth leaving in the code. 

Also, note that the base for this PR is LC-871.


mrwrench24:
An overkill `APIIntegration` test that takes ~7 sec but ensures the removed ID is part of the response appropriately

```java
@Test
  public void testConfigRemoval() {
    Response toRemoveResp = client.target(url + "v1/config").request()
        .header(HttpHeaders.AUTHORIZATION, authHeader).post(Entity.entity("{}", MediaType.APPLICATION_JSON));
    String idToRemove = toRemoveResp.readEntity(CreateConfigResult.class).getConfigId();

    for (int i = 1; i < 10000; i++) {
      client.target(url + "v1/config").request()
          .header(HttpHeaders.AUTHORIZATION, authHeader).post(Entity.entity("{}", MediaType.APPLICATION_JSON));
    }

    Response configStatus = client.target(url + "v1/config").request()
        .header(HttpHeaders.AUTHORIZATION, authHeader).post(Entity.entity("{}", MediaType.APPLICATION_JSON));

    CreateConfigResult result = configStatus.readEntity(CreateConfigResult.class);
    assertEquals(idToRemove, result.getRemovedConfigId());
  }
```

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -107,7 +108,7 @@ public synchronized RunDetails run(String runId) throws RunnerManagerException {
    * @throws RunnerManagerException
    */
   protected synchronized RunDetails runWithConfig(Config config) throws RunnerManagerException {
-    String configId = createConfig(config);
+    String configId = createConfig(config).getConfigId();
```
mrwrench24: this method, as well as the one above, are unused. if they do become used, this would be a case where the user would not be told of a removed config ID.

rseitz:
Looking over this implementation, I think ConfigMaps have a different usage pattern to RunResults (in the LC-871 PR). ConfigMaps can be re-used indefinitely and someone could theoretically create one config first that they "love" and want to keep reusing, then create 10K configs each for one-time use. But the current PR would evict their favorite config. Unlikely scenario but possible.

There's a somewhat subtle/fancy way we could change this PR to evict the least recently _used_ configs instead of the least recently added ones. The LinkedHashMap constructor takes a boolean arg we can set to true to get LRU semantics along with overriding removeEldestEntry(). However, I'm still not that comfortable with deciding on the user's behalf which reusable config we're going to evict.

I'd propose an alternate approach for this. We can impose a hard limit on the number of configs that can be registered _and_ add an API method that lets a user delete configs, so they can clean up manually/deliberately if they ever reach the limit. 

We wouldn't need CreateConfigResult for this simpler approach, but should think through whether it's still worth keeping for other reasons?

rseitz (changes requested):
suggest imposing a hard limit on registered configs and adding an API method that allows manual cleanup

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -96,58 +98,55 @@ public void waitForRunCompletion(String runId) throws Exception {
...
-   * Run with config and an internal runId will be generated
-   * 
-   * @param config
-   * @return RunDetails
-   * @throws RunnerManagerException
-   */
-  protected synchronized RunDetails runWithConfig(Config config) throws RunnerManagerException {
-    String configId = createConfig(config);
-    return runWithConfig(Runner.generateRunId(), configId);
-  }
```
mrwrench24, line 122: decided to delete these unused methods. if we end up wanting them, we'll want to make sure they interact nicely with the config limit etc.

mrwrench24:
pr description has small discussion of `CreateConfigResult` as well - I think it is worth keeping but we can remove it

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
```diff
@@ -96,58 +98,55 @@ public void waitForRunCompletion(String runId) throws Exception {
...
    * @param config The configuration to store.
-   * @param name The name under which the configuration should be keyed / referenced.
-   * @return Whether the config was successfully stored and associated with the provided name.
+   * @param desiredKey The key under which the configuration should be stored.
+   * @return Whether the config was successfully stored and associated with the provided desiredKey.
+   */
+  public synchronized boolean createConfigWithKey(Config config, String desiredKey) throws RunnerManagerException {
+    if (configMap.containsKey(desiredKey)) {
+      log.warn("Attempted to add config with key {}, which already exists", desiredKey);
+      return false;
```
rseitz: I'm thinking we should change this to throwing an exception to be consistent with how we're throwing an exception when max configs is exceeded

this method could then return void instead of boolean

I don't see anywhere except for tests where we actually look at the return value of this method, so the rest of the code might be telling us this is fine to change :)

but let me know if you see a reason to handle existing-key and max-configs-exceeded differently

rseitz (approved):
approving with suggestion to change return value of createConfigWithKey() to void and throw exception for duplicate keys

PR #517, opened 2026-06-08
LC-890: SharedMetricRegistries Cleanup

mrwrench24:
Lucille Runs create `MetricRegistry`s under the `SharedMetricsRegistry`. In many cases (and for every run in the API), these are prefixed by the runID. They have never been cleaned up, because we have never needed to. This PR has the `RunnerManager` remove, after a run, all metrics starting with the runID. A straightforward unit test ensures this works, but it will also warn us if additional metrics are being introduced and not removed when needed.


lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerManagerTest.java
```diff
@@ -197,6 +200,36 @@ public void testNonTrivialSimultaneousRuns() throws Exception {
...
+    runnerManager.runWithConfig("run-1", noopConfigId);
+    runnerManager.runWithConfig("run-2", noopConfigId);
+
+    runnerManager.waitForRunCompletion("run-1");
+    runnerManager.waitForRunCompletion("run-2");
+
+    // default registry gets recreated, but should be empty
+    MetricRegistry metricRegistry = SharedMetricRegistries.getOrCreate(LogUtils.METRICS_REG);
+
+    // Two indexer metrics get "left behind" but are not related to a specific run.
```
rseitz: What are these metrics that get left behind?
The intent in Indexer is that all metrics get created with an appropriate metricsPrefix, which should always start with the runId in local mode. If there's somewhere this isn't happening, let's investigate...

mrwrench24: `.indexer.batchTimeOverSize` and `.indexer.docsIndexed`.

I did a quick look and it looks like this is the result of running in validation mode.

We could provide a special metrics prefix during all of the construction we do during validation. That would allow us to easily clear these "dummy metrics". But, this shouldn't be a "memory leak" - we'll only ever have these two metrics floating around.

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
rseitz: I know this line wasn't changed as part of this PR but I think we may want to change it here.
I don't think we want to call runWithResultLog() in this context but should call run() instead.

runWithResultLog() outputs logs report via SharedMetricsRegistry that isn't filtered by run ID, so it could be confusing in an API scenario with concurrent runs... and it's never returned to the caller anyway, so it's only potentially visible to an admin:

```      
Slf4jReporter.forRegistry(SharedMetricRegistries.getOrCreate(LogUtils.METRICS_REG))
.outputTo(log).withLoggingLevel(getMetricsLoggingLevel(config)).build().report();
```

rseitz: Spent some time exploring this. I think we should move the metrics cleanup into Runner. The reasoning is this, since RunnerManager doesn't create or interact with metrics directly, so it's not the ideal class to have the responsibility for cleaning them up. Would be better to move that responsibility into Runner. Would be easy to move the cleanup into the current runWithResultLog(), but that method always prints a metrics report which we don't want to do in the API (per my first comment in this thread).

To resolve this, I think we _can_ move the cleanup to runWithResultLog() but add a boolean arg to that method that controls whether we log the report... and take the chance to rename the method while we're at it. The refactored method would be something like:

```
public static RunResult runAndReport(Config config, RunType runType, String runId, boolean reportMetrics)
    throws Exception {
  StopWatch stopWatch = new StopWatch();
  stopWatch.start();

  try {
    RunResult result = run(config, runType, runId);

    if (reportMetrics) {
      Slf4jReporter.forRegistry(SharedMetricRegistries.getOrCreate(LogUtils.METRICS_REG))
          .outputTo(log).withLoggingLevel(getMetricsLoggingLevel(config)).build().report();
    }
    log.info(result.toString());
    return result;
  } finally {
    stopWatch.stop();
    log.info(String.format("Run took %.2f secs.",
        (double) stopWatch.getTime(TimeUnit.MILLISECONDS) / 1000));
    // Clean up metrics prefixed by this run's ID.
    if (runId != null) {
      SharedMetricRegistries.getOrCreate(LogUtils.METRICS_REG)
          .removeMatching(MetricFilter.startsWith(runId));
    }
  }
}
```

rseitz (changes requested):

lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java
mrwrench24: okay... it took me a bit longer to wrap my ahead around some of the changes for some reason but I think the changes are hopefully in line with what you're looking for...

I found it most straightforward to require that the runId presented to the renamed method now responsible for metrics cleanup is not null. It felt like, without this requirement, the code was a bit complex and it could lead to issues for callers - providing a run id does metrics cleanup, using a generated one doesn't, etc...

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -463,33 +485,27 @@ public static RunResult runWithResultLog(Config config, RunType runType, String
...
-   *
-   * @param config
-   * @param type
-   * @return
-   * @throws Exception
-   */
-  public static RunResult run(Config config, RunType type) throws Exception {
-    return run(config, type, null);
+      SharedMetricRegistries.getOrCreate(LogUtils.METRICS_REG)
+          .removeMatching(MetricFilter.startsWith(runId));
```
rseitz, line 506: should make sure runId is not actually null
since the javadoc says it shouldn't be null, we should probably do the check near the top of the method and throw an illegalargumentexception if we find otherwise
(if the javadoc didn't say a non-null runId is an explicit requirement, then simply wrapping this SharedMetricsRegistries cleanup line in a null-check would be fine)

rseitz (approved):

PR #518, opened 2026-06-08
LC-872: "Safe" Incremental FileConnector for API

mrwrench24:
Here goes:

The JavaDoc for the `FileConnectorStateManager` notes that it relies on Connectors running sequentially and in a single thread. While this is not changed by this PR, what do have to contend with the fact that multiple _runs_ within the same JVM may exist (via the API).

Connectors (can) have _side effects_. For example, the `FileConnector` can move files after they are processed. So, generally, there is an onus on users of the API to ensure that the Connectors they are using _in separate runs_ do not have overlaps. However, the state database is a service we provide to users - so we do want to take more responsibility for preventing them from ruining their stateful `FileConnector` traversals.

A stateful `FileConnector` has a pretty specific lifecycle. For example, they have the option of deleting rows for files that were not encountered. So, we don't just want to make this "thread-safe", we really do have to lock it down to these kinds of potential concurrent connections.

To that end, this PR specifically prevents users of the "stateful" File Connector from concurrently using the same `connectionString` _and_ `tableName` within the same JVM. We lock access using the string `{connectionString}:{tableName}`. The FileConnector will be blocked (waiting for a lock) when trying to make these a "concurrent connection".

Here's the twist: unfortunately, we also have to prevent unbounded heap usage (as we have been doing recently!). So, we build the `LockRegistry`, which has to maintain a "reference count" of sorts before it knows which key-value pairs can be safely removed from the map. The `FileConnectorStateManager` uses one of these (static) for _all_ `init`/`shutdown` calls to prevent data races within the same database / table.

TLDR: unfortunately we had to jump through a lot of hoops to limit a very specific issue in a safe & scalable way


mrwrench24:
Moving to draft since this is a lower priority + is subject to a second look / enhancement, perhaps...

mrwrench24:
One alternative I have been looking into: we could use `Striped.lazyWeakLock(256)`. The downside is that this involves hashing and introduces the (relatively high) risk for "false contention" of two file connectors (when there are many).

rseitz:
Just noting there's a merge conflict here to resolve. 

Context -- the merged PR #561 allows FileConnector to start multiple traversal _threads_ which may interact with the same state manager (so, same connectionString and tableName) -- all within one particular run. Meanwhile, this #518 PR prevents multiple concurrent _runs_ from using the same connectionString and tableName. These are two different purposes and -- maybe stating the obvious -- the goal of this #518 PR is still valid and needed. But now that #561 is merged we do need to resolve conflicts.

PR #522, opened 2026-06-11
LC-846: Add indexer batching based on payload byte size

Otto-Cline:
New config option batchByteSize: Adds up document byte sizes until cap is exceeded, causing batch to flush.


lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -99,6 +101,8 @@
 public abstract class Indexer implements Runnable {
 
   public static final int DEFAULT_BATCH_SIZE = 100;
+  public static final int NO_BATCH_SIZE = Integer.MAX_VALUE;
+  public static final long NO_BATCH_SIZE_BYTES = Long.MAX_VALUE;
```
spencersolomon6, line 107: Why is one of these an int and the other a long?

Otto-Cline: The byte size of individual documents can be very large, so the equivalent byte cap of a regular doc cap is much larger numerically

spencersolomon6 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -65,7 +65,9 @@
  *   <li>indexer.whitelist (List&lt;String&gt;, Optional) : Only include these fields from the document when sending to the
  *   destination.</li>
  *   <li>indexer.batchSize (Integer, Optional) : Number of documents to accumulate before sending to the destination. Defaults to
- *   {@value #DEFAULT_BATCH_SIZE}.</li>
+ *   {@value #DEFAULT_BATCH_SIZE} or {@value #NO_BATCH_SIZE} if batchByteSize is set.</li>
+ *   <li>indexer.batchByteSize (Long, Optional) : Total size of documents accumulated before sending to the destination. Defaults to
+ *   {@value #NO_BATCH_SIZE_BYTES}.</li>
```
mrwrench24, line 70: would recommend adding a small note emphasizing that it is valid to use both options, and when we do so, batches are flushed when either of the limits are breached

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -183,18 +187,26 @@ public Indexer(Config config, IndexerMessenger messenger, boolean bypass, String
...
+    int batchSize;
+    long batchByteSize;
+
+    if (config.hasPath("indexer.batchSize") || config.hasPath("indexer.batchByteSize")) {
+      batchSize = ConfigUtils.getOrDefault(config, "indexer.batchSize", NO_BATCH_SIZE);
+      batchByteSize = ConfigUtils.getOrDefault(config, "indexer.batchSizeBytes", NO_BATCH_SIZE_BYTES);
+    } else {
+      batchSize = DEFAULT_BATCH_SIZE;
+      batchByteSize = NO_BATCH_SIZE_BYTES;
+    }
```
mrwrench24, line 204: this is absolutely NOT a big deal, but I think we could call

`batchByteSize = ConfigUtils.getOrDefault(config, "indexer.batchSizeBytes", NO_BATCH_SIZE_BYTES)`

in both cases, and then make a call to `ConfigUtils.getOrDefault()` for `batchSize` with a different default based on whether or not there is `indexer.batchSizeBytes`

I'm not exactly sure what it would look like and if the code actually might end up being slightly more confusing, but maybe just something to consider!

lucille-core/src/main/java/com/kmwllc/lucille/core/SingleBatch.java
```diff
@@ -32,14 +35,17 @@ public SingleBatch(int capacity, int timeout) {
...
       queue.drainTo(docs);
+      byteAccumulator = doc.getByteSize();
     }
 
-    if (!queue.offer(doc)) {
+    if (byteAccumulator > byteCapacity || !queue.offer(doc)) {
       queue.drainTo(docs);
       queue.offer(doc);
+      byteAccumulator = doc.getByteSize();
     }
```
mrwrench24, line 53: If I'm not mistaken... I think we want to reset the `byteAccumulator` in `flush` too.

I can see it looks like we really call `flush` at the end of an indexer's life, but it may still be worth doing

mrwrench24 (commented):
a few small comments

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
```diff
@@ -845,6 +845,12 @@ public Map<String, Object> asMap() {
     return MAPPER.convertValue(data, TYPE);
   }
 
+  @Override
+  public long getByteSize() {
+    byte[] bytes = this.toString().getBytes();
```
rseitz: this string serialization is expensive and is likely going to be repeated when document is passed to the search client

we should be able to relax the requirement to return an exact byte size (in fact, the stringified json of the lucille doc is going to have a different number of bytes than the search engine actually uses to persist the document anyway)

we should be able to create an estimateJsonSize method that recursively walks the json tree. This would basically be a switch statement with cases for each json type that might be encountered, with recursive calls where the node we're looking at contains nested nodes. 

when we stringify the doc, the json tree needs to be walked anyway
it's cheaper to visit each node and just estimate its size & add the size to a running count, than to have to stringify the node and append it

lucille-core/src/main/java/com/kmwllc/lucille/core/HashMapDocument.java
```diff
@@ -612,6 +612,12 @@ public Map<String, Object> asMap() {
     return (Map<String, Object>)data.getData().clone();
   }
 
+  @Override
+  public long getByteSize() {
+    byte[] bytes = this.toString().getBytes();
```
rseitz, line 617: I left comment on JsonDocument about changing this to a more efficient estimation approach
that optimization is not necessary to do here because HashMapDocument is deprecated and will be removed from the codebase

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -183,18 +187,26 @@ public Indexer(Config config, IndexerMessenger messenger, boolean bypass, String
...
-        config.hasPath("indexer.batchSize")
-            ? config.getInt("indexer.batchSize")
-            : DEFAULT_BATCH_SIZE;
+
+    int batchSize;
+    long batchByteSize;
+
+    if (config.hasPath("indexer.batchSize") || config.hasPath("indexer.batchByteSize")) {
+      batchSize = ConfigUtils.getOrDefault(config, "indexer.batchSize", NO_BATCH_SIZE);
+      batchByteSize = ConfigUtils.getOrDefault(config, "indexer.batchSizeBytes", NO_BATCH_SIZE_BYTES);
```
rseitz: we're looking at indexer.batchByteSize above (l194) but here it's indexer.batchSizeBytes

how are tests able to pass with this issue?

rseitz (changes requested):

lucille-core/src/main/java/com/kmwllc/lucille/core/SingleBatch.java
Otto-Cline: good catch I should've been testing that

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
Otto-Cline: implemented and tested; the testGetByteSizeEstimate() test I did for this is a little wishy washy.. let me know if it's ok @rseitz

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
```diff
@@ -845,6 +846,55 @@ public Map<String, Object> asMap() {
...
+          size += estimateJsonSize(element);
+        }
+        return size;
+      }
+      case STRING:
+        return node.textValue().getBytes(StandardCharsets.UTF_8).length + 2; // content plus quotes
+      case NUMBER:
+        return node.asText().length();
+      case BOOLEAN:
+        return node.booleanValue() ? 4 : 5; // true / false
```
spencer-solomon-tfs, line 888: Why are true/false different byte sizes?

Otto-Cline: It's the literal length of the words true and false

spencersolomon6 (approved):

mrwrench24 (approved):

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
```diff
@@ -845,6 +846,55 @@ public Map<String, Object> asMap() {
...
+          if (!first) {
+            size += 1; // comma between elements
+          }
+          first = false;
+          size += estimateJsonSize(element);
+        }
+        return size;
+      }
+      case STRING:
+        return node.textValue().getBytes(StandardCharsets.UTF_8).length + 2; // content plus quotes
```
rseitz: calling getBytes() with UTF_8 as the requested encoding forces a new byte array to allocated with the data in that particular encoding (which is different from the encoding that String uses internally).

So, this getBytes() call is not just returning the internal byte array as-is, it's creating a new one in UTF_8, so it's a little more costly than it looks

we should be able to create a utf8Length() helper method that looks at at a string itself (e.g. the string returned by node.textValue()), iterates character by character and estimates the bytes that are required to represent that character in UTF_8, adding them all up

the same helper method could be used where we currently do entry.getKey().getBytes() on l866

iterating over the string directly in this way will probably be faster than calling getBytes() which needs to do basically the same iteration to generate the new byte[], but has the additional cost of taking up some memory and not computing the byte count that we want during the initial iteration

lucille-core/src/main/java/com/kmwllc/lucille/core/SingleBatch.java
```diff
@@ -32,14 +35,17 @@ public SingleBatch(int capacity, int timeout) {
   @Override
   public List<Document> add(Document doc) {
     List<Document> docs = new ArrayList<>();
+    byteAccumulator = byteAccumulator + doc.getByteSize();
```
rseitz: we don't need to track the size of the batch in bytes if we haven't been asked to -- otherwise it's unnecessary overhead

if "batchByteSize" hasn't been set then byteCapacity should be NO_BATCH_SIZE_BYTES = Long.MAX_VALUE

we can look for that and avoid doing doc.getByteSize whenever that's the case

lucille-core/src/test/java/com/kmwllc/lucille/core/SingleBatchTest.java
```diff
@@ -130,8 +130,210 @@ public void testNonTimeout() throws InterruptedException {
...
+    assertEquals(2, byteFlush.size());
+    assertEquals("s4", byteFlush.get(0).getId());
+    assertEquals("b1", byteFlush.get(1).getId());
+
+    List<Document> finalFlush = batch.flush();
+    assertEquals(1, finalFlush.size());
+    assertEquals("b2", finalFlush.get(0).getId());
+  }
+
+  private static Document bigDoc(String id) {
```
rseitz: minor, but this only used in one test, and only used twice there; probably not worth defining this as a separate helper method

rseitz (approved):

PR #523, opened 2026-06-12
LC-367: Support Child Docs for Elastic/OpenSearch Indexers

mrwrench24:
This PR allows users of the `ElasticsearchIndexer` and the `OpensearchIndexer` to index attached/nested child documents as nested fields in these search engines. Each indexer takes in a `childDocumentsField` parameter specifying which field they want to store child documents in.

The major question is why to use nested fields in lieu of parent/joins. We already support `join` configuration in `ElasticsearchIndexer`; my understanding is it would be straightforward (famous last words) to add it to `OpensearchIndexer`. If users want to use a parent/join, they can use `EmitNestedChildren`. I think they may need to use two configs / runs to accomplish this. 

There was also some brief discussion as to whether we want separate field filtering (white/blacklist) for the attached child documents themselves. In the current version of the PR, we mimic the `SolrIndexer` which just applies the top level filters to the attached children as well.

Potential follow-ups or things we should add to this PR: 
1. Support for parent joins in Opensearch
2. A "drop children" stage. (The current PR still allows you to just ignore attached children while indexing though!)


lucille-core/src/main/java/com/kmwllc/lucille/indexer/ElasticsearchIndexer.java
```diff
@@ -243,24 +250,24 @@ public void closeConnection() {
     }
   }
 
+  /** Only call on documents that have children. */
   private void addChildren(Document doc, Map<String, Object> indexerDoc) {
     List<Document> children = doc.getChildren();
-    if (children == null || children.isEmpty()) {
+
+    if (childDocumentsField == null) {
```
rseitz, line 449: should do this before `List<Document> children = doc.getChildren();`

on JsonDocument, getChildren() does work -- deep copying jackson nodes and creating new lucille docs for each child, so it's best to skip that if it won't be used

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -409,24 +418,24 @@ private Set<Pair<Document, Exception>> uploadDocuments(Map<Pair<String, String>,
     return failedDocs;
   }
 
+  /** Only call on documents that have children. */
   private void addChildren(Document doc, Map<String, Object> indexerDoc) {
     List<Document> children = doc.getChildren();
-    if (children == null || children.isEmpty()) {
+
+    if (childDocumentsField == null) {
```
rseitz, line 424: same comment as in ElasticSearchIndexer -- should do this before `List<Document> children = doc.getChildren();`

rseitz (approved):

Otto-Cline (approved):

PR #524, opened 2026-06-15
LC-893: Skipping Directories in FileConnector

mrwrench24:
This PR allows users to specify a list of URIs, `pathsToSkip`, that they would like skip and not traverse during a `FileConnector` traversal. (Relative / absolute paths to the file system can be used as a convenience as well.)

Accomplishing this is relatively simple in the `LocalStorageClient`. However, for each of the cloud clients, the `traverse` method has been updated quite a bit. Each cloud client now uses appropriate methods in each SDK to actually "traverse" the cloud file system. Previously, we were really just getting flat lists of files, meaning we wouldn't be able to realize the benefits of actually _skipping_ our `pathsToSkip`.


kyle-kmw (approved):

Otto-Cline (approved):
Nice

PR #525, opened 2026-06-15
LC-887: Stay on page when switching doc site version

Otto-Cline:
Changes the sidebar.html file to be aware of the current page, so that when the version is switched the page is maintained. If the page does not exist in the version being switched to, the user will be directed to the landing page for that version (matching the behavior prior to this PR)


mrwrench24 (approved):
LGTM! (but wait for others to review of course)

rseitz (approved):
tried it and it worked nicely

obviously you can only test it after running ./github/scripts/versionworkflow.sh and the launching the site from doc/site-preview

PR #526, opened 2026-06-16
LC-694: let specs handle storageclient validation for local and google

Otto-Cline:
Makes validateOptions() a regular method that gets overridden for S3 and Azure storage clients.


lucille-core/src/test/java/com/kmwllc/lucille/connector/storageclient/StorageClientTest.java
```diff
@@ -45,11 +45,6 @@ public void testCreate() throws Exception {
     assertThrows(IllegalArgumentException.class, () -> StorageClient.create(new URI("https://storagename.blob.core.windows.net/testblob"),
         azCloudOptionsBad));
 
-    // test bad g cloud options
-    Config gCloudOptionsBad = ConfigFactory.load("StorageClientTest/googleBad.conf");
-    assertThrows(IllegalArgumentException.class, () -> StorageClient.create(new URI("gs://bucket/"),
-        gCloudOptionsBad));
```
spencersolomon6, line 51: Can we confirm that the validation still catches this case?

mrwrench24 (changes requested):
Looks good to me but I have a few notes / suggestions:

At some point I would imagine we want to add more flexibility to `Spec`. For example, we can declare "one of" properties. For example, in the `XMLFileHandler`, it would be nice to require in the Spec that exactly one of `xmlIdPath` and `xpathIdPath` must be declared. Perhaps we could do the work for this in this ticket - for now, we would need to do something like "showsUpTogether" (S3) and "oneOf" for (Azure). It could be slightly complex... I can't think (off the top of my head) what the load would be adding this to `Spec`. It would be fairly new functionality. 

On the other hand, I think there is a fair argument to be made that, for now, we can actually just remove `validateOptions` and put the additional validation in the respective constructors. This is done many times throughout the code, so we'd be following that similar pattern.

TLDR: For now I would recommend just removing `validateOptions` altogether and moving "additional" validation into constructors. _Perhaps_ give some brief consideration to adding additional Spec functionality, see how complex it would be, etc.

Otto-Cline:
> Looks good to me but I have a few notes / suggestions:
> 
> At some point I would imagine we want to add more flexibility to `Spec`. For example, we can declare "one of" properties. For example, in the `XMLFileHandler`, it would be nice to require in the Spec that exactly one of `xmlIdPath` and `xpathIdPath` must be declared. Perhaps we could do the work for this in this ticket - for now, we would need to do something like "showsUpTogether" (S3) and "oneOf" for (Azure). It could be slightly complex... I can't think (off the top of my head) what the load would be adding this to `Spec`. It would be fairly new functionality.
> 
> On the other hand, I think there is a fair argument to be made that, for now, we can actually just remove `validateOptions` and put the additional validation in the respective constructors. This is done many times throughout the code, so we'd be following that similar pattern.
> 
> TLDR: For now I would recommend just removing `validateOptions` altogether and moving "additional" validation into constructors. _Perhaps_ give some brief consideration to adding additional Spec functionality, see how complex it would be, etc.

Agreed that this functionality should be added (ticket created at [LC-898](https://kmwllc.atlassian.net/browse/LC-898). I still think validateOptions() should be its own method in this PR since it's only being used in the constructor of BaseStorageClient (there is no work being done in the impl constructors), and the BaseStorageClient doesn't know which options need to be validated. The current implementation allows us to keep the work in BaseStorageClient while handling individual cases.

mrwrench24 (approved):
LGTM as long as we just add a little bit more to the javadoc for `BaseStorageClient.validateOptions` describing its purpose is to validate what is not currently supported by `Spec`

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -60,9 +60,11 @@ public BaseStorageClient(Config config) {
 
   /**
    * Validate that the given config is sufficient to construct an instance of this StorageClient.
+   * Catches cases in which a config option being set requires another option to be set or not set, which is not covered by
```
spencersolomon6: lets say "...requires the presence or absence of another option(s)..."

spencersolomon6 (approved):

PR #527, opened 2026-06-17
LC-272: optimize dropped documents in pipeline

Otto-Cline:
moves checking of dropped and skipped logic to next() method to prevent unnecessary method calls on documents which don't need to be looked at in the pipeline

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
@@ -139,9 +139,6 @@ public abstract class Stage {
    * @return boolean representing - should we process this doc according to its conditionals?
    */
   public boolean shouldProcess(Document doc) {
-    if (doc.isDropped() || doc.isSkipped()) {
-      return false;
-    }
     return condition.test(doc);
   }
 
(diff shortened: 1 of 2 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
```diff
@@ -139,9 +139,6 @@ public void logMetrics() {
    * @return boolean representing - should we process this doc according to its conditionals?
    */
   public boolean shouldProcess(Document doc) {
-    if (doc.isDropped() || doc.isSkipped()) {
```
rseitz, line 142: this needs to stay here for correctness, whether or not we add an additional optimization somewhere else

if someone calls shouldProcess directly, it has to return the right answer, and that right answer depends on whether the doc is dropped or skipped

mrwrench24 (commented):
looks good but I will point out that the `PipelineTest` you provided does pass on `main`. My understanding here is that this additional check is saving us some overhead in `Stage` but is not _newly_ preventing Stages from being run (unnecessarily) on dropped documents. perhaps the solution here is to find a way to assert that these "overhead" methods are not called extra times (e.g. doc.iterator()). maybe a spy.

spencersolomon6 (approved):

mrwrench24 (approved):
LGTM, thank you !

PR #528, opened 2026-06-17
LC-842: Support Arbitrary Kafka Properties

mrwrench24:
This PR allows `kafka` config blocks to contain arbitrary properties. These properties can be placed under `kafka.admin`, `kafka.consumer`, or `kafka.producer`.


lucille-core/src/test/java/com/kmwllc/lucille/core/KafkaTest.java
```diff
@@ -64,6 +64,8 @@ public void testRunInKafkaLocalMode() throws Exception {
     // in "kafka local mode" Workers, Indexers, the Connector/Publisher will run in separate threads,
     // but they will communicate via Kafka (here we use an embedded kafka cluster)
     String brokers = embeddedKafka.getBrokersAsString();
+    // also has arbitrary properties for consumer / producer to ensure no errors
```
mrwrench24, line 67: should be fine, but will want to check against LC-701 / could potentially be conflicting changes

lucille-core/src/main/java/com/kmwllc/lucille/message/KafkaUtils.java
```diff
@@ -135,17 +183,25 @@ public static KafkaConsumer<String, String> createEventConsumer(Config config, S
...
+    } else {
+      producerProps = new Properties();
+
+      producerProps.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, config.getString("kafka.bootstrapServers"));
+
+      if (config.hasPath("kafka.securityProtocol")) {
+        producerProps.put(CommonClientConfigs.SECURITY_PROTOCOL_CONFIG, config.getString("kafka.securityProtocol"));
+      }
+
+      producerProps.put(ProducerConfig.MAX_REQUEST_SIZE_CONFIG, config.getInt("kafka.maxRequestSize"));
```
rseitz, line 199: there are a few related changes coming to this code in the LC-391 PR https://github.com/kmwtechnology/lucille/pull/554/changes#diff-b04fe5894f55beaecbdf7f8643a1e2957782724eb16a5359f86738652a8fcca7

it turned out to be a bug/oversight that we were setting buffer memory equal to max request size, when we really want buffer memory to be greater

also, we want to make sure someone doesn't mistakenly disable kafka's per-partion message ordering guarantees by setting "enable idempotence" to false

suggest that we merge LC-391 first, then resolve conflicts here in LC-842

rseitz (approved):

PR #529, opened 2026-06-17
LC-701: Use Specs for "Catch-All" Validation

mrwrench24:
Last year, IIRC, we had a lot of tickets related to Specs in flight at once. We introduced a LOT of new validation related features in a very short span of time - including making them static(ally accessible), making them use types, etc.

One of the features we wanted to add was validation of the full Lucille config - meaning the "other" blocks that can be found in the root of a Lucille Config: `publisher`, `worker`, `log`, etc. We accomplished this by declaring, in a `Config` itself (confusing), the required and optional properties (untyped) for each of these blocks.

This PR updates this piece of the code to instead use statically accessible Specs in appropriate locations to validate this part of a Lucille config.

As such, the `AnyProperty` has been removed and `withRequired/OptionalProperties` have been removed, as they had no more usages. We can bring them back if needed, of course, but it seems we have not yet stumbled upon a case where a field's type is ambiguous.


lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -59,6 +61,19 @@
  */
 public class Runner {
 
+  public static final Spec SPEC = SpecBuilder.withoutDefaults()
+      .optionalString("metricsLoggingLevel")
+      .optionalNumber("connectorTimeout").build();
+
+  private static final List<Pair<String, Spec>> PARENTS_AND_SPECS = List.of(
```
rseitz: would be ideal to add test coverage for all of these in stage/ConfigValidationTest.testValidationOtherParents() 

runner and zookeeper are already covered there

rseitz: a comment would be helpful, indicating that this is the list of top level properties or "parents" in the config file that get validated... and note to developers that if any new top level properties are added to the config, they should be included in PARENTS_AND_SPECS as well

rseitz (approved):
approving, with a request for more test coverage and a comment on PARENTS_AND_SPECS saying what it is and when it should be updated

lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
```diff
@@ -36,6 +38,10 @@
  */
 public class PublisherImpl implements Publisher {
 
+  public static final Spec SPEC = SpecBuilder.withoutDefaults()
+      .optionalNumber("queueCapacity", "maxPendingDocs").build();
+
+
```
spencersolomon6: [nit] extra new line

spencersolomon6 (approved):

PR #530, opened 2026-06-18
LC-897: CVEs for 0.10.0 release

mrwrench24:


lucille-parent/pom.xml
```diff
@@ -43,30 +43,30 @@
     <maven-source-plugin.version>3.3.0</maven-source-plugin.version>
     <maven-gpg-plugin.version>3.0.1</maven-gpg-plugin.version>
     <maven-assembly-plugin.version>3.3.0</maven-assembly-plugin.version>
-    <jackson.version>2.19.1</jackson.version>
```
mrwrench24, line 46: removed because the Jackson dependencies have started to use different version numbers / systems. For example, `Jackson-annotations` only has `2.20`, `2.21`, etc... so providing it `2.21.2` causes issues. just using the bom appears to be the best / cleanest way to handle this

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -300,7 +299,7 @@ private void deleteByQuery(Map<Pair<String, String>, List<String>> termsToDelete
       }
 
       if (!response.failures().isEmpty()) {
-        for (BulkIndexByScrollFailure failure : response.failures()) {
+        for (BulkByScrollFailure failure : response.failures()) {
```
mrwrench24, line 302: From "UPGRADING.md":

> ```
> ### BulkIndexByScrollFailure renamed to BulkByScrollFailure
> --
> - The `BulkIndexByScrollFailure` class has been renamed to `BulkByScrollFailure`, this affects:
> - `DeleteByQueryResponse`'s `failures` field.
> - `ReindexResponse`'s `failures` field.
> - `UpdateByQueryResponse`'s `failures` field.
> ```

lucille-core/src/main/java/com/kmwllc/lucille/indexer/OpenSearchIndexer.java
```diff
@@ -300,8 +299,13 @@ private void deleteByQuery(Map<Pair<String, String>, List<String>> termsToDelete
       }
 
       if (!response.failures().isEmpty()) {
-        for (BulkIndexByScrollFailure failure : response.failures()) {
-          log.debug("Error while deleting by query: {}, because of: {}", failure.cause().reason(), failure.cause());
+        for (BulkByScrollFailure failure : response.failures()) {
+          if (failure.cause() == null) {
+            log.debug("Error while deleting by query, cause was null");
```
mrwrench24, line 304: however,  the failure's `.cause()` is now annotated as `@Nullable`

lucille-core/pom.xml
```diff
@@ -47,30 +62,25 @@
...
       <groupId>org.apache.kafka</groupId>
       <artifactId>kafka-clients</artifactId>
-      <version>4.0.0</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.lz4</groupId>
-          <artifactId>lz4-java</artifactId>
-        </exclusion>
-      </exclusions>
+      <version>4.1.2</version>
```
kiratraynor: I see 4.3.0 is the newest release of kafka-clients. Does that release have breaking changes or can we upgrade to 4.3.0?

mrwrench24: I was able to upgrade to `4.2.0`. Looks like `spring-kafka` and `spring-kafka-test` do not yet support `4.3.0`.

kiratraynor (approved):

lucille-parent/pom.xml
```diff
@@ -193,7 +193,7 @@
           <redirectTestOutputToFile>false</redirectTestOutputToFile>
           <!--ensure consistent timezone for unit tests-->
           <!--suppress UnresolvedMavenProperty -->
-          <argLine>${argLine} -Dtest.solr.allowed.securerandom=NativePRNG -Duser.timezone=UTC</argLine>
+          <argLine>${argLine} -Dtest.solr.allowed.securerandom=NativePRNG -Dtest.solr.allowed.securerandom=NativePRNG -Duser.timezone=UTC</argLine>
```
kiratraynor: I'm not super familiar with what should be here, but I'm thinking `-Dtest.solr.allowed.securerandom=NativePRNG` should not be here twice?

mrwrench24: you are very right. this is a symptom of me having gone a little crazy addressing some of the dependency issues earlier today...

rseitz (approved):

kiratraynor (approved):

PR #533, opened 2026-06-24
LC-903: Allow use of ForkedParser in TextExtractor 

mrwrench24:
When there are exceptionally large documents running through `TextExtractor`, they can cause an OOM error, causing the entire JVM to crash. 

When the `ForkedParser` is used, text extraction runs in a separate process, ensuring OOM errors are isolated from the main application. This does add a bit of overhead to every document, so using the `ForkedParser` is optional.


lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -171,8 +213,8 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
 
       try (InputStream contentStream = fileFetcher.getInputStream(filePath)) {
         parseInputStream(doc, contentStream);
-      } catch (Exception e) {
-        log.warn("Error with InputStream for file path.", e);
+      } catch (IOException e) {
+        log.warn("Error processing file {}", filePath, e);
```
mrwrench24, line 229: changed to allow stage exception to propagate through

kyle-kmw:
I think that the ForkParser will go away if/when we upgrade to Tika 4.x. Is there a way to do this directly with tika-pipes now or do we just need to plan for migrating later?

https://cwiki.apache.org/confluence/spaces/TIKA/pages/404160772/4.x+Migration+Guide

mrwrench24:
I think that the easiest path would be to use the `ForkParser` now and then upgrade to the `ParseForkedParser` as part of a future upgrade to `4.x` (which has not been released). 

My general understanding (and from some experimental changes I have locally) of using pipes directly is it adds a bit of overhead / boilerplate to the code. Based on the [Tika Examples](https://github.com/apache/tika/blob/main/tika-example/src/main/java/org/apache/tika/example/PipesForkParserExample.java) it looks like our implementation using the `ForkParser` now would be a relatively (?) simple migration to the `PipesForkParser`. Using pipes directly now and then migrating to the `PipesForkParser` would not be so straightforward.

let me know what you think...

lucille-plugins/lucille-tika/src/test/java/com/kmwllc/lucille/tika/stage/TextExtractorTest.java
```diff
@@ -436,4 +436,16 @@ public void testTimeout() throws Exception {
     // Document should have text after processing without interruption.
     assertEquals("Document should have text.", "Hi There!\n", doc2.getString("text"));
   }
+
+  @Test
+  public void testForkingParser() throws Exception {
```
kyle-kmw, line 448: Is it possible to have a test that runs processDocument on the stage with two documents, the first with a problematic file (or simulates a problematic input stream) that crashes the forked process and the second with a normal text file that demonstrates that the stage survives the problematic file and processes the second document successfully?

mrwrench24: yes... took me a little longer than I had hoped to get a creative solution, but just added this!

mrwrench24:
@kyle-kmw
Sanity check after our brief discussion yesterday. My understanding is the forked parsers should be closed appropriately after a timeout

kyle-kmw (approved):

PR #535, opened 2026-06-26
LC-897 (Pt. 2): Bump Solr to 10.0.0

mrwrench24:
Upgrades Solr libraries used in Lucille to 10.0.0.


lucille-core/pom.xml
```diff
@@ -224,6 +224,289 @@
...
+      <version>6.0.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>jakarta.ws.rs</groupId>
+      <artifactId>jakarta.ws.rs-api</artifactId>
+      <version>3.1.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
```
rseitz: caught by an AI-assisted review -- this one dependency is missing `<scope>test</scope>` -- can't say I would have found this on my own :)

lucille-core/pom.xml
```diff
@@ -530,6 +813,15 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <excludes>
+            <exclude>**/SolrIndexerIntegrationTest.java</exclude>
```
rseitz: I think if we renamed the test to end with IT -- as in SolrIndexerIT -- it will get picked up automatically by failsafe, not surefire, and we wouldn't need this exclude or the later surefire include

looks like the codebase is currently inconsistent in naming integration tests: we have FileToFileIngestIT but also APIIntegrationTest etc.

my understanding is we want to make sure this solr indexer integration test runs in failsafe because we want to be able to set tests.src.home?

I'd suggest going with SolrIndexerIT if it works and moving in the direction of more consistent naming for integration tests (but let's wait to fix the others for a separate PR, to keep this one small)

rseitz (approved):
approving but with 2 significant comments

lucille-core/pom.xml
mrwrench24: It is missing that scope but that is because it is needed by `RunDetails`. It appears we were just taking advantage of the fact that Solr was pulling it in for us before. Adding test scope or just removing the dependency causes compilation to fail in `RunDetails`.

I moved the dependency & added a comment making it clear why it does not have test scope.

lucille-core/src/main/java/com/kmwllc/lucille/util/SolrUtils.java
```diff
@@ -71,23 +71,24 @@ private static CloudHttp2SolrClient getCloudClient(Config config) {
...
+    HttpJdkSolrClient httpClient = getHttpClientAndSetCheckPeerName(config);
+    cloudBuilder.withHttpClient(httpClient);
+
+    // CloudHttp2SolrClient does not close a caller-supplied httpClient, so wrap it to prevent leaks.
+    return new CloudHttp2SolrClient(cloudBuilder) {
+      @Override
+      public void close() {
+        super.close();
+        try {
           httpClient.close();
```
kyle-kmw: check for NULL httpClient before closing

lucille-core/pom.xml
```diff
@@ -530,6 +816,10 @@
           </archive>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
```
kiratraynor: Just curious why this is needed here? I think this would be inherited from the parent pom.

mrwrench24: removed, thank you for pointing this out !

kiratraynor (approved):

kyle-kmw (approved):

lucille-examples/lucille-distributed-example/kafka/server.properties
```diff
@@ -13,34 +13,36 @@
...
-#
-# This configuration file is intended for use in ZK-based mode, where Apache ZooKeeper is required.
-# See kafka.server.KafkaConfig for additional details and defaults
-#
-
 ############################# Server Basics #############################
 
-# The id of the broker. This must be set to a unique integer for each broker.
-broker.id=0
+# The role of this server. Setting this puts us in KRaft mode
```
rseitz, line 18: kafka/Dockerfile was passing the path of server.properties to the kafka server during launch
now that we don't have our own Dockerfile for kafka, how does the server know to use this particular server.properties
are we sure this file is still being used?

lucille-examples/lucille-distributed-example/solr/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM ubuntu
 RUN apt-get upgrade
 RUN apt-get update && apt-get install curl -y
 RUN apt-get install lsof bc openjdk-21-jre-headless -y --fix-missing
-RUN curl -L https://www.apache.org/dyn/closer.lua/solr/solr/9.5.0/solr-9.5.0.tgz?action=download --output solr.tgz
+RUN curl -L https://www.apache.org/dyn/closer.lua/solr/solr/10.0.0/solr-10.0.0.tgz?action=download --output solr.tgz
```
rseitz: instead of downloading solr into an ubuntu image, can we use the solr:10.0.0 official docker image?

lucille-examples/lucille-distributed-example/kafka/server.properties
mrwrench24: It is now being set (mounted) in the Docker compose. The kafka image defaults to [checking /mnt/shared/config](https://github.com/apache/kafka/blob/trunk/docker/examples/README.md#:~:text=The%20Command%20docker%20run%20%2D%2Dvolume%20/path/to/property/folder:/mnt/shared/config%20%2Dp%209092:9092%20apache/kafka:latest%20can%20be%20used%20to%20mount%20the%20folder%20containing%20the%20property%20files.).

mrwrench24:
Here is my theory / understanding of why the `SolrIndexerIntegrationTest` is intermittently failing.

I believe this issue has popped up since we merged the first LC-897 PR (addressing general CVEs). It appears that the test fails due to a timeout error when "Randomized ssl" is true. If you search for `SolrIndexerIntegrationTest` (or, in this PR, `SolrIndexerIT`), this line should show up right after the first mention of the test:

`2026-06-23T17:37:30.0638119Z 26/06/23 17:37:30   INFO SolrTestCaseJ4: Randomized ssl (false) and clientAuth (false) via: @org.apache.solr.util.RandomizeSSL(reason="", value=0.0/0.0, ssl=0.0/0.0, clientAuth=0.0/0.0)`

This morning I looked at about 6 old CI runs - before we merged the first CVE PR - to check this setting. I wanted to be sure that, for whatever reason, we did not "break" support for SSL recently. Interestingly, in each of these runs, randomized ssl was always false. The second part, `clientAuth`, was sometimes true.

Since the CVE upgrades were merged, it seems, randomized ssl is sometimes true. It appears we are not equipped or configured for this, and the test ultimately times out.

PR #543, opened 2026-07-15
LC-874: Disabling Stages (& GetLegalProperties Refactor)

mrwrench24:
Allows users of any Stage to set the `enabled` property to `false` in the Config. In doing so, the Stage is not constructed, `start()` and `stop()` are not called, and documents are not processed by the `Stage`. 

Due to how Stage is used, the onus is on users (namely, `Pipeline`) to check this property and handle disabled stages as appropriate. In fact, the `Stage` class itself has no knowledge of the property (besides declaring the default legal property in `SpecBuilder`). 

Disabled stages are still validated in both validation mode and when kicking off a regular run. I could see going either way on this. It would be a straightforward change to _not_ validate Stages that are disabled, if desired!

Since we are changing one of the Stage default properties, this was also a good opportunity to revisit `GetLegalProperties`. It was renamed to `getUniqueProperties`, and only returns the properties _unique_ to a Stage. (It no longer returns properties like name and class.) So, those unit tests have been updated and renamed appropriately as well.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
@@ -360,7 +360,7 @@ public abstract class Stage {
    * @return The unique property names for this Stage declared in its Spec. Does not return the default properties
    * for a Stage (such as "name" and "conditions").
    */
-  public Set<String> getUniqueProperties() {
+  public Set<String> getNonDefaultLegalProperties() {
     Set<String> propNames = getSpec().getLegalProperties();
 
     Set<String> defaultPropNames = DEFAULT_LEGAL_PROPERTIES.stream().map(Property::getName).collect(Collectors.toSet());
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
@@ -4,8 +4,13 @@ import com.codahale.metrics.Counter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.SharedMetricRegistries;
 import com.codahale.metrics.Timer;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.kmwllc.lucille.core.spec.BooleanProperty;
+import com.kmwllc.lucille.core.spec.ListProperty;
+import com.kmwllc.lucille.core.spec.Property;
 import com.kmwllc.lucille.core.spec.Spec;
 import com.kmwllc.lucille.core.spec.SpecBuilder;
+import com.kmwllc.lucille.core.spec.StringProperty;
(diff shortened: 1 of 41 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
```diff
@@ -332,8 +356,17 @@ public Config getConfig() {
     return config;
   }
 
-  public Set<String> getLegalProperties() {
-    return getSpec().getLegalProperties();
+  /**
+   * @return The unique property names for this Stage declared in its Spec. Does not return the default properties
+   * for a Stage (such as "name" and "conditions").
+   */
+  public Set<String> getUniqueProperties() {
```
rseitz: very minor but I'm getting slightly thrown off by "unique" -- these aren't necessarily unique in that other stages might have the same properties -- I would suggest getNonDefaultProperties()... or to be truly pedantic, getNonDefaultLegalProperties()

lucille-core/src/test/java/com/kmwllc/lucille/core/PipelineTest.java
```diff
@@ -261,6 +261,37 @@ public void testDroppedDocumentSkipsDownstreamStages() throws Exception {
...
+    assertEquals("TimestampEnabled", pipeline.getStages().get(1).getName());
+  }
+
+  @Test
+  public void testAllDisabledStages() throws Exception {
+    Config pipelineConfig = ConfigFactory.load("PipelineTest/all-disabled.conf");
+    Pipeline pipeline = Pipeline.fromConfig(pipelineConfig, "my-pipeline", "");
+
+    // two of the stages have invalid config (missing "destField"). Note that no validation
+    // errors are thrown!
```
rseitz: comment could be a bit clearer...

_why_ are no validation errors thrown? it's because we never instantiated the disabled stages, so the Stage constructor wasn't invoked, so the disabled stages never attempted to validate themselves via     getSpec().validate(config, getDisplayName()) in the constructor

however... if we had called Pipeline.validateStages(), there would have been validation errors

doc update in this PR correctly says "The Stage will still be validated against its Spec, warning you of any missing, invalid, or unknown properties in the Config." but the point is that only happens in validation mode

lucille-core/src/main/java/com/kmwllc/lucille/core/Stage.java
rseitz: and if we go with getNonDefaultLegalProperties() this raises the question of what the stage test methods should be named. I'd sort of hate to have a testGetNonDefaultLegalProperties() in lots of places so what about just testSpec() ?

rseitz (approved):
approving with requests to:

1) rename
getUniqueProperties() -> getNonDefaultLegalProperties()
testGetUniqueProperties() -> testSpec()

2) update a test comment

spencersolomon6 (approved):

PR #544, opened 2026-07-23
LC-514: Replace -usekafka and -local with -distributed and -external

LunkadV:
Replaced -usekafka and -local flags with -distributed and -distributedsandbox respectively with enforced mutual exclusivity. 

-distributed: Used exactly how -usekafka was earlier.

-distributedsandbox: Results in a local kafka run, mutually exclusive from -distributed. 

No flag: default to fully local mode (unchanged from previous default).

Additional validation for flags added. An argument not matching any options or one entered without leading dashes will print a help message and exit the system.

Testing: All tests pass, including the containerized nightly smoke test.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
@@ -20,6 +20,7 @@ import org.apache.commons.lang3.time.StopWatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.util.*;
@@ -68,8 +69,8 @@ public class Runner {
   public enum RunType {
     LOCAL, // launch Worker(s) and Indexer as threads; have all components communicate via in-memory queues
     TEST, // same as LOCAL, but bypass Solr, and store message traffic so it can be inspected after the run
-    KAFKA_LOCAL, // launch Worker(s) and Indexer as threads; have all components communicate via Kafka
-    KAFKA_DISTRIBUTED // assume Workers/Indexers were started separately (don't launch threads); have all components communicate via Kafka
+    DISTRIBUTED, // launch Worker(s) and Indexer as threads; have all components communicate via Kafka
+    DISTRIBUTED_SANDBOX // assume Workers/Indexers were started separately (don't launch threads); have all components communicate via Kafka
   }
 
   // no need to instantiate Runner; all methods currently static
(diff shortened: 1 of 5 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -68,8 +69,8 @@ public class Runner {
   public enum RunType {
     LOCAL, // launch Worker(s) and Indexer as threads; have all components communicate via in-memory queues
     TEST, // same as LOCAL, but bypass Solr, and store message traffic so it can be inspected after the run
-    KAFKA_LOCAL, // launch Worker(s) and Indexer as threads; have all components communicate via Kafka
-    KAFKA_DISTRIBUTED // assume Workers/Indexers were started separately (don't launch threads); have all components communicate via Kafka
+    DISTRIBUTED, // launch Worker(s) and Indexer as threads; have all components communicate via Kafka
```
rseitz: DISTRIBUTED doesn't launch Worker(s) and Indexers as threads -- that's the mode called DISTRIBUTED_SANDBOX here
looks like the descriptions are swapped

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -155,10 +156,10 @@ public void close() {
    * no args: pipelines workers and indexers will be executed in separate threads within the same JVM; communication
    * between components will take place in memory and Kafka will not be used
    * <p>
-   * -useKafka: connectors will be run, sending documents and receiving events via Kafka. Pipeline workers
+   * -distributed: connectors will be run, sending documents and receiving events via Kafka. Pipeline workers
    * and indexers will not be run. The assumption is that these have been deployed as separate processes.
    * <p>
-   * -local: modifies -useKafka so that workers and indexers are started as separate threads within the same JVM;
+   * -distributedSandbox: modified distributed where workers and indexers are started as separate threads within the same JVM;
```
rseitz: conceptually it's not really a modified distributed mode -- it's more of a modified local mode where the modification is that inter-component communication goes through Kafka

I'm going to suggest we rename "distributed sandbox" as "external"

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -827,4 +829,24 @@ private static Slf4jReporter.LoggingLevel getMetricsLoggingLevel(Config config)
...
+    try (StringWriter writer = new StringWriter();
+        PrintWriter printer = new PrintWriter(writer)) {
+      String header = "Run a sequence of connectors";
+      new HelpFormatter().printHelp(printer, 256, "Runner", header, cliOptions,
+          2, 10, "", true);
+      log.info(writer.toString());
+    } catch (IOException e) {
+      log.debug("Unexpected IOException", e);
+    }
+    SystemHelper.exit(1);
```
rseitz: it looks like we're only calling printHelpAndExit in an error scenario, so it's correct to exit(1) here; but it's common for CLIs to also print help info when requested, which is not an error scenario and should use exit(0) instead. Since we don't currently have a help option and aren't adding one in this PR, you don't have to change this method to take the exit code as a parameter, but the Javadoc should make it clear that this method exits with code 1 and should only be used when there's an error

lucille-core/src/main/java/com/kmwllc/lucille/core/Runner.java
```diff
@@ -183,16 +191,12 @@ public static void main(String[] args) throws Exception {
...
-        String header = "Run a sequence of connectors";
-        new HelpFormatter().printHelp(printer, 256, "Runner", header, cliOptions,
-            2, 10, "", true);
-        log.info(writer.toString());
+      if (!cli.getArgList().isEmpty()) {
+        printHelpAndExit(cliOptions, "Unrecognized argument(s): " + cli.getArgList());
       }
-      SystemHelper.exit(1);
+    } catch (UnrecognizedOptionException | MissingOptionException | AlreadySelectedException e) {
+      printHelpAndExit(cliOptions, e.getMessage());
```
spencersolomon6: Remove exit from this method and add it here instead

rseitz (approved):

PR #545, opened 2026-07-28
LC-915: CVEs for impending 0.11.0 release

mrwrench24:
Upgrades were relatively straightforward so far. The only "major" upgrade was for Elasticsearch (8.18.4 --> 9.0.0), which is a **breaking change** + required code changes


kiratraynor (approved):
Approving, we can see if elastic releases a non-vulnerable version of 8.19.x before we release but I think it's fine.

PR #546, opened 2026-07-29
LC-913 enabling hit highlighting for lucille docs

LunkadV:
Summary
- Search result excerpts now center around the matched search term instead of a static excerpt for each page. 
- Matched term is highlighted in the search results and clicking on a result navigates to the first instance of the match term on the destination page with every instance highlighted.

Templated Files
- Mark.min.js : Pulled from https://github.com/julkue/mark.js/blob/master/dist/mark.min.js and unchanged. 
- offline-search.js : Pulled from https://github.com/google/docsy/blob/v0.14.0/assets/js/offline-search.js in order to override search result display and click behavior.


mrwrench24 (changes requested):
It seems like we need to tighten up certain searches. It looks like searches with 1-3 characters can return some odd results - at least, within the previews we see after hitting "enter".

For example, if I look up "aws", I can see text like "0.", "1", or "(21)" being highlighted. It does seem like the links, when clicked, _do_ highlight "AWS" properly - and these previews do have some kind of "relevant" results. 

one potential offender is wildcard/edit distance being applied for queries that are too short...

LunkadV:
> It seems like we need to tighten up certain searches. It looks like searches with 1-3 characters can return some odd results - at least, within the previews we see after hitting "enter".
> 
> For example, if I look up "aws", I can see text like "0.", "1", or "(21)" being highlighted. It does seem like the links, when clicked, _do_ highlight "AWS" properly - and these previews do have some kind of "relevant" results.
> 
> one potential offender is wildcard/edit distance being applied for queries that are too short...

You're right, besides an exact match Docsy queries once with a leading and trailing wildcard and once with an edit distance of 2. On top of this it seems AWS gets stemmed to aw which is why we're getting these odd results.

spencersolomon6:
@LunkadV Would it be possible to highlight the words if they appear in the right sidebar menu as well?

mrwrench24 (approved):
LGTM ! thanks for fixing!

doc/site/layouts/partials/hooks/body-end.html
```diff
@@ -0,0 +1,49 @@
...
+
+{{ if .Site.Params.offlineSearch -}}
+<style>
+  ::highlight(search-match) { background-color: yellow; }
+</style>
+<script>
+  (function () {
+    var q = new URLSearchParams(window.location.search).get('q');
+    var content = document.querySelector('.td-content');
+    var words = q.toLowerCase().split(/\s+/).filter(Boolean);
```
rseitz: this line references q so it should be moved after the `if (!q.... ) return;`

spencersolomon6 (approved):

rseitz (approved):

PR #547, opened 2026-07-29
LC-514 Validation Tests added for Runner flags

LunkadV:
Added tests to verify Runner CLI Flag validation that was added in https://github.com/kmwtechnology/lucille/pull/544

### DIFF
--- a/lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
+++ b/lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
@@ -845,7 +845,7 @@ public class RunnerTest {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/core/RunnerTest.java
```diff
@@ -893,6 +893,43 @@ public void testValidateAndRenderCliFlags() throws Exception {
     });
```
mrwrench24, line 893: unrelated to your PR, but the comments for `testRunTypeCliFlags()` (line ~848) mention the old flag names. while we are at it, we may as well update, just in case :)

mrwrench24 (approved):
LGTM just one _very_ small ask

spencersolomon6 (approved):

PR #549, opened 2026-07-31
LC-909: Transient Tika Test Failure (Again)

mrwrench24:
In the previous PR, we were looking at the wrong `Thread.sleep()` call to identify a race condition.

This PR increases the sleep in the `InterruptTrackingParser` substantially to prevent the parse from completing before the test is ready to check for an interrupt. We do not wait for this entire duration - because it gets _interrupted_.


mrwrench24:
I am reasonably confident this *may* be fixed. I ran CI on GH Actions seven times this morning with an updated change (after a failure on Friday afternoon) and it passed each time

kyle-kmw (approved):

PR #550, opened 2026-08-05
LC-916: Support ability to pass info to metadata field (Tika module)

kyle-kmw:
## LC-916: Support ability to pass info to metadata field (Tika module)

---

## Why

The `TextExtractor` stage handed Tika a plain `InputStream` with no way to supply per-document hints. That left two gaps in content-type detection:

1. **No detection hints.** Tika detects types more accurately when given the file name (`resourceName`) and/or an advertised `Content-Type` — e.g. distinguishing a CSV from generic text, or picking the right parser when magic bytes are ambiguous. There was no way to feed values the pipeline already has (a filename field, a source-provided content type) into the parser.
2. **Mime-Magic only.** Because a plain `InputStream` was passed, Tika's container-aware detectors (which need to spool content to a file) were skipped, so detection fell back to Mime-Magic on the leading bytes alone.

See [Tika content type detection](https://tika.apache.org/3.2.3/detection.html).

## What

- **New `metadataFields` config option** (`Map`) — maps a Tika metadata key to the name of a document field whose value(s) are copied into the `Metadata` handed to Tika *before* parsing. Absent fields are skipped; multi-valued fields contribute all values. Primary use case is `resourceName` / `Content-Type` detection hints.
- **Wrap content in `TikaInputStream`** in `processDocument` (both the byte-array and file-path paths) so container-aware detectors can run. Closing the wrapper closes the underlying stream, so it's the only resource managed.
- **Bug fix:** `metadataPrefix == ""` used reference equality (always false), making the "no prefix" path unreachable — changed to `.isEmpty()`.
- **Docs:** expanded the class-level Javadoc explaining the detection use case and the `tikaKey → documentFieldName` mapping direction.
- Added `log/*` to the plugin `.gitignore` (stray test-run output).

## Testing

- `mvn -pl lucille-plugins/lucille-tika -am -Dtest=TextExtractorTest test` — **27/27 pass**, including the existing detection/whitelist/timeout and fork/OOM regression tests.
- New tests:
  - `metadataFields` wiring: values wired in, multi-valued, missing field skipped, unconfigured → empty, and an end-to-end check that the injected `resourceName` round-trips through the parser.
  - `TikaInputStream` wrapping: `spy` + `ArgumentCaptor` assert the stream reaching the parser is a `TikaInputStream` for both the file-path and byte-array paths.
- Note: on the non-fork path `AutoDetectParser` already wraps internally for detection, so wrapping is a no-op there; the tangible benefit is on the `ForkParser` path and for direct callers. Tests assert the wrapping contract, and existing detection tests guard against regressions.

---


lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -62,6 +64,28 @@
  * s3 (Map, Optional) : If your dictionary files are held in S3. See FileConnector for the appropriate arguments to provide.
  * azure (Map, Optional) : If your dictionary files are held in Azure. See FileConnector for the appropriate arguments to provide.
  * gcp (Map, Optional) : If your dictionary files are held in Google Cloud. See FileConnector for the appropriate arguments to provide.
+ * metadataFields (Map, Optional) : Maps Tika metadata keys to the names of document fields whose values should be
+ * copied into the {@link Metadata} handed to Tika <i>before</i> parsing. This is the input to Tika, and is distinct
+ * from the metadata Tika produces <i>after</i> parsing (which is written back to the document under metadataPrefix).
```
mrwrench24: The metadata fields will get written back to the Lucille documents. I don't think this is an issue but is something we should point out, and the comment *may* be interpreted as saying "this won't happen".

This test passes:
```java
 @Test
  public void testResourceNameHintLeaksIntoOutput() throws Exception {
    TextExtractor stage = (TextExtractor) factory.get("TextExtractorTest/metadatafields.conf");
    stage.start();

    Document doc = Document.create("doc1");
    doc.setField("byte_array", Files.readAllBytes(
        new File("src/test/resources/TextExtractorTest/tika.txt").toPath()));
    doc.setField("filename", "myfile.csv");   // pure INPUT hint

    // not on the doc yet
    assertFalse(doc.has("tika_resourcename"));

    stage.processDocument(doc);

    assertEquals("myfile.csv", doc.getString("tika_resourcename"));

    stage.stop();
  }
```

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -289,6 +354,14 @@ public void parseInputStream(Document doc, InputStream inputStream) throws Stage
     }
   }
 
+  /**
+   * Parses given input stream, close it, and adds the text data and metadata to given document
+   */
+  public void parseInputStream(Document doc, InputStream inputStream) throws StageException {
```
mrwrench24: Looks like this doesn't have any uses?

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -223,15 +255,45 @@ public Iterator<Document> processDocument(Document doc) throws StageException {
       // get fileObject from path
       String filePath = doc.getString(filePathField);
 
-      try (InputStream contentStream = fileFetcher.getInputStream(filePath)) {
-        parseInputStream(doc, contentStream);
+      // wrap in a TikaInputStream so container-aware detectors (which need to spool the content to a file) can run;
+      // a plain InputStream would limit detection to Mime-Magic on the leading bytes. Closing the TikaInputStream
+      // also closes the underlying content stream, so it is the only resource we need to manage here.
+      try (TikaInputStream tikaStream = TikaInputStream.get(fileFetcher.getInputStream(filePath))) {
```
mrwrench24: extremely minuscule, but we may as well separate out the two input stream related calls to prevent (small) risk of a leak

mrwrench24 (changes requested):

mrwrench24 (approved):
LGTM when test failure is resolved! thx!

PR #552, opened 2026-08-06
LC-921: CVEs for 0.11.0 (Pt. 3)

mrwrench24:
Bumps python language and removes unnecessary bouncy castle dependency / pin. Also small bump of JSoup + small test change for low CVE. Test is changed to reflect how JSoup returns lists now.


LunkadV (approved):
LGTM

wfingersoll-kmwllc (approved):

PR #553, opened 2026-08-17
LC-831 NPE if auth type is not set in config for api

LunkadV:
Problem: If AuthType was not assigned in the config and Auth was enabled, .equals() would be called on a null string and crash with an NPE. 

Fix: 
- Null Check AuthType before use
- Unsupported AuthType now errors 
- New tests to confirm behavior

### DIFF
--- a/lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
+++ b/lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
@@ -91,6 +91,7 @@ public class APIApplication extends Application<LucilleAPIConfiguration> {
    * Starts the Lucille API server, configures authentication, and registers resources. Throws Exception on unsupported auth type, aborting startup.
    * @param config the Lucille API configuration
    * @param env the Dropwizard environment
+   * @throws IllegalArgumentException if authentication is enabled but no auth type is set
    * @throws Exception if authentication type is unsupported (startup abort) or other startup failures
    */
   @Override
@@ -104,7 +105,13 @@ public class APIApplication extends Application<LucilleAPIConfiguration> {
 
     // Enable Basic Auth only if it's enabled in the configuration
     if (authEnabled) {
-      if (config.getAuthConfig().getType().equals(AuthType.BASIC_AUTH)) {
+      AuthType authType = config.getAuthConfig().getType();
+
+      if (authType == null) {
+        throw new IllegalArgumentException("auth.type must be set when auth.enabled is true.");
+      }
+
+      if (AuthType.BASIC_AUTH.equals(authType)) {
         env.jersey()
             .register(new AuthDynamicFeature(new BasicCredentialAuthFilter.Builder<PrincipalImpl>()
                 .setAuthenticator(new BasicAuthenticator(config.getAuthConfig().getPassword()))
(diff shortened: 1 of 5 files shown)


lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/config/AuthConfiguration.java
```diff
@@ -61,11 +61,7 @@ public AuthType getType() {
    */
   @JsonProperty
   public void setType(String type) {
-    if (type.equals("basicAuth")) {
-      this.type = AuthType.BASIC_AUTH;
-    } else {
-      this.type = AuthType.NO_AUTH;
-    }
+    this.type = "basicAuth".equals(type) ? AuthType.BASIC_AUTH : AuthType.NO_AUTH;
```
rseitz: as discussed, throw an exception if the provided type is not recognized

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
```diff
@@ -104,15 +105,21 @@ public void run(LucilleAPIConfiguration config, Environment env) throws Exceptio
 
     // Enable Basic Auth only if it's enabled in the configuration
     if (authEnabled) {
-      if (config.getAuthConfig().getType().equals(AuthType.BASIC_AUTH)) {
+      AuthType authType = config.getAuthConfig().getType();
+
+      if (authType == null) {
+        throw new IllegalArgumentException("auth.type must be set when auth.enabled is true.");
+      }
```
spencersolomon6, line 112: Remove exception here which is no longer needed

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
```diff
@@ -104,15 +105,21 @@ public void run(LucilleAPIConfiguration config, Environment env) throws Exceptio
...
+      if (AuthType.BASIC_AUTH.equals(authType)) {
         env.jersey()
             .register(new AuthDynamicFeature(new BasicCredentialAuthFilter.Builder<PrincipalImpl>()
                 .setAuthenticator(new BasicAuthenticator(config.getAuthConfig().getPassword()))
                 .buildAuthFilter()));
         env.jersey().register(new AuthValueFactoryProvider.Binder<>(PrincipalImpl.class));
         log.info("Basic authentication has been enabled.");
       } else {
-        throw new Exception("Unsupported auth type configured for the Lucille Admin API.");
+        throw new Exception("Unsupported auth type configured for the Lucille Admin API: " + authType);
```
spencersolomon6: What if this is set to NO_AUTH?

LunkadV: If authEnabled is true, AuthType should not be NO_AUTH and will error before this line at 110

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
LunkadV: Changed to error when AuthType is NO_AUTH instead of when it is null

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/APIApplication.java
spencersolomon6: So as it stands right now, if you set authEnabled: true, you HAVE to set authTYPE to "basicAuth" exactly or you get an error?

And you set authEnabled:false, you do not have to set anything for authType?

LunkadV: yes, we can choose to handle case differences if needed. 

As it is currently, the AuthType field is redundant but that would change if more AuthTypes are added in the future

spencersolomon6 (approved):

rseitz (approved):

PR #554, opened 2026-08-17
LC-391: update uses of KafkaProducer to improve throughput

rseitz:
This PR updates how Lucille sends messages to Kafka in the Publisher, Worker, and Indexer.

Previously, all uses of `kafkaProducer.send()` were followed by a chained `.get()` — making each send synchronous, requiring a full round trip to the Kafka broker for acknowledgement — and then followed by a `kafkaProducer.flush()` which was redundant. The `.get()` call already blocks until the broker acknowledges the record, leaving nothing in the producer's buffer for `flush()` to act on.

### Changes

**1. Remove redundant `flush()` calls**

All 7 `flush()` calls that followed `send().get()` were removed across `KafkaPublisherMessenger`, `KafkaWorkerMessenger`, `KafkaIndexerMessenger`, `HybridWorkerMessenger`, and `HybridIndexerMessenger`. These were effectively no-ops — the broker had already acknowledged the record — so removing them has no behavioral impact.

**2. Make selected `send()` calls asynchronous**

The remaining `send().get()` sites were evaluated for whether removing the `.get()` and using an asynchronous send would improve throughput without sacrificing correctness. Several were left as synchronous:

- **KafkaWorkerMessenger.sendForIndexing() and sendEvent()** — sendForIndexing
must confirm delivery before the source offset is committed; otherwise a crash could lose the document. Since sendForIndexing must remain synchronous, making the adjacent sendEvent calls async in the same iteration would save at most one additional Kafka RTT (round trip time) per document — probably not enough to justify the added error-handling complexity.
- **KafkaWorkerMessenger.sendFailed()** — A rare error path (retry count exceeded). One send per exhausted-retry document; async provides no meaningful benefit.
- **KafkaIndexerMessenger.sendEvent() (single-event method)** — Still used for the failure path in the Indexer, which handles a small number of documents per batch. Left synchronous for simplicity and per-document error attribution.

Two primary updates were made:

- **IndexerMessenger.sendEvents() (new batch-aware method)** — After indexing a batch of N documents, the Indexer previously sent N individual synchronous Kafka events back-to-back in a tight loop. Each blocked for one broker round trip. The new `sendEvents(List<Document>, String, Event.Type)` method fires all N sends asynchronously into the producer's buffer, then calls `kafkaProducer.flush()` once at the end. This reduces the event-send phase from O(N × RTT) to O(1 × RTT). For a 200-document batch with 2ms broker latency, this saves ~400ms per batch cycle. Errors are captured via an `AtomicReference` callback and thrown after the flush.

- **KafkaPublisherMessenger.sendForProcessing()** — Replaced `send().get()` with an async send using a callback. Failures are captured and surfaced on the next call to `sendForProcessing()`, preserving fail-fast behavior. A new `flush()` method on `PublisherMessenger` (called from `PublisherImpl.flush()` at the end of connector execution) ensures all buffered sends are confirmed before the connector thread declares itself done. This prevents `waitForCompletion()` from hanging if the final send fails. The throughput benefit here is that the Kafka round trip now overlaps with the connector's work producing the next document, rather than blocking the connector thread.

**3. Consistent error handling across connectors**

With asynchronous sends, a publishing failure now surfaces as an exception on the *next* call to `publish()` rather than on the failing call itself. Connectors that previously swallowed publishing exceptions would silently continue producing documents indefinitely without discovering that Kafka is unreachable.

Three connectors were updated to propagate publishing exceptions rather than catching and continuing:

- **ChunkingXMLHandler.internalPublishDocument()** — now throws `SAXException`
- **FileConnector (tombstone loop)** — now throws `ConnectorException`
- **BaseFileHandler.processFileAndPublish()** — now throws `FileHandlerException`

This adopts a consistent stance: Kafka publishing failures are unrecoverable and abort the run. Retrying or working around the error would add significant complexity for minimal benefit — if the broker is unreachable, continuing to produce documents is pointless.

In `BaseFileHandler`, the loop now separates iterator failures (e.g. a corrupt CSV row) from publishing failures. Iterator exceptions are logged and skipped (per-record parse errors are recoverable); publishing exceptions are thrown (Kafka errors are not).


lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
```diff
@@ -441,25 +441,20 @@ private void sendToIndexWithAccounting(List<Document> batchedDocs) {
...
-        } catch (Exception e) {
-          docLogger.error("Error sending completion event for doc {}. RUN WILL HANG.", d.getId(), e);
-        } finally {
-          if (d.getRunId() != null) {
-            MDC.popByKey(RUNID_FIELD);
+      try {
+        messenger.sendEvents(succeededDocs, "SUCCEEDED", Event.Type.FINISH);
+        for (Document d : succeededDocs) {
+          try (MDCCloseable docIdMDC = MDC.putCloseable(ID_FIELD, d.getId())) {
+            docLogger.info("Sent success message for doc {}.", d.getId());
```
mrwrench24, line 452: It is a small detail, but I wonder whether this `docLogger` call is still worthwhile or informative... I suppose we are still tracking the lifecycle of every document, but figured I would maybe start a conversation about it

mrwrench24: we will have a large batch of docs / doc logger statements with roughly the same timestamp

mrwrench24 (approved):
looks good to me - going to sanity run the distributed example. just one VERY small discussion point about a docLogger statement, whether it is still worthwhile

mrwrench24:
Hmmmm... I am seeing it fail locally. There are 8 output docs, not the (expected) 6. Investigating ...

mrwrench24:
okay... seems I just had to do some resetting locally. passing now

lucille-core/src/main/java/com/kmwllc/lucille/core/Indexer.java
rseitz: thanks for the comment. I think we _do_ want to keep it. Agree, it'll seem like overkill when looking at logs as a whole without focusing on one specific document; but the value comes about when log events are indexed and we want to search for specific doc IDs and get a full picture of the lifecycle of that particular document

kyle-kmw (approved):

PR #555, opened 2026-08-18
Make the Kafka publish path asynchronous and configure the producer deliberately

gsingers:
> **Draft** — opened for early visibility while testing is still in progress. Please don't merge yet. See "Status" at the bottom for what is outstanding.
>
> No Jira key on the title yet — happy to file one or take a key from a maintainer and rename.

Two related changes to the Kafka path, developed as a stack ([gsingers#1](https://github.com/gsingers/lucille/pull/1), [gsingers#2](https://github.com/gsingers/lucille/pull/2)) and combined here because the base branch of the second can't exist in this repo. The first makes publishing asynchronous; the second stops the producer's throughput and durability settings from being inherited defaults. They're separable, but the second is what makes the first pay off, so reviewing them together is probably easier. Commit history is preserved if you'd rather read them apart.

---

## Part 1 — Make the publish path asynchronous, drop the per-record `flush()`

`KafkaPublisherMessenger.sendForProcessing` used the producer as a synchronous RPC client: `send().get()` followed by a producer-global `flush()`, per document.

`flush()` blocks the caller until every record buffered by every **other** publishing thread has been acked, so its cost grows with the publishing thread count rather than staying at one ack round trip. It is also strictly redundant — the preceding `.get()` had already blocked until that record was acked by all in-sync replicas. A per-record `flush()` measures out equivalent to globally setting `linger.ms=0`.

The same `send().get()` + `flush()` pair was copy-pasted across five messengers; two sites carried a bare `// TODO` next to the flush. It dates to the first Kafka commit in July 2021, when the publisher was a single thread reading a CSV file.

- Remove the seven redundant `flush()` calls across the five messengers.
- Add `PublisherMessenger.sendForProcessing(Document, SendCallback)`, defaulting to a synchronous send with an inline callback, so `LocalMessenger` and any external implementation are unaffected. `KafkaPublisherMessenger` overrides it with a real callback-based send.
- `PublisherImpl` publishes through the callback form; on failure the callback stops tracking the document and signals the `maxPendingDocs` condition.
- Convert `numCreated`/`numFailed`/`numSucceeded`/`numDropped` to `AtomicLong` — previously safe only because `handleEvent()` ran on a single thread.

### Failure semantics — worth review attention

Going async means a send can fail after `publish()` has returned, with nowhere to throw. Untracking the document is mandatory (otherwise `waitForCompletion` blocks to the connector timeout awaiting a terminal event that can never arrive), but untracking *alone* converts today's fail-fast into silent data loss — the run would report success having dropped documents. So the first such failure is latched and surfaced twice: thrown from the next `publish()`, and checked at `waitForCompletion`'s termination branch so the run fails. Happy to drop either check if you'd rather publish failures be tolerated the way pipeline FAIL events are.

### Ordering

The async path preserves per-document ordering only if idempotence is on, which Lucille never set explicitly. Resolved against the kafka-clients 4.2.0 jar: `acks=-1`, `enable.idempotence=true`, `max.in.flight=5`, `retries=MAX` — so a retried batch cannot land after a later one. Since these defaults have already moved twice in routine upgrades (`acks` 1→all in 3.0, `linger.ms` 0→5 in 4.0), `KafkaUtilsTest` now pins them rather than trusting a comment.

---

## Part 2 — Configure the producer deliberately instead of by inheritance

`createProducerProps` set five properties. Everything else governing throughput and durability was an inherited client default — including the three above, each arriving in a dependency bump whose commit message says nothing about the behavioural change. The current values happen to be right, but they are right *by accident*.

**Pinned, deliberately not configurable:** `acks=all`, `enable.idempotence=true`. The Publisher stops tracking a document as soon as the producer reports it accepted, so acceptance has to mean replicated, and a retried batch must not land after a later one. `kafka.producerPropertyFile` remains the escape hatch for a deployment that genuinely needs otherwise.

**New optional config keys:**

| key | default | why |
|---|---|---|
| `batchSize` | 128 KB | the client's 16 KB is small for Lucille documents now that batches actually fill |
| `compressionType` | `lz4` | documents are JSON; cost is per batch, so this only pays once sends stopped being one record at a time |
| `deliveryTimeoutMs` | 120000 | the client default, stated rather than inherited |
| `bufferMemory` | 32 MB, or `maxRequestSize` if larger | no longer aliased to `maxRequestSize` |
| `lingerMs` | 0 | see below |

`buffer.memory` and `max.request.size` bound unrelated things — the whole accumulator versus a single record. Bringing `maxRequestSize` down to a size a broker will actually accept (the example config's 250 MB is 250× a typical `message.max.bytes`) silently shrank the send buffer with it, and far enough down would make `send()` block on `max.block.ms` for want of space. The default still covers an oversized `maxRequestSize`, so existing configs keep working.

`linger.ms` stays at 0: the publisher's sends are async now and batch from queue depth without one, while the worker's document sends and every event send still block on the ack, where any linger adds straight to per-record latency. Worth revisiting when those paths go async.

Also: every producer now carries a role-suffixed `client.id` (`-documents` / `-events`), sharing the random suffix the consumers already use — without it the broker cannot tell Lucille's producers apart in request logs, metrics, or quotas. And the `producerPropertyFile` branch previously returned before the key serializer was set, so such a deployment had to know to supply it; it now gets the client id and a default key serializer, with the rest of the file still left alone.

---

## Testing

`KafkaPublisherFlushTest` characterizes the defect and guards the fix — its first commit fails against the old code. `PublisherImplAsyncSendTest` covers the new async-failure paths, including that a synchronous send failure signals publishers parked on `maxPendingDocs`. `KafkaUtilsTest` pins the producer settings the async path depends on.

Measured on an embedded broker, 16 threads × 40 records — the cheapest possible case for the synchronous arm, since round-trip time is near zero:

```
[async] records/request=91.43  wall=14ms
[sync]  records/request=2.79   wall=81ms
```

## Status / outstanding

- Testing is **not** complete — this is why the PR is a draft. The full-reactor verification on JDK 21 (what CI runs) has not been re-run since the most recent commits.
- `KafkaTest`, the end-to-end `KAFKA_LOCAL` run through `Runner`, is the check that matters most for run accounting under async sends; it needs a clean re-run before this leaves draft.
- The `.gitignore` commit adds a local tooling entry and should be dropped before merge if you'd rather not carry it.
- Related: #528 (LC-842, "Support Arbitrary Kafka Properties") would make much of Part 2 settable from config instead. It has been open since June and untouched since 2026-07-15 — happy to rebase onto it, or to rework Part 2 in its terms, if it's still live.

## Not included

Bridging `producer.metrics()` into `SharedMetricRegistries`, and the same async treatment for the worker and indexer event paths.

https://claude.ai/code/session_01CjfYT54zz8bYCHv6Np9i26


rseitz:
Thanks for this PR (#555). We have an another in-flight PR (#554) that make similar changes but takes a simpler approach, opting for fail-fast behavior without any attempt to recover from a kafka send failure. Here's an AI-assisted comparison of the two approaches:

> Both PRs make the publisher path asynchronous, but they take different approaches to the consequences:
> 
> PR #554 assumes fail-fast behavior: a Kafka send failure aborts the run, so there is no need to precisely clean up docIdsToTrack for a failed document or make event counters thread-safe. This allows simpler code — the messenger encapsulates async internally, PublisherImpl is unchanged, and no new interfaces or callback types are introduced. PR #554 also makes the Indexer's per-batch event sending asynchronous (reducing the event-send phase from N sequential round trips to a single flush), which likely provides a larger throughput win than async publishing — the Indexer sends events in a tight back-to-back loop with no intervening work, whereas the publisher's send overlaps with connector I/O that's often the bottleneck anyway. PR #554 additionally enforces consistency across connector implementations, ensuring none swallow publish() exceptions — which is important because async send failures now surface on a subsequent publish call, and a connector that swallows that exception would silently continue against a broken broker.
> 
> PR #555 takes a more precise approach: it passes a callback to the messenger so that PublisherImpl can clean up the specific failed docId and increment a failure counter from the producer's I/O thread. This requires AtomicLong counters, a new SendCallback interface, a second overload on PublisherMessenger, and a new check in waitForCompletion(). This machinery would be necessary if the system were designed to tolerate partial publish failures and continue running — but given the fail-fast policy, it solves a problem that doesn't arise. PR #555 also adds explicit configuration keys for producer properties (kafka.lingerMs, kafka.batchSize, kafka.compressionType, etc.), but these can already be configured via the existing kafka.producerPropertyFile mechanism, which loads a standard Java properties file containing any Kafka producer settings.


And here's a more detailed comparison and explanation of why we'd likely opt for the #554 approach:


---

The Lucille framework is not currently resilient to Kafka communication failures beyond whatever failover and retry logic resides inside the Kafka client itself. For example, if the Indexer fails to send a FINISH event to the event topic, the run will hang because the Publisher will never learn that the document completed — the Indexer has no retry logic for Kafka sends. In practice, Kafka communication failures have proven rare and tend to be systemic and unrecoverable (broker down, topic deleted, authorization revoked), so the added code complexity of handling them gracefully inside the Lucille framework has not seemed justifiable.

Therefore, as we revisit uses of `kafkaProducer` throughout Lucille — removing redundant `flush()` calls and making `send()` async where appropriate — we want to preserve a fail-fast approach and keep the code as simple as possible. Specifically:

**We don't need to expose sync vs. async as a caller choice.** PR #555 introduces a `sendForProcessing(Document, SendCallback)` overload alongside the original sync `sendForProcessing(Document)`, giving callers a choice of paradigm. Since the framework should uniformly use async sends for the publisher path, there is no reason to maintain both signatures. The messenger can simply be async internally — callers don't need to know or decide.

**We don't need to allow the caller to supply a callback.** In #555, `PublisherImpl` passes an `onSendComplete` callback that handles docId cleanup and error capture. This pushes async error-handling complexity into `PublisherImpl` (a class that previously had no awareness of async semantics) and introduces a new `SendCallback` functional interface. In #554, the messenger encapsulates the async behavior entirely — `PublisherImpl` continues to call `sendForProcessing()` and catch exceptions as it always has. The callback and error state live inside the messenger where they can't leak into other concerns.

**We don't need to remove the failed docId from `docIdsToTrack`.** In #555, the callback explicitly removes the failed document's ID from tracking so that `waitForCompletion()` doesn't hang waiting for an event that will never arrive. This is precise but unnecessary: because the run is going to abort, `waitForCompletion()` will detect `thread.hasException()` and return immediately without checking pending docs. The orphaned docId is inert.

**We don't need to make the event counters atomic.** PR #555 converts `numCreated`, `numFailed`, `numSucceeded`, and `numDropped` from `long` to `AtomicLong` because its callback increments `numFailed` from the Kafka producer's I/O thread. This thread-safety concern only exists because of the callback design — the callback touches state that was previously only accessed from a single thread. In #554, failures surface as exceptions on the connector thread, so the counters are never mutated from a background thread and don't need atomicity.

**We don't need a separate check in `waitForCompletion()`.** PR #555 adds a `publishException` check at the end of `waitForCompletion()` to catch the case where the connector finishes cleanly but a send failed asynchronously. In #554, this case is handled by the existing `ConnectorThread` lifecycle: `publisher.flush()` is already called after `connector.execute()` returns, and the messenger's `flush()` ensures all buffered sends are acked (or surfaces an error) before the thread exits. If the last send failed, the exception propagates through `ConnectorThread`'s existing catch block, and `waitForCompletion()` detects it via `thread.hasException()` — the same mechanism it uses for any other connector failure.

**We don't need the `signalIfPendingDocsBelowMax()` refactor.** PR #555 extracts the "notify waiting threads" logic into a helper so it can be called from both `handleEvent()` and the async failure callback. Since #554 doesn't have a callback that touches `docIdsToTrack`, this refactoring is unnecessary — the existing inline code in `handleEvent()` remains the only call site.

In summary, #555's additional machinery is internally consistent and would be correct for a design that tolerates partial failures and continues running. But that's not the goal. Given that publish failures abort the run, the simpler approach — encapsulate async in the messenger, let the exception propagate through existing mechanisms, don't clean up state that's about to be discarded — achieves the same outcome with fewer moving parts and no changes to `PublisherImpl`'s threading model.

meghanboyd:
Hi @gsingers , we have merged [PR 554 (LC-391)](https://github.com/kmwtechnology/lucille/pull/554) from @rseitz and this change is now available as part of [Lucille 0.11.1 release. ](https://github.com/kmwtechnology/lucille/releases/tag/0.11.1).

We think this, in addition to some other changes in this release, will result in performance improvements for Kafka use cases at scale. Based on this I would like to propose closing this PR, and please do let us know if you see any further issues to address. Thank you!

gsingers:
Superseded — this work landed upstream via #528 (arbitrary Kafka properties + idempotence), #554 (async send path, publish exception propagation), and #558 (PublisherImpl ConcurrentHashMap). Closing.

PR #557, opened 2026-08-20
LC-923 Resolved CVE's for 0.11.1

LunkadV:
CVE bumps for lucille-core dependencies.

| Change | CVE |
|---|---|
| `httpclient5` 5.6.2 → 5.6.3 | CVE-2026-64607 |
| `netty` 4.1.136.Final → 4.1.137.Final | CVE-2026-59903 (`netty-codec-http`) |
| `jetty-http` / `jetty-util` — removed `<scope>test</scope>` | CVE-2026-2332, CVE-2026-8384 |

The two version bumps in `lucille-parent` cover the Snyk findings against `azure-storage-blob`, `awssdk:s3`, `elasticsearch-java`, `opensearch-java` and `opensearch-log4j-appender`. They were all flagged for `httpclient5` / `netty-codec-http`, so the libraries themselves don't need upgrading.

The Jetty pins were test-scoped,  every downstream module still resolved the vulnerable 12.0.27 at runtime. Removing the scope makes the updated version propagate. The pinned versions were not changed.

### DIFF
--- a/lucille-parent/pom.xml
+++ b/lucille-parent/pom.xml
@@ -64,7 +64,7 @@
     <dropwizard-core.version>5.0.2</dropwizard-core.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <junit4.version>4.13.2</junit4.version>
-    <aws-sdk.version>2.49.4</aws-sdk.version>
+    <aws-sdk.version>2.52.1</aws-sdk.version>
     <hadoop.version>3.5.0</hadoop.version>
   </properties>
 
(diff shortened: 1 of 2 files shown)


lucille-parent/pom.xml
mrwrench24: recommend upgrade to `2.52.1`

mrwrench24 (changes requested):
can we upgrade aws-sdk and azure-storage-blob while we are going?

these are new updates from snyk that just popped up - not something that you missed

bump aws-sdk to `2.52.1`, azure-storage-blob to `12.35.1`

mrwrench24 (approved):
thx!

spencersolomon6 (approved):

PR #558, opened 2026-08-20
LC-926: Replace SynchronizedBag with ConcurrentHashMap in PublisherImpl to eliminate lock contention under multi-threaded publishing

rseitz:
This PR aims to improve publishing throughput in the context of a fast, multi-threaded connector.

Previously, PublisherImpl's `docIdsToTrack` (and `docIdsIndexedBeforeTracking`) used a `SynchronizedBag` — a single-mutex wrapper around a `HashBag`. Only one publishing thread could add a new document ID to the tracking structure at a time, regardless of whether the IDs were different. Every `publish()` call acquired the same lock, serializing all publishing threads against each other and against the event-handling thread.

With this change, the tracking structures use `ConcurrentHashMap<String, AtomicInteger>`. Multiple threads can add, remove, or check distinct document IDs concurrently without acquiring a shared lock. Contention only occurs when two threads operate on the *same* key simultaneously, which is rare in practice.

This improvement is most likely to be impactful when publishing is very fast — for example, a multi-threaded connector generating lightweight documents with minimal I/O per document. In slower publishing scenarios where each thread spends most of its time doing connector work (HTTP fetches, database queries, file reads), the previous lock contention would not have been an observable bottleneck.

The original `SynchronizedBag` was introduced as a performance improvement over a `Collections.synchronizedList` (O(1) removal by value vs. O(n)) before the codebase supported multi-threaded publishing. When multi-threaded publishing was added later, the existing synchronized structure was inherited. This change is a natural follow-up to multi-threaded publishing support.

`numPending()` now reports the number of distinct document IDs being tracked, rather than the sum of all occurrence counts. Preserving the old behavior with the new data structure would require iterating the entire map and summing all `AtomicInteger` values on every call — unnecessary overhead for a value that is checked frequently (every event-poll iteration in `waitForCompletion`, every `publish()` call when `maxPendingDocs` is configured). The new semantics use `ConcurrentHashMap.size()` directly. This only differs from the old behavior when the same document ID is published multiple times (e.g. streaming-mode create/update sequences or at-least-once redelivery). The `maxPendingDocs` threshold is documented as non-strict and is unaffected in practice.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
@@ -82,10 +82,11 @@ public class PublisherImpl implements Publisher {
 
   // Map of published document IDs that have not reached a terminal state. Also tracks children.
   // Uses ConcurrentHashMap for lock-free concurrent access from multiple publishing threads and
-  // the event-handling thread. The AtomicInteger value is a count: if two documents with the same
-  // ID are published (e.g. create + update in streaming mode, or at-least-once redelivery), we
-  // expect to receive separate terminal events for each, so each add increments and each remove
-  // decrements. The entry is evicted when the count reaches zero.
+  // the event-handling thread. The AtomicInteger value is used as a mutable int counter (its
+  // atomicity is not relied upon — all access is within compute() which holds the bin lock).
+  // If two documents with the same ID are published (e.g. create + update in streaming mode,
+  // or at-least-once redelivery), we expect to receive separate terminal events for each, so
+  // each add increments and each remove decrements. The entry is evicted when the count reaches zero.
   //
   // Note that a Publisher may be shared by a Runner and a Connector: the connector may be publishing
   // new Documents (via publish()) while the Runner is receiving Events and calling handleEvent().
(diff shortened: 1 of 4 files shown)


lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
```diff
@@ -261,14 +265,14 @@ private void sendForProcessing(Document document) throws Exception {
     // As soon as the document has been sent for processing, the publisher could begin receiving Events relating to that document.
     // If the publisher quickly receives a DROP event, for example, we want to be sure that the docId has already been
     // added to docIdsToTrack so that it can be found there and removed, not mistakenly added to docIdsIndexedBeforeTracking
-    docIdsToTrack.add(docId);
+    docIdsToTrack.computeIfAbsent(docId, k -> new AtomicInteger(0)).incrementAndGet();
```
mrwrench24: **Caveat: This was picked up by an AI review**. I have carefully dug into it and think it may be worth verifying or discussing...

We may want to make adding a `docId` an atomic operation to prevent a small race condition. Adding is currently two steps: first, `computeIfAbsent` and `incrementAndGet`. Then, we do `decrementOrRemove`. 

If another (publish) thread increments a key in between these two steps, the entry will get removed, and a published doc will no longer be tracked. This means an undercount in `numPending`. 

It's rare, since it implies the same `docId` is somehow in flight on two threads at once, but it is a new potential issue.  Using `merge` for the add and `compute` for the remove should be a straightforward resolution, if we want.

mrwrench24 (approved):
one small comment I want to double-check but otherwise LGTM

kyle-kmw (approved):

lucille-core/src/main/java/com/kmwllc/lucille/core/PublisherImpl.java
rseitz: thanks for flagging this. I committed a fix

mrwrench24 (approved):

PR #559, opened 2026-08-21
LC-908: Improved & Tested Lucille Core Docker Image

mrwrench24:
This PR updates the `Lucille-core` Docker Image to support running as an indexer, worker, or runner specifically (as well as using distributed mode etc.). Previously, the image only supported specific local runs. 

To that end, the distributed example has been updated, as we can now use our `Lucille-core` Docker Image to build the worker, runner, and indexer in that example. This means we are now testing the Docker Images as part of the example / the nightly smoketest.


lucille-examples/lucille-distributed-example/docker-compose.yml
```diff
@@ -1,4 +1,24 @@
 name: distributed-lucille
+
+# Runner, Worker, and Indexer all run from the same published lucille-core image;
```
rseitz, line 3: This comment says the services run from the same _published_ lucille-core image, but it looks like we're actually building it here, not referencing a published one — which is appropriate. Could the comment reflect that? What we're really doing is building from the same root Dockerfile that's used to publish lucille-core. So we get coverage of that Dockerfile, but we aren't proving the most-recently-published image itself is in good shape (out of scope for this PR anyway).

lucille-examples/lucille-distributed-example/pom.xml
```diff
@@ -51,12 +51,7 @@
               </execution>
             </executions>
             <configuration>
-              <executable>docker</executable>
-              <arguments>
-                <argument>compose</argument>
-                <argument>up</argument>
-                <argument>--abort-on-container-exit</argument>
-              </arguments>
+              <executable>${project.basedir}/scripts/run-nightly.sh</executable>
```
rseitz, line 54: where is this run-nightly.sh script? I don't see it in this PR or on main

rseitz:
I attempted to run the smoketest locally and saw a failure (0 documents in solr while 6 were expected). I attempted to delete existing docker images and rebuild but this didn't resolve it. An AI assisted investigation took multiple turns and eventually produced this theory (not verified):

> Context: Solr was recently bumped to 10.0.0 (commit 95eea1bc, PR #535), which also switched the Solr cloud client in SolrUtils.getCloudClient to be backed by HttpJdkSolrClient (Java's built-in HTTP client). So both the server version and the client transport in this path are new. Note also that in Solr 10 the Jetty client Http2SolrClient was renamed to HttpJettySolrClient.

> Finding: it's a classpath/dependency problem in lucille-core, not client code: the same HttpJdkSolrClient.add() against Solr 10.0.0 works on a clean SolrJ classpath but fails on lucille-core/target/lib/*.  Likely a conflicting dependency version in the assembled lib (standout: jackson-core 2.22.1 vs SolrJ's expected 2.20.0), not yet isolated.

> Symptom: on the failing classpath, add() returns success with no exception, but Solr receives an empty request body (missing content stream) and indexes nothing — silent data loss.

rseitz (approved):

PR #560, opened 2026-08-21
LC-925: Tika Performance Improvements

mrwrench24:
This PR introduces a number of improvements to the `TextExtractor` Stage, a number of which are performance-related workarounds for certain Tika use cases:

- Introduces the ability to skip certain types of embedded content types. Currently only works with the non-forked parser. 
- Creates a fixed size (1,000) cache for Tika Configs, since reading them from a file can be expensive. This is a very elementary cache that should support a number of use cases and does not have unbounded growth.
- Calling `parseCtx.set(TikaConfig.class, tikaConfig)`. This prevents a `new TikaConfig()` from being created for every embedded document. Perhaps the only change that users may need to be aware of - if they were reliant on the fact that Tika used the default config for their embedded documents. 
- `handleParseTimeout` hook / callback for subclasses
- Improved timeout handling: close the stream and discard + recreate the executor on timeout to prevent blocking documents / zombie threads 
- Better handling of a `WriteLimitReachedException` - when `textContentLimit` is reached

### DIFF
--- a/lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
+++ b/lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
@@ -374,6 +374,11 @@ public class TextExtractor extends Stage {
       } catch (TimeoutException e) {
         future.cancel(true);
         log.warn("Tika parsing timed out after {} ms", parseTimeout);
+        // future.cancel() doesn't stop a parse thread (may be stuck), but discard + recreate does -
+        // prevent zombie thread / blocking queued documents
+        executorService.shutdownNow();
+        executorService = Executors.newSingleThreadExecutor();
+        handleParseTimeout(inputStream);
       } catch (Exception e) {
         log.warn("Error during async Tika parsing: {}", e.getMessage());
       }
@@ -394,6 +399,13 @@ public class TextExtractor extends Stage {
         }
       }
     }
+  }
+
+  /**
+   * Hook invoked when a parse attempt times out, after the executor has been discarded and recreated.
+   * No-op by default; subclasses can override it, e.g. to record metrics on the timed-out document.
+   */
+  protected void handleParseTimeout(InputStream inputStream) {
   }
 
   private void parse(Document doc, InputStream inputStream, Metadata metadata, ContentHandler bch) throws StageException {
--- a/lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
+++ b/lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
@@ -48,7 +48,9 @@ import org.xml.sax.SAXException;
  * filePathField (String, Optional) : name of field from which file path can be extracted, if filePathField
  * and byteArrayField both not provided, stage will do nothing
  * byteArrayField (String, Optional) : name of field from which byte array data can be extracted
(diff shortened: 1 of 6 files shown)


lucille-plugins/lucille-tika/src/test/java/com/kmwllc/lucille/tika/stage/TextExtractorTest.java
```diff
@@ -605,6 +631,41 @@ public void testForkingParserWithMemoryHogParser() throws Exception {
...
 
+  /**
+   * Sanity check that a TikaConfig is built once per config path and reused across stage instances,
+   * rather than rebuilt for every stage (as would happen with each worker thread in a real run).
+   */
+  @Test
+  public void testTikaConfigCaching() throws Exception {
+    // Clear the JVM-wide cache so this test doesn't depend on ordering with other tests, and so we
+    // don't leave a mock config behind for them afterward.
+    Field cacheField = TextExtractor.class.getDeclaredField("TIKA_CONFIG_CACHE");
```
rseitz, line 642: Avoid using reflection to update field visibility.
I've seen AI agents freely use this approach in tests, but it makes code difficult to reason about since it basically removes the distinction between public and private, so we should find a different/better way of setting up the code to be testable. In this case, you could put a package-private static method on TextExtractor that lets you clear the cache.

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -354,14 +430,26 @@ public void parseInputStream(Metadata metadata, Document doc, InputStream inputS
     }
   }
 
+  /**
+   * Hook invoked when a parse attempt times out, after the executor has been discarded and recreated.
+   * No-op by default; subclasses can override it, e.g. to record metrics on the timed-out document.
+   */
+  protected void handleParseTimeout(InputStream inputStream) {
```
rseitz, line 437: What might a subclass overriding this method actually want to do with the inputStream? We've already tried to close it by the time handleParseTimeout is invoked. And if I'm reading correctly, another thread might still have a reference to this inputStream. Suggest documenting the intended use of this param or else removing it.

lucille-plugins/lucille-tika/src/main/java/com/kmwllc/lucille/tika/stage/TextExtractor.java
```diff
@@ -109,6 +118,16 @@ public class TextExtractor extends Stage {
...
   private static final Logger log = LoggerFactory.getLogger(TextExtractor.class);
+
+  // Building TikaConfigs from files is expensive, so we cache these configs across the JVM.
+  // This is particularly helpful in multi-threaded runs, where each thread has its own instance of
+  // TextExtractor but is using the same TikaConfig.
+  // The cache is bounded by MAX_CACHED_CONFIGS. It is a "soft cap" - once it is reached,
+  // no additional TikaConfigs will be cached.
+  // Canonical paths are used as keys to prevent unwarranted collisions.
+  private static final int MAX_CACHED_CONFIGS = 1000;
+  private static final ConcurrentHashMap<String, TikaConfig> TIKA_CONFIG_CACHE = new ConcurrentHashMap<>();
```
rseitz, line 129: could set this up similar to the way we configure the historicalDetailsMap in RunnerManager to get auto-eviction according to get FIFO eviction 

https://github.com/kmwtechnology/lucille/blob/b2b93278bad1623c24d1664c9f9f9d88183eda8c/lucille-core/src/main/java/com/kmwllc/lucille/core/RunnerManager.java#L50

note: RunnerManager uses a plain LinkedHashMap but here we'd need to wrap it in Collections.synchronizedMap

PR #561, opened 2026-08-25
LC-815: Add optional multithreaded traversal to FileConnector

LunkadV:
Adds an optional `multithreaded` flag to the FileConnector. When set to true, each configured path is traversed on its own thread with one thread per path. Files under a single path are not split across threads and paths are assumed to not overlap. 

`FileConnectorStateManager` retains the same public methods and they are still called the same way from `BaseStorageClient`. However those methods now delegate to the calling thread's `TraversalState` a new thread local class holding one JDBC connections and its prepared statements. `init()` binds a state for the thread that calls it, each additional traversal thread binds its own with `openStateForThread()` and releases it with `closeStateForThread()`. 

The manager retains its own separate connection for whole-table operations such as table creation, resetting encountered flags, `incrementRunsNotEncountered()`, `listExpiredFiles()`, and deletions which run only on the connector thread, before or after the traversal window.

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
@@ -10,9 +10,17 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.ThreadFactory;
 
+import org.apache.commons.lang3.concurrent.BasicThreadFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.kmwllc.lucille.util.ThreadNameUtils;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.kmwllc.lucille.connector.storageclient.BaseFileReference;
@@ -45,6 +53,8 @@ import com.typesafe.config.Config;
  * <ul>
  *   <li>paths (List&lt;String&gt;, Required) : Paths or URIs to traverse (local paths or cloud storage URIs). s3 URIs must be
  *   percent-encoded; unencoded spaces or special characters will not be recognized. For example, use s3://test/folder%20with%20spaces.</li>
+ *   <li>multithreaded (Boolean, Optional) : Traverse each of the paths concurrently, on its own thread. Only applies
+ *   when more than one path is given; Requires that the paths do not overlap. Defaults to false.</li>>
  *   <li>filterOptions.includes (List&lt;String&gt;, Optional) : Regex patterns to include files.</li>
  *   <li>filterOptions.excludes (List&lt;String&gt;, Optional) : Regex patterns to exclude files.</li>
  *   <li>filterOptions.pathsToSkip (List&lt;String&gt;, Optional) : URIs of paths to directories that should be skipped and not traversed.</li>
@@ -123,6 +133,7 @@ public class FileConnector extends AbstractConnector {
 
   public static final Spec SPEC = SpecBuilder.connector()
       .requiredList("paths", new TypeReference<List<String>>(){})
+      .optionalBoolean("multithreaded")
       .optionalParent(
           SpecBuilder.parent("filterOptions")
               .optionalList("includes", new TypeReference<List<String>>(){})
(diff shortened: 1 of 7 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/connector/FileConnectorTest.java
```diff
@@ -893,4 +897,167 @@ public void testTraversalMultiplePathsAndSkipPath() throws Exception {
...
+
+      // the paths are waited on in order, so directory1 becomes the cause and directory3 is suppressed onto it
+      assertTrue(e.getCause().getMessage().contains("directory1"));
+      assertEquals(1, e.getSuppressed().length);
+      assertTrue(e.getSuppressed()[0].getMessage().contains("directory3"));
+    }
+  }
+
+  // A StorageClient that fails when traversing any of the named directories, and otherwise publishes one Document
+  // named after the directory it traversed.
```
spencersolomon6: [nit] This comment could be changed to improve ease of understanding, took me a few read-throughs to get it. For inspiration:

"Returns a mock StorageClient for testing purposes. Fails and throws an IOException when traversing any of the given `directoriesToFail`. For all other directories, one document is published with an id equal to the name of the directory."

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -123,6 +133,7 @@ public class FileConnector extends AbstractConnector {
 
   public static final Spec SPEC = SpecBuilder.connector()
       .requiredList("paths", new TypeReference<List<String>>(){})
+      .optionalBoolean("multithreaded")
```
rseitz: suggest changing "multithreaded" to "concurrent" here, as well is in comments and variable names

the reason is: threading is an implementation detail. What we're offering here from a user perspective is that the traversals will happen simultaneously or close-to-simultaneously. Ideally the user shouldn't have to know what "threads" are in order to understand what the feature is providing. 

Note that "parallel" would be a bit too strong a claim since you don't get true parallelism unless you're running on hardware with enough CPU cores to map each thread to its own core. So "concurrent" is the more honest promise.

FWIW this wasn't flagged initially by AI but I did ask AI where the changes should be made and here's the output I got:

Config key: multithreaded → concurrent (this is the most user-facing one and the most worth getting right)
Field: private final boolean multithreaded → concurrent
Method: traversePathsWithMultithreading → traversePathsConcurrently
Method: traverseStoragePathOnThisThread → keep or rename; "on this thread" is accurate and describes what the method does, not the broader feature [rseitz note: yes, let's definitely keep]
Javadoc on the class and method that explains the behavior: replace "multithreaded traversal" with "concurrent traversal"
The assumption note in FileConnectorStateManager's Javadoc ("A FileConnector traversal may be multithreaded") → "A FileConnector traversal may be concurrent" [rseitz note: arguable we could keep this one as as is too]

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -147,9 +158,13 @@ public class FileConnector extends AbstractConnector {
 
   private final FileConnectorStateManager stateManager;
 
+  private final boolean multithreaded;
+
   public FileConnector(Config config) throws ConnectorException {
     super(config);
 
+    this.multithreaded = ConfigUtils.getOrDefault(config, "multithreaded", false);
```
rseitz: as discussed, we should validate that `collapse` has not been enabled on the connector if we're being asked to do a multithreaded/concurrent traversal

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
rseitz, line 171: let's investigate how we might enforce that the paths should be non-overlapping
might be tricky because the paths can be for arbitrary storage backends which each have their own format
it's acceptable to defer the work for another PR or say we won't do anything --- but let's do the initial investigation now and at least create a ticket for any followups

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/BaseStorageClient.java
```diff
@@ -112,6 +113,47 @@ public final void traverse(Publisher publisher, TraversalParams params, FileConn
...
+
+    return pathContains(normalizedParent.getPath(), normalizedChild.getPath());
+  }
+
+  /**
+   * Returns whether the child path is the parent path, or sits underneath it, treating both as '/'-delimited.
+   * @param parentPath The path that may contain the other.
+   * @param childPath The path that may sit under the other.
+   */
+  protected static boolean pathContains(String parentPath, String childPath) {
```
rseitz: the naming choice of pathContains() creates some confusion with containsPath()

suggest renaming this pathContains() method to something like isUnder()

also suggest making this private, not protected; there is no current need to override it in a subclass and arguably an attempt to override just this helper by itself could be confusing

rseitz (approved):

PR #562, opened 2026-09-01
Raise Jackson maxStringLength so Documents with large byte[] fields survive the Kafka round-trip

gsingers:
`KafkaDocumentDeserializer`, `JsonDocument` and `HashMapDocument` used a stock `ObjectMapper`, whose default `StreamReadConstraints.maxStringLength` is 20,000,000 chars. A `byte[]` field (e.g. `file_content` from `FileConnector` with `getFileContent: true`) serializes as one base64 string, so any file of 15 MiB or more is written by the serializer and accepted by the broker, then fails in every consumer with `StreamConstraintsException` → `RecordDeserializationException`, which kills the Worker thread and orphans its partitions.

This adds `JsonDocument.createDocumentMapper()` with `maxStringLength` raised to `Integer.MAX_VALUE` (other constraints unchanged) and uses it for the three Document-reading mappers. The record size is already bounded by `max.request.size` / `message.max.bytes`, so a lower consumer-side limit only adds a hidden cliff below limits users have already sized. `KafkaDocumentDeserializerTest` round-trips a 15 MiB `byte[]` through the Kafka serializer/deserializer and `Document.createFromJson`; both cases fail on `main` and pass here.

Generated with Claude Code

https://claude.ai/code/session_01CcuhP2BcLFwZqHQVjbXe6u

### DIFF
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
@@ -2,7 +2,9 @@ package com.kmwllc.lucille.core;
 
 import com.dashjoin.jsonata.Jsonata;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -29,13 +31,32 @@ import java.util.function.UnaryOperator;
  */
 public class JsonDocument implements Document {
 
-  protected static final ObjectMapper MAPPER = new ObjectMapper();
+  protected static final ObjectMapper MAPPER = createDocumentMapper();
   private static final TypeReference<Map<String, Object>> TYPE = new TypeReference<Map<String, Object>>() {
   };
   private static final Logger log = LoggerFactory.getLogger(JsonDocument.class);
 
   @JsonValue
   protected ObjectNode data;
+
+  /**
+   * Creates the ObjectMapper used to parse Documents from JSON.
+   *
+   * Jackson's default StreamReadConstraints cap any single JSON string value at 20,000,000 chars
+   * (StreamReadConstraints.DEFAULT_MAX_STRING_LEN). A Document's byte[] field is serialized as one
+   * base64 string, so a byte[] of 15 MiB or more (20,054,016 chars) is written without complaint by
+   * KafkaDocumentSerializer, accepted by the broker, and then rejected with a StreamConstraintsException
+   * by every consumer that tries to read it back. The constraint exists to bound untrusted input, but a
+   * Kafka record is already bounded by the producer's max.request.size (kafka.maxRequestSize) and the
+   * broker's message.max.bytes, so a lower limit here only adds a hidden consumer-side cliff below
+   * limits the user has already sized. maxStringLength is therefore raised to Integer.MAX_VALUE; all
+   * other constraints keep their defaults.
+   */
+  public static ObjectMapper createDocumentMapper() {
+    return new ObjectMapper(JsonFactory.builder()
+        .streamReadConstraints(StreamReadConstraints.builder().maxStringLength(Integer.MAX_VALUE).build())
+        .build());
+  }
 
   /**
    * A copy constructor for {@link Document} that deep copies the ObjectNode and verifies the
--- a/lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
+++ b/lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
@@ -40,17 +40,26 @@ public class JsonDocument implements Document {
   protected ObjectNode data;
 
   /**
-   * Creates the ObjectMapper used to parse Documents from JSON.
+   * Creates the ObjectMapper used to parse JSON into Documents.
+   *
+   * This is not a Kafka-only mapper. It backs {@link #MAPPER} (and therefore {@link #fromJsonString},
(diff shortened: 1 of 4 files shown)


lucille-core/src/test/java/com/kmwllc/lucille/message/KafkaDocumentDeserializerTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+import com.kmwllc.lucille.core.Document;
+import com.kmwllc.lucille.core.KafkaDocument;
+import org.junit.Test;
+
+public class KafkaDocumentDeserializerTest {
+
+  // 15 MiB of bytes base64-encodes to 20,054,016 chars, just over Jackson's default
```
rseitz: a Claude review flagged this comment, saying:

`Minor: the boundary math in the javadoc and test comment is wrong. Both claim 15 MiB base64-encodes to 20,054,016 characters; the correct figure is 20,971,520 `

Please double-check

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
```diff
@@ -29,14 +31,33 @@
...
   @JsonValue
   protected ObjectNode data;
 
+  /**
+   * Creates the ObjectMapper used to parse Documents from JSON.
+   *
+   * Jackson's default StreamReadConstraints cap any single JSON string value at 20,000,000 chars
+   * (StreamReadConstraints.DEFAULT_MAX_STRING_LEN). A Document's byte[] field is serialized as one
+   * base64 string, so a byte[] of 15 MiB or more (20,054,016 chars) is written without complaint by
+   * KafkaDocumentSerializer, accepted by the broker, and then rejected with a StreamConstraintsException
```
rseitz: This comment is correct but incomplete: Kafka deserialization is only one of several pathways where a new JsonDocument can be created via an ObjectMapper, so changing JsonDocument.MAPPER has additional consequences. For example, in a filesystem traversal using JsonFileHandler, json lines get passed through Document.createFromJson which uses MAPPER, and this PR removes the max string length cap that would be applied there too. It is likely safe to proceed with the change because Document.createFromJson accepts a String that's already been loaded into memory; if we were handling an unusually large String we'd probably have run out of memory at some point before MAPPER gets invoked on it, so the string-length cap inside MAPPER isn't critical for safety. The point is just to update the javadoc here to acknowledge that this ObjectMapper is used in multiple places, not only the kafka pathway.

Claude's revision/expansion of the above:


> This comment is accurate but scoped too narrowly to Kafka. JsonDocument.MAPPER is also what JsonDocument.fromJsonString() uses (which Document.createFromJson() dispatches to), so this change affects every caller of that method — not just KafkaDocumentDeserializer, which has its own separate MAPPER field and is unaffected either way.
> 
> Concretely: JsonFileHandler calls JsonDocument.fromJsonString() directly for JSON-Lines files when no composite idFields are configured (JsonFileHandler.java:153, the default case) — this PR's relaxed limit applies there too. (The idFields-configured branch uses its own separate ObjectMapper and isn't touched.)
> 
> That said, I think it's safe to proceed. fromJsonString() takes a String, and any caller has necessarily already materialized that full string in memory before calling it — maxStringLength can only reject it after that allocation, not prevent it. For JsonFileHandler specifically, the line is built via BufferedReader.readLine() (through IOUtils.lineIterator), which has no length limit of its own, so an oversized line already risks OOM before MAPPER ever runs — the cap wasn't providing first-line protection here even at its old default. It does add a little defense-in-depth (failing before Jackson builds a TextNode copy and the field gets decoded to a byte[] elsewhere), but that's marginal, not a safety-critical guard, so I don't think it's worth blocking this PR over.
> 
> Separately, JsonFileHandler's unbounded line length looks like a pre-existing gap independent of this PR — worth its own ticket.

lucille-core/src/test/java/com/kmwllc/lucille/message/KafkaDocumentDeserializerTest.java
gsingers: Confirmed and fixed in bb8da3c3. 15 MiB (15,728,640 bytes) encodes to 20,971,520 chars, and the real cliff under the 20,000,000-char default is any byte[] over 15,000,000 bytes (~14.3 MiB). Corrected the javadoc, the deserializer comment and this test comment.

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
gsingers: Agreed. Updated the javadoc in bb8da3c3 to say the mapper backs every JSON-to-Document path (fromJsonString/createFromJson including the JsonFileHandler JSON-Lines case, HashMapDocument, KafkaDocument, KafkaDocumentDeserializer), and to note why relaxing the cap is still safe there: the caller already holds the whole String, so the cap could only reject after the allocation. Left the unbounded readLine() in JsonFileHandler alone as a separate issue.

lucille-core/src/main/java/com/kmwllc/lucille/core/JsonDocument.java
```diff
@@ -40,17 +40,26 @@ public class JsonDocument implements Document {
   protected ObjectNode data;
 
   /**
-   * Creates the ObjectMapper used to parse Documents from JSON.
+   * Creates the ObjectMapper used to parse JSON into Documents.
+   *
+   * This is not a Kafka-only mapper. It backs {@link #MAPPER} (and therefore {@link #fromJsonString},
```
rseitz: minor but I would remove "This is not a Kafka-only mapper" and rephrase the opening of this paragraph to just mention that the mapper is used in multiple places (and continue to list what they are). Someone reading this who had not followed along with the PR history would not have any assumption-to-be-corrected that it *is* a kafka-only mapper.

gsingers: Done in 2351ac7a — the paragraph now just says the mapper is used in several places and lists them.

rseitz (approved):

mrwrench24 (approved):

PR #566, opened 2026-09-09
LC-918: Improve doc site search relevancy

LunkadV:
Replaced docsy's stock search config: five fields, clauses weighted 100/10/1, and a flat edit distance of 2 on every term.

`.Content` is split into `body` and `code` so the two can be weighed apart, along with two derived fields:
- `identifiers` — space-split camelCase and PascalCase, so `batch size` reaches `batchSize`
- `section` — hierarchy titles, so `stages` reaches a page under Stages

Weights: `title^20`, `description^4`, `identifiers^3`, `section^1`, `body^1`, `code^0.25`, tuned in OpenSearch Search Relevance Workbench. 

- Edit distance scales with term length (0 when length < 4, 1 when < 7, 2 otherwise)
- Prefix (`term*`) and substring (`*term*`) weighed separately, 20 and 5.

`buildSnippet` tries `body` then `code`, so a match inside a config sample gets a
real snippet instead of the static excerpt. Also falls back to the stem when the
raw query word is not matched literally.


spencersolomon6:
[curiosity] How were the ranges/values for the dynamic edit distance chosen? For example, OpenSearch uses the following rules for its "AUTO" edit distance mode:

```
- Terms containing 0–2 characters: Requires an exact match (0 edits).
- Terms containing 3–5 characters: Allows a maximum of 1 edit.
- Terms containing 6 or more characters: Allows a maximum of 2 edits.
```
https://docs.opensearch.org/latest/query-dsl/term/fuzzy/

doc/site/assets/js/offline-search.js
```diff
@@ -56,75 +54,113 @@
...
-    // Every relevant occurrence within the snippet window is highlighted.
+    // Edit distance for the fuzzy clause, tiered by term length
+    const fuzzyDistance = (len) => (len < 4 ? 0 : len < 7 ? 1 : 2);
+
+    // Fields a snippet can come from, in priority order
+    // identifiers are excluded because they hold space-split camelCase names that do not appear on screen
+    const SNIPPET_FIELDS = ['body', 'code'];
+
+    // Where a snippet may start and end, per field. Code has no sentences or line breaks so it is cut at whitespace
+    const SNIPPET_BOUNDARY = { body: /[.!?;]\s|\n/, code: /\s/ };
```
spencersolomon6, line 73: Should code be split on a new line as well?

doc/site/assets/js/offline-search.js
```diff
@@ -56,75 +54,113 @@
...
+
+    // Build a snippet around the first relevant match of a query term, trying each
+    // snippet field in order and falling back to the static excerpt
+    // A relevant match is an exact match of a query stem, or contains the raw query word/stem
+    // Fuzzy-only matches are discarded and snippets are cut at sentence/line boundaries 80 characters before and after the match
+    // If no boundary is found, snippet starts from the first match
+    // Every relevant occurrence within the snippet window is highlighted
```
spencersolomon6, line 80: Should probably use a multiline comment for these and other long comments. Believe its /* ... */ in JS.

doc/site/assets/js/offline-search.js
LunkadV: Hugo's transform.plainify strips all html tags and new lines but inserts new lines back in place of </p> and <br> but not for code blocks which is why we lose all new line characters. I could maybe override plainify and simplify field handling here.

LunkadV:
> [curiosity] How were the ranges/values for the dynamic edit distance chosen? For example, OpenSearch uses the following rules for its "AUTO" edit distance mode:
> 
> ```
> - Terms containing 0–2 characters: Requires an exact match (0 edits).
> - Terms containing 3–5 characters: Allows a maximum of 1 edit.
> - Terms containing 6 or more characters: Allows a maximum of 2 edits.
> ```
> 
> https://docs.opensearch.org/latest/query-dsl/term/fuzzy/

Current edit distance barriers were basically chosen off of intuition since it is hard to get genuine metrics for it when I am in charge of coming up with the misspelled queries to test the barriers. However I've now realized that edit distance is decided by the raw term length but applied on the stem. So the query "running" at length 7 gets an edit distance of 2 which is then applied to its stem "run" during search hitting a lot more terms on the index then it should. Unsure which way the fix should go, will address soon.

doc/site/assets/js/offline-search.js
spencersolomon6: Ok, that makes sense. I think its fine to just split on whitespace for code then

spencersolomon6:
> > [curiosity] How were the ranges/values for the dynamic edit distance chosen? For example, OpenSearch uses the following rules for its "AUTO" edit distance mode:
> > ```
> > - Terms containing 0–2 characters: Requires an exact match (0 edits).
> > - Terms containing 3–5 characters: Allows a maximum of 1 edit.
> > - Terms containing 6 or more characters: Allows a maximum of 2 edits.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [docs.opensearch.org/latest/query-dsl/term/fuzzy](https://docs.opensearch.org/latest/query-dsl/term/fuzzy/)
> 
> Current edit distance barriers were basically chosen off of intuition since it is hard to get genuine metrics for it when I am in charge of coming up with the misspelled queries to test the barriers. However I've now realized that edit distance is decided by the raw term length but applied on the stem. So the query "running" at length 7 gets an edit distance of 2 which is then applied to its stem "run" during search hitting a lot more terms on the index then it should. Unsure which way the fix should go, will address soon.

Maybe we have separate edit distances for the stem and raw term, since we are searching for both?

And I think I would be in support of standardizing the edit distance ranges to match those defined in OpenSearch, since it is *probably* safe to assume they went through some level of effort to determine those ranges as a good baseline for a variety of query use cases. Open to differing opinions there though @LunkadV @rseitz

PR #568, opened 2026-09-16
LC-610 Standardize Authentication in Lucille api

LunkadV:
``AuthHandler.authenticate(user)`` was previously duplicated in all 10 API methods.

Authentication now comes from ``RequireAuthDynamicFeature``, registered once in ``APIApplication``. Jersey calls it per resource method at startup and attaches the Basic auth filter to each one, unless the method carries ``@AuthNotRequired`` or its class is in an exempt package. 

``RequireAuthDynamicFeature`` extends Dropwizard's ``AuthDynamicFeature`` which did the opposite, it would protect methods that opt in which could let an endpoint ship unauthenticated by omission.

Which endpoints require authentication has not been changed. ``/swagger`` and ``/openapi.json`` were previously reachable and still are as exempt packages.


lucille-plugins/lucille-api/src/test/java/APIAuthCoverageTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.glassfish.jersey.server.model.Resource;
+import org.glassfish.jersey.server.model.ResourceMethod;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+// Asserts that every endpoint the API registers requires authentication, except for an explicit allow list,
+// and that valid credentials are accepted everywhere. Endpoints are discovered from Jersey's own resource
+// model rather than listed here, so an endpoint added without authentication fails this test without anybody
+// having to remember to extend it.
+public class APIAuthCoverageTest {
```
rseitz, line 31: should add some basic coverage for auth.enabled=false

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/LivenessResource.java
```diff
@@ -24,7 +23,6 @@
 @Path("/v1/livez")
 @Tag(name = "Health", description = "Health info.")
 @Produces(MediaType.APPLICATION_JSON)
-@PermitAll
```
rseitz, line 27: I suspect the intent of the earlier code with the PermitAll annotation was bypass authentication on these liveness methods even when running with auth.enabled=true

This may not have been working as intended -- we likely never tested that particular scenario

In any case, we should consider adding the AuthNotRequired annotation here. I think it's standard for liveness/readiness endpoints to not require auth.

lucille-plugins/lucille-api/src/main/java/com/kmwllc/lucille/endpoints/ReadinessResource.java
```diff
@@ -24,7 +23,6 @@
 @Path("/v1/readyz")
 @Tag(name = "Health", description = "Health info.")
 @Produces(MediaType.APPLICATION_JSON)
-@PermitAll
```
rseitz, line 27: see comment on LivenessResource -- consider adding AuthNotRequired annotation

spencersolomon6:
LGTM overall, this is a very good approach for standardizing auth across the endpoints. Agreed with Rudi's comments re Liveness/Readiness endpoints, we should make those excluded from Auth.

The one other thing I would like to see as part of this PR is some information in the `lucille-api` README informing developers who are looking to extend the API how auth is handled (and excluded).

lucille-plugins/lucille-api/src/test/java/APIAuthCoverageTest.java
```diff
@@ -0,0 +1,143 @@
...
+  private static final String CREDENTIALS =
+      "Basic " + Base64.getEncoder().encodeToString("admin:password".getBytes());
+
+  private final Client client = RULE.client();
+  private final String url = String.format("http://localhost:%d", RULE.getLocalPort());
+
+  private record Endpoint(String httpMethod, String path) {}
+
+  @Test
+  public void testEveryEndpointRequiresAuthentication() {
```
rseitz, line 55: should this method be renamed now that some methods don't require auth?

rseitz (approved):
approving with request to change test method name

spencersolomon6 (approved):

PR #571, opened 2026-09-22
LC-936: Enable Anonymous S3 Access through S3 StorageClient

LunkadV:
The S3 connector always expects credentials even though reading a public bucket
doesn't need an AWS account at all. Trying to read from one with no configured
account fails with `SdkClientException: Unable to load credentials from any of
the providers in the chain`.

Adds an optional `s3.anonymous` flag that uses the SDK's
`AnonymousCredentialsProvider`, already on the classpath.


lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -44,6 +46,18 @@ protected void validateOptions(Config config) {
...
+    if (anonymous(config) && config.hasPath(S3_ACCESS_KEY_ID)) {
+      throw new IllegalArgumentException("'" + S3_ANONYMOUS + "' cannot be combined with '" + S3_ACCESS_KEY_ID
+          + "' / '" + S3_SECRET_ACCESS_KEY + "' in Config for S3StorageClient.");
+    }
+  }
+
+  // us-east-1 acts as S3's global endpoint and redirects to the bucket's real region.
+  static final Region ANONYMOUS_DEFAULT_REGION = Region.US_EAST_1;
+
+  private static boolean anonymous(Config config) {
```
rseitz: rename to `isAnonymous`

lucille-core/src/main/java/com/kmwllc/lucille/connector/storageclient/S3StorageClient.java
```diff
@@ -44,6 +46,18 @@ protected void validateOptions(Config config) {
...
     }
+
+    if (anonymous(config) && config.hasPath(S3_ACCESS_KEY_ID)) {
+      throw new IllegalArgumentException("'" + S3_ANONYMOUS + "' cannot be combined with '" + S3_ACCESS_KEY_ID
+          + "' / '" + S3_SECRET_ACCESS_KEY + "' in Config for S3StorageClient.");
+    }
+  }
+
+  // us-east-1 acts as S3's global endpoint and redirects to the bucket's real region.
+  static final Region ANONYMOUS_DEFAULT_REGION = Region.US_EAST_1;
```
rseitz: put this constant (static final) up above, near where log is defined

lucille-core/src/main/java/com/kmwllc/lucille/connector/FileConnector.java
```diff
@@ -124,6 +128,7 @@ public class FileConnector extends AbstractConnector {
       .optionalNumber("maxNumOfPages").build();
   public static final Spec S3_PARENT_SPEC = SpecBuilder.parent("s3")
       .optionalString("accessKeyId", "secretAccessKey", "region")
+      .optionalBoolean("anonymous")
```
rseitz, line 131: should add this to application-example.conf under "configuration for S3 (AWS)" 

application-example.conf is meant to illustrate all the config options, not to be a working config by itself, so it's OK to add "anonymous" there alongside secretAccessKey even though they are not compatible. you could add a comment about this

rseitz (approved):

